PR #124, opened 2018-02-15
Modified ctf_stats/init.lua

ClobberXD:
## Changes with respect to original code:
- The maximum attainable points for a kill has been increased from 150 to 200
- The good_weapons list has been modified to include all mese and diamond implements. Steel and bronze tools other than swords have been omitted on purpose as they aren't that good.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,0 +1,282 @@
+ctf_stats = {}
+
+local storage = minetest.get_mod_storage()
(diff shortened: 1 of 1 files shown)


ClobberXD:
I've closed this PR as I accidentally uploaded ctf_stats/init.lua to the root directory of repo. Another PR has been created after modifying the correct file.

PR #156, opened 2018-04-30
Replace team allocation algorithm with one that takes more statistics into account

rdococ:
Mostly untested because I cannot figure out how to get either version to run properly in a singleplayer world without errors.

The algorithm sorts the table in descending order of score to begin with, similarly to the current algorithm (might change that). For the first player and any time in which the total K/D ratio of both teams happen to be equal, a random team is selected. Otherwise, the player is added to the team with the lowest K/D ratio.

It doesn't keep track of the total K/D ratio of both teams; rather it keeps a single "K/D difference" variable which increases for red K/D and decreases for blue K/D.

The algorithm now treats every successful capture as equivalent to 5 kills, meaning that it is now more complex than the above explanation but the code is hopefully simple to understand.

### DIFF
--- a/mods/ctf_alloc/init.lua
+++ b/mods/ctf_alloc/init.lua
@@ -44,8 +44,12 @@ function ctf_alloc.set_all()
 		return {
 			player = a,
 			score = stats.score,
+			
 			kills = stats.kills,
-			deaths = stats.deaths
+			deaths = stats.deaths,
+			
+			captures = stats.captures,
+			attempts = stats.attempts
 		}
 	end)
 	table.sort(players, function(a, b)
@@ -54,7 +58,7 @@ function ctf_alloc.set_all()
 
 	minetest.log("warning", dump(players))
 	
-	local kd_diff = 0
+	local index_diff = 0
 	
 	for _, spair in pairs(players) do
 		local player     = spair.player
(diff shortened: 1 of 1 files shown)


ClobberXD:
K/D ratio is not a consistent measure of a player's pro-ness, and is subject to major fluctuations from player to player based on their style of gameplay. Some great players have very less K/D, which can be attributed to their guns-blaring kind of style. You can possibly see where this is going...

Maybe the algo should also take the players' score into consideration in some way.

rdococ:
Fair enough. I considered using 'score' but rubenwardy mentioned that would also be an inconsistent measure. Might devise a more complex algorithm that uses more stats to determine the general pro-ness of a player (including captures and capture attempts).

ClobberXD:
Interesting proposal... Using the captures and attempts stats is a great idea, but the current implementation still isn't robust. If 1 capture = 5 kills, I feel that players who go for the flag more often will be recognised as a better player, while that might not always be the case. While they do fulfil the game's sole objective of capturing the flag, they are by no means considered a pro just for that one reason. This is primarily a PVP server, and a player's considered to be a pro only if they're good at PVP...

That being said, you're efforts are appreciated, and I encourage you to improve upon this idea, in order to make it merge-worthy.

PR #164, opened 2018-05-23
Added ind_dirt and ind_dirt_with_grass

xerox123official:
Added ctf_map:ind_dirt and ctf_map:ind_dirt_with_grass. Will be used for my new CTF map for the base islands to ensure they don't get destroyed.

### DIFF
--- a/mods/ctf_map/nodes.lua
+++ b/mods/ctf_map/nodes.lua
@@ -34,6 +34,28 @@ minetest.register_node("ctf_map:ind_stone", {
 	is_ground_content = false
 })
 
+minetest.register_node("ctf_map:ind_dirt", {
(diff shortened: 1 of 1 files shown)


mods/ctf_map/nodes.lua
```diff
@@ -34,6 +34,28 @@ minetest.register_node("ctf_map:ind_stone", {
 	is_ground_content = false
 })
 
+minetest.register_node("ctf_map:ind_dirt", {
+	description = "Cheater!",
```
SmallJoker: May distract helping mods (guides) which might be used on the server.

xerox123official: So should all the ind_* nodes be properly described?

PR #186, opened 2018-08-03
[WIP] Implement item blacklisting for team chests

### LINKED ISSUE #63: don't allow junk items in team chest

ClobberXD:
- Restrict unnecessary junk from being added to the team-chest.

The blacklisting code is in place, but the actual items to be blacklisted needs to be decided upon. Items that were useless some time ago now have some added value. Think `default:wood_<tool>`->`default:wood`. Even leaves can be used to quickly climb heights by making 1x1 towers without using cobble... I find saplings useless, but in that case, they should be completely removed from the game. Is this even required? Nevertheless, closes #63 :P

### DIFF
--- a/mods/ctf_team_base/chest.lua
+++ b/mods/ctf_team_base/chest.lua
@@ -1,3 +1,6 @@
+local blacklist = {
+}
+
 local function max(a, b)
 	return (a > b) and a or b
 end
@@ -110,6 +113,12 @@ for _, chest_color in pairs(colors) do
 			if chest_color ~= ctf.player(player:get_player_name()).team then
 				minetest.chat_send_player(player:get_player_name(), "You're not on team " .. chest_color)
 				return 0
+			end
+
+			for itemstring in ipairs(blacklist) do
+				if stack:get_name() == itemstring then
+					return 0
+				end
 			end
 
 			local pstat = ctf_stats.player(player:get_player_name())


ClobberXD:
Adding all wooden tools, as they should ideally be converted into wood and sticks if they need to be stored. Should I also remove all saplings from `default`? @rubenwardy

mods/ctf_team_base/chest.lua
```diff
@@ -1,3 +1,15 @@
+local blacklist = {
+	"default:pick_wood",
+	"default:axe_wood",
+	"default:shovel_wood",
+	"default:sword_wood",
+	"default:leaves",
```
Thomas--S: Why are leaves blacklisted? They are a valid building material, e.g. usable as a counter measure against the loulicats strategy and thus quite important.

ClobberXD: This wasn't supposed to be merged at all, TBH. Initially, I had added leaves, wooden tools and the like, but then came e9f3da3, which made me rethink. You can see that I voiced my concerns against blacklisting leaves and wooden tools [here](https://github.com/rubenwardy/capturetheflag/pull/186#issuecomment-416029128)

Thomas--S: Then I suggest to at least (paritally) revert  this.

@rubenwardy What do you think?

PR #189, opened 2018-08-05
[Trivial] Enable crafting of powerful tools

ClobberXD:
- Allow crafting of axes and shovels from all ores.


ClobberXD:
Powerful shovels and axes are very useful at times, and they can't be compromised upon due to the lack of real estate.

MinetestSam:
2 pages isn't a lot though

ClobberXD:
If the order of registering crafts is preserved when displayed to the player, I could also re-order the registrations to ensure that the important crafts are on the first page.

MinetestSam:
Yeah things like ladders and stone things and furnace could be moved to the 2 page

PR #191, opened 2018-08-06
Add reinforced cobble

ClobberXD:
Is there anything left to do with this feature? Apart from being a very good material for building bases with, this would a good choice for the floor of xerox123's supermarket


AKryukov92:
I think that making just 1 block from 1 iron ingot is too few.
I'd rather spend 2 iron ingots to make steel sword rather than to make 2 blocks.
If you want to build something, you'll need much more blocks.
I think that recipe should contain 8 blocks of cobble + 1 iron ingot while result is 9 blocks of reinforced cobble.

ClobberXD:
But one ingot for 9 blocks sounds too less. Maybe like so?

`*` = `default:steel_ingot`
`#` = `default:cobble`

Crafting recipe: `4 steel ingots + 5 cobblestones = 9 reinforced-cobble`
```
# * #
* # *
# * #
```

AKryukov92:
2 steel swords still better than 9 blocks of reinforced cobble.

This is how 9 blocks look like in the game.
image

AKryukov92:
I'm not sure if reinforced cobble should be actually *reinforced* immediately after placing.
Currently there are tunnelers, which block passage behind them by simple cobble. It takes some time to break cobble and tunneler can run far away. The only cure for this situation is diamond pick.
Now we are adding reinforced cobble. It will be much harder than normal cobble and diamond pick won't help to catch tunneler.

ClobberXD:
Good concept, but I'm not sure if gravel is a good idea... I think we could use copper, tin, or bronze as they're largely unused anyway.

AKryukov92:
I am proposing gravel because you can find it on any map in meaningful quantities. Copper and tin is a bad option because they are present only on few maps.

Other option is to use dirt and cobble. IIRC, tunnelers are a problem only on the Cave map because there are no "surface" to travel from base to base. If there is no dirt in the Cave map, then it will just be impossible to craft reinforced cobble. Though, I don't like this option because we need to keep in mind additional implicit nuance about blocks allowed on the map.

ClobberXD:
Gravel isn't prevalent on the surface either - large quantities of gravel can only be obtained in Caverns. What if this does not depend on other craftitems at all. Instead, the craft recipe could consist only of cobble, so that reinforced cobble is just as craftable in all the maps.

AKryukov92:
> Instead, the craft recipe could consist only of cobble, so that reinforced cobble is just as craftable in all the maps.

Sounds fair.
I think cobblestone-only recipe should give less output than input. It is supposed to be special building block.
Something like
`#` - cobblestone
`0` - empty slot
`#0#`
`0#0`
`#0#`
to make 3 reinforced cobble. 

or
`##`
`##`
to make 1 reinforced cobble.

ClobberXD:
I like the second recipe better as it is more balanced, IMO.

ClobberXD:
Continued in #206

PR #194, opened 2018-08-22
Add support for transferring player stats

ClobberXD:
- Provide `/transfer_rankings <old_name> <new_name>` (requires `ctf_admin`).
- As of now, old player's stats is nullified. Should the stats be swapped instead?


mods/ctf_stats/gui.lua
```diff
@@ -305,7 +307,35 @@ minetest.register_chatcommand("reset_rankings", {
...
+		if not ctf_stats.player(src) then
+			return false, "Player '" .. src .. "' does not exist."
+		end
+		if not ctf_stats.player(dest) then
+			return false, "Player '" .. dest .. "' does not exist."
+		end
+
+		local pstat_src, _  = ctf_stats.player(src)
+		local pstat_dest, _ = ctf_stats.player(dest)
+		pstat_dest = pstat_src
```
rubenwardy: this doesn't work as you're setting a local and not a reference

mods/ctf_stats/gui.lua
```diff
@@ -305,7 +307,35 @@ minetest.register_chatcommand("reset_rankings", {
...
+local transfer_y = {}
+minetest.register_chatcommand("transfer_rankings", {
+	params = "<src> <dest>",
+	description = "Transfer rankings of one player to another.",
+	privs = {ctf_admin = true},
+	func = function(name, param)
+		if not param then
+			return false, "Invalid syntax. Provide source and destination player names."
+		end
+		local src, dest = param:trim():gmatch("([^"..sep.."]+)")
```
rubenwardy: sep isn't defined. Also, there's a pattern for player names already: `([%a%d_-]+) ([%a%d_-]+)`

mods/ctf_stats/gui.lua
ClobberXD: Thought so, but how do the rest of the stats-modification code work? They have a local `main` and `match`, modify them, and set `ctf.needs_save` to true, similar to what I've done here...

mods/ctf_stats/gui.lua
ClobberXD: Copy-pasted the code, sorry 'bout that. Dunno much about Lua patterns :/

Will use what you suggested, instead

ClobberXD:
Fixed the code. _Should_ work now, but still untested. I'm still unsure about what happens to the original player's stats - they're currently nullified after the transfer. Should the stats be swapped instead?

ClobberXD:
Fixed a ton of issues which surfaced during testing. Everything works just as expected now.

ClobberXD:
**BUMP**

ClobberXD:
**BUMP**

ClobberXD:
@rubenwardy ^

fjmng:
qwe_download
I did /transfer_rankings fjmng fjmng, and after doing /r I got this. You should make it so you cant transfer to yourself

ClobberXD:
@fjmng Please open a separate issue for this

PR #196, opened 2018-08-26
Remove gold-related nodes and craftitems

ClobberXD:
Removed items:
- `default:stone_with_gold`
- `default:gold_ingot`
- `default:gold_lump`

Alias registered to convert gold ore to plain stone.


ClobberXD:
AntiAbuse? :D

PR #198, opened 2018-08-30
Prevent damage instead of disabling PvP during buildtime

### LINKED ISSUE #59: Don't turn off PvP, block damage instead
This stops non-pvp being reported to server

ClobberXD:
Uses a `register_on_punchplayer` callback to achieve the same result.

**Warning: Untested**

closes #59

### DIFF
--- a/mods/ctf_match/buildtime.lua
+++ b/mods/ctf_match/buildtime.lua
@@ -62,10 +62,16 @@ minetest.register_globalstep(function(delta)
 	end
 end)
 
+minetest.register_on_punchplayer(function(_, hitter)
+	if ctf_match.is_in_build_time() then
+		minetest.chat_send_player(hitter:get_player_name(), "Match hasn't started yet!")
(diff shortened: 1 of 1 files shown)


mods/ctf_match/buildtime.lua
```diff
@@ -62,18 +62,23 @@ minetest.register_globalstep(function(delta)
 	end
 end)
 
+minetest.register_on_punchplayer(function(_, hitter)
+	if ctf_match.is_in_build_time() then
+		minetest.chat_send_player(hitter:get_player_name(), "Match hasn't started yet!")
+		return true
```
ClobberXD: Note that the default damage mechanism is disabled when `register_on_punchplayer` returns `true`. https://github.com/minetest/minetest/blob/7e21cec55bba7e5507cc84c2a53895429b812aff/doc/lua_api.txt#L3611

PR #201, opened 2018-09-02
Slow down auto-regen, add regen-on-kill

ClobberXD:
- Killer gains 6 HP on kill, if and only if kill is more than 10 nodes away from the victim's spawn-point.
- Auto-regen has been slowed down to encourage participation for faster healing.

No major scope for error code-wise, but requires testing with respect to playability and balance.


Gael-de-Sailly:
:-1: This makes harder to regen, wouldn't encourage players to get into uncertain fights (which is all the fun of CTF). Players would spend their time killing noobs to regen (as fighting a pro when you're at low health is basically doomed).

ClobberXD:
This would make players participate in the game, as they can only regen if they kill someone (ideally, this would also support kill assists). This would also foster team-work, where players with low health team up to take down a player with full health.

AKryukov92:
There should be some way to restore hitpoints without apples or suicides.
I think you should allow regeneration near spawn. Red regenerate around red spawn. Blues regenerate around blue spawn.

ClobberXD:
That would defeat the purpose of regen-on-kill which aims to make the player contribute to their team in order to regen. Even if a player is on low health, they can use strategy to successfully kill a couple of enemies, instead of approaching head-on. There's nothing that strategy can't achieve :)

AKryukov92:
You are too hardcore. I usually play builder and stick around own base. If you disable regen and will not enable it around the base, then I will just suicide by jumping from height to heal myself.

ClobberXD:
> I usually play builder and stick around own base

Nothing wrong with that. Quite a lot of action happens near the team-bases, and killing enemies is much easier since it's generally an all vs. one situation there.

AKryukov92:
I think that with disabled regeneration we will end up with suicides on team base just to "restore hitpoints".

ClobberXD:
Removing auto-regen is indeed an extreme step. I've restored auto-regen, but it will be much slower, in order to encourage players to participate in the game to regen _faster_.

Thomas--S:
Sorry, but my -1 will stay.

I think this feature encourages a very aggressive (and sword-wasting) gameplay, while disadvantaging rather defensive players.

Gael-de-Sailly:
Still :-1: as well. While I don't see why it encourages a more aggressive gameplay (if it's harder to regen I would less venture far from my base), it clearly encourages noobs-farming. Also, players from the same team will compete for the kills, this can be bad for team spirit.

AKryukov92:
Lets assume we have almost equal teams.
Blue player found steel sword. Red player found only stone ones.
Now if I play red, I can do 2 reckless attacks and maybe kill blue one. It is possible because I can take 70% of his hitpoints and attack again.
You are proposing the situation, in which blue player will regenerate 6 hp after my first attack. I will hardly kill him in this situation.

Regeneration on kill makes strong players much stronger. You can't actually *participate* if you have no weapon to kill enemy.

ClobberXD:
All the points put forward against regen-on-kill are very valid and reasonable. Closing in the interest of the community. Thanks for the feedback, again. :)

PR #202, opened 2018-09-04
[WIP] Reward player for kill streaks; post streak messages to kill-list

ClobberXD:
- Kills in rapid succession add towards player's kill streak if they're no more than 10 seconds after the previous kill.
- At streaks of 5, 10, and 20 kills, players receive kill streak rewards, and their streak is posted to the kill-list.
  - The awarded points is equal to `(2 * streak)`% of kill_reward (Very crude, will be modified to be more scale-able soon)

**WIP**

### DIFF
--- a/mods/ctf_events/init.lua
+++ b/mods/ctf_events/init.lua
@@ -1,9 +1,5 @@
 local hud = hudkit()
 
-minetest.register_on_leaveplayer(function(player)
-	hud.players[player:get_player_name()] = nil
-end)
-
 local NUM_EVT = 6
 
 ctf_events = {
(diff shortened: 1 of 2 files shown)


Thomas--S:
This might possibly encourage spawn killing.

ClobberXD:
In the current state, yes, but it should be noted that I'm not yet finished with this. I'm yet to add a couple of things including a check for killer's proximity to enemy base. Spawn-killing or not, kill streaks are way too easy near enemy base. ;)

ClobberXD:
This would be better if integrated with [awards](https://github.com/rubenwardy/capturetheflag/issues/17) whenever it is added. Closing

PR #210, opened 2018-10-21
Increase player movement speed

ClobberXD:
CTF is a fast-paced PvP game, and Minetest's default player speeds are too slow for this game in my opinion. This PR aims to improve player speed to a level which suits CTF.

Tested, requires feedback. Video showcase: https://youtu.be/yBNhx5J_V_4

### DIFF
--- a/minetest.conf
+++ b/minetest.conf
@@ -10,6 +10,17 @@ sprint_stamina = 5
 enable_lavacooling = false
 
 #
+# Player physics
+#
+
+movement_acceleration_default = 5
+movement_acceleration_air = 3
+movement_speed_walk = 6
+movement_speed_crouch = 2
+movement_speed_jump = 7
+movement_liquid_fluidity = 1.25
+
+#
 # CTF_PVP_ENGINE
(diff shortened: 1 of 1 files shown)


ClobberXD:
Here's the video showcasing this PR: https://youtu.be/YghgbS6T7HI

ClobberXD:
There was a lot of feedback on the previous video that it was too fast, so I tuned 'em down a bit, and used the default FOV (72) for video capture. Here's another video that reflects the current values: https://youtu.be/yBNhx5J_V_4

I had also forgotten about sneak speed, but that too has been taken care of, now...

AKryukov92:
Will players be able to do anything new with their new speed?
I hope they will not be able to run faster than arrow nor bullet.
Running on the slopes will probably feel different. Jumping from tree to tree too.

ClobberXD:
The proposed speeds will just make the player (and hence, the pace of the gameplay) feel a lot more brisk. The proposed values are just as proportional as they are now, and won't offer any major advantages.

tidle:
This seems good.

PR #213, opened 2018-10-28
Replace middle barrier with water if surrounded by water

ClobberXD:
`ind_glass_red` is replaced by water if surrounded by water on both sides. So therefore, `ind_glass_red` is also placed underwater, in map-maker mode. This allows #91 and other such issues to be fixed.

Tested; demonstration video: https://youtu.be/k3LbyINU70g

### DIFF
--- a/mods/ctf_map/barrier.lua
+++ b/mods/ctf_map/barrier.lua
@@ -4,7 +4,7 @@ local c_glass      = minetest.get_content_id("ctf_map:ind_glass")
 local c_glass_red  = minetest.get_content_id("ctf_map:ind_glass_red")
 local c_map_ignore = minetest.get_content_id("ctf_map:ignore")
 local c_actual_st  = minetest.get_content_id("default:stone")
--- local c_water      = minetest.get_content_id("default:water_source")
+local c_water      = minetest.get_content_id("default:water_source")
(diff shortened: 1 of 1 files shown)


DS-Minetest:
<details><summary>What if something like this happens?</summary>

```
w = water
l = land
a = air
B = barrier
L = land (with barrier above it)


wwwwwww
lwwwwww
llwwwww
lllwwww
lllwwww
llllwww
llllllw
lllllll

   ↓

wwwBwww
lwwBwww
llwBwww
lllBwww
lllBwww
lllLwww
lllLllw
lllLlll

   ↓

wwwwwww
lwwwwww
llwwwww
lllawww
lllawww
llllwww
llllllw
lllllll
```
</details>

<details><summary>

~~Wouldn't it make sense to just require one source near?~~</summary>

```
w = water
l = land
a = air
B = barrier
L = land (with barrier above it)

www
lwl
www

 ↓

wbw
lbl
wbw

 ↓

www
lal
www
```
</details>
What about adding a second type of barrier?

ClobberXD:
I don't see the need for a second node just for placing barriers underwater. `ind_glass_red` was created for the purpose of a middle barrier, and in my opinion, it should remain that way for the purpose of simplicity. Since there's only one node for both air and water, placing and removing barriers remains simple.

PR #216, opened 2018-11-01
Provide a large set of indestructible nodes with ctf_map

ClobberXD:
- This PR makes `ctf_map` provide most nodes from `default` (excluding grass and liquids) and all nodes from `stairs` and `wool` bundled with this (sub-)game, after making them indestructible. This means that map makers can add indestructible elements to their maps without the fear of them being broken down to the last node by noobs.
- When in map maker mode, players are provided a special admin pickaxe (`ctf_map:adminpick`), for breaking the indestructible nodes during map creation.

#### Use-cases

This would be very useful for all the artificial maps sans natural landscapes.
e.g. City-style maps with lots of buildings, which when destroyed, ruin the uniqueness of that map.

#### TODO

- Inventory image overlay to visually distinguish indestructible nodes from their default counterparts. I'm yet to figure out how to overlay images on `inventorycube`s.


ClobberXD:
As the commit message suggests, most nodes from `default` (excluding liquids and nodes with plantlike drawtype), and all nodes from `stairs` and `wool` have been added.

The reason I'm going with a manual approach w/ respect to node re-registration is because people might be using 5.0 to make maps, which adds a ton of new nodes (esp. in `default`). If they're re-registered as indestructible, map-makers would use them in maps, which in turn would be used in a 0.4.17-based subgame on a 0.4.17 server, causing a mass rebellion of Unknown Nodes.

ClobberXD:
Apart from that, I think this PR is ready-to-merge.

ClobberXD:
Needs updating w/ respect to the recent update of `stairs` to the latest version (#256)

ClobberXD:
Updated to include all nodes from `stairs`. Ready for review, @rubenwardy 

Here's what can be expected in testing:
- All non-liquids and non-plantlike nodes from `default`, all nodes from `stairs` and `wool` should be duplicated with the `ctf_map:` prefix, and they would also be indestructible.
- When in map-maker mode, players would be given `ctf_map:adminpick` on join, if they don't have one already.

PR #224, opened 2018-11-10
Add wool mod

ClobberXD:
Copy-pasted the latest version of the `wool` mod from `minetest_game/5.0-dev`, **sans the crafting recipe**, as wool is intended to be used for maps only.

Haven't tested, but I don't see anything out of place - it should work right off the bat.


AKryukov92:
Why did you add it?

ClobberXD:
Wool is a lightweight addition, which is a simple way to artificially add colour to the map.

AKryukov92:
I think I have got it. You add wool for map makers.

Issue #218, opened 2018-11-07
Preventing noclip


AKryukov92:
Today I have seen player, who was running through walls.
I think we can prevent so obvious hacking.
1) Periodically check position of player. If he programmaticaly spotted inside the wall and in report list, then kick him.
2) Periodically check coordinates of player. If he moves too fast, then kick him.

Thomas--S:
> Periodically check position of player. If he programmaticaly spotted inside the wall and in report list, then kick him.

Other players can build a wall where the player is standing, so the player is inside a wall without a noclip cheat. → No kick neccessary

Additionally, I'm sure these checks will introduce quite a bit of lag. I think it's not worth it.

afshaan4:
Or we could do it the way gaara said

>Only have the system kick or report if the player goes through a bunch of blocks, or if they go through 2 or 3 blocks in rapid succession like a "hacker" normally would.

AKryukov92:
How are you going to detect "whether player moves through 2 or 3 blocks in rapid succession"? I assume there are no lua callbacks for player movement.

> Other players can build a wall where the player is standing, so the player is inside a wall without a noclip cheat.

That is why I added "AND in report list" part to first part. If someone spotted this player to use noclip cheat, then he hopefully report him.
If we got him in wall programmaticaly, then he is very likely to be cheater.
To reduce false positives, we can do multiple checks every X second.

afshaan4:
>That is why I added "AND in report list" part to first part. If someone spotted this player to use noclip cheat, then he hopefully report him.

agreed

> How are you going to detect "whether player moves through 2 or 3 blocks in rapid succession"? I  
  assume there are no lua callbacks for player movement.

dunno, with a check running every few seconds?

Thomas--S:
> That is why I added "AND in report list" part to first part. If someone spotted this player to use noclip cheat, then he hopefully report him.

The report list is a really unreliable source. There are many players who report everybody who is better in PvP than them. Even moderators get reported quite often for "cheating".

Thomas--S:
> That's why there is "and player was actually spotted inside the wall" part and we should not kick for single false positive. Only after several checks over time.

But both of these indicators are prone to errors (I'd even say they are wrong most of the time.). So, even if both indicators are positive, the probability of a false positive is quite high.

> All features are lag-inducing. Should we stop any development because of this?

Not all features are equally lag-inducing.
[...]

PR #225, opened 2018-11-10
Add suffocation 

AKryukov92:
Suffocation damages players for 3 (like stone sword) if head is inside the wall
This is supposed to help with issue #218.

### DIFF
--- a/mods/suffocation/init.lua
+++ b/mods/suffocation/init.lua
@@ -1,0 +1,66 @@
+--[[
+	suffocation when head is inside solid node
+
+	all other features of playerplus was removed
+
+	PlayerPlus by TenPlus1, modified by AKryukov92
+	Original code can be downloaded here https://notabug.org/TenPlus1/playerplus
+]]
+
+-- get node but use fallback for nil or unknown
+local function node_ndef(pos)
+
+	local node = minetest.get_node_or_nil(pos)
+
+	local node_name = "air"
+	if node and minetest.registered_nodes[node.name] then
+		node_name = node.name
+	end
+
+	return minetest.registered_nodes[node_name]
+end
+
+local function is_normal_node(ndef)
+    return ndef.walkable == true
+       and ndef.drowning == 0
+       and ndef.damage_per_second <= 0
+       and ndef.groups.disable_suffocation ~= 1
+       and ndef.drawtype == "normal"
+end
+
+local time = 0
+
+minetest.register_globalstep(function(dtime)
+
+	time = time + dtime
+
+	-- every 1 second
+	if time < 1 then
+		return
+	end
+
+	-- reset time for next check
+	time = 0
+
+	local name, pos
+
+	-- loop through players
+	for _,player in ipairs(minetest.get_connected_players()) do
+
+		name = player:get_player_name()
+
+		pos = player:get_pos()
+		pos.y = pos.y + 1.4 -- head level
+
+		-- Is player suffocating inside a normal node without no_clip privs?
+		local ndef = node_ndef(pos)
+
+		if is_normal_node(ndef) and not minetest.check_player_privs(name, {noclip = true}) then
+			if player:get_hp() > 0 then
+				player:set_hp(player:get_hp() - 3)
+			end
+		end
+
+	end
+
+end)
--- a/mods/suffocation/depends.txt
+++ b/mods/suffocation/depends.txt
@@ -1,0 +1,1 @@
+default?


mods/suffocation/init.lua
```diff
@@ -0,0 +1,64 @@
...
+	suffocation when head is inside solid node
+	
+	all other features of playerplus was removed
+
+	PlayerPlus by TenPlus1, modified by AKryukov92
+	Original code can be downloaded here https://notabug.org/TenPlus1/playerplus
+]]
+
+-- get node but use fallback for nil or unknown
+local node_ndef = function(pos)
```
ClobberXD: The following syntax is preferred:

```suggestion
local function node_ndef(pos)
```

ClobberXD: `local` is missing. I assume this isn't intended to be a global function.

mods/suffocation/init.lua
```diff
@@ -0,0 +1,64 @@
...
+
+		-- Is player suffocating inside a normal node without no_clip privs?
+		local ndef = node_ndef(pos)
+
+		if ndef.walkable == true
+		and ndef.drowning == 0
+		and ndef.damage_per_second <= 0
+		and ndef.groups.disable_suffocation ~= 1
+		and ndef.drawtype == "normal"
+		and not minetest.check_player_privs(name, {noclip = true}) then
```
ClobberXD: Indent the conditions on newlines by two tab-spaces, and the actual condition body should be indented by one tab-space as usual. i.e.

```lua
if condition_a
		and condition_b
		and condition_c then
	print("condition body")
end
```

AKryukov92: I think it is better to make local function with meaningful name.

Thomas--S:
I think this is a bad idea because of this two reasons:

1) Lag will be a serious problem with this. A globalstep looping over all players every second and getting their positions, getting the nodes at this positions and two minetest.registered_nodes lookups per node will very likely introduce a noticeable amount of lag.

2) As written in #218: 
> This would "punish" players who are getting trapped by their team mates. Do you think this is right?
This will also probably not help with hackers.

Hackers usually also have increased spead, and so than can quickly fly through walls without getting very much damage. They can stay within solid nodes for 6 seconds (`20HP / 3HP/s = 6s`) before being killed and this time can be further increased with apples easily.

This feature will most likely only introduce a new possibility for team sabotage.

AKryukov92:
If a griefer comes to your server, then he will surely find some way to grief. He might just block you even without suffocation, which can lead to your death too.

If there is a valid reason to build where player stands, then you still can do it. Suffocation don't kill instantly and blocked player still able to move out of the wall.
Note that player gets damage only when his head is blocked. Placing blocks on the feet level will not change a thing.

The only way to introduce an additional large workload for moderators is to assemble a team of griefers. Suffocation is only gameplay feature, just like lava or apples. Feature itself will not generate griefing players.

LoneWolfHT:
Just add `damage_per_second = 3` to all types of stone and dirt_with_grass. No point in adding it to any other nodes.
And you could also create a 'ban chest' which bans the player that opens it. You could then place them out of reach outside of the barriers in maps, then a few noclippers might try to open it and get banned.
If you're worried about mobile users with the increased range bug then you can add a distance check

LoneWolfHT:
Why the thumbs down? It's rare for players to be placing stone on themselves, and you can't place dirt_with_grass.
It might even be possible to make cobble nodes smelt into a stone other than the current one that looks the same but doesn't have DPS

Thomas--S:
If I remember correctly, `damage_per_second` doesn't help against hackers (at least in 0.4.x) because the damage is calculated client side.

So, this whole change wouldn't be very useful.

AKryukov92:
> because the damage is calculated client side.

Really? It is weird. Do you have any proof of this?

LoneWolfHT:
> If I remember correctly, `damage_per_second` doesn't help against hackers (at least in 0.4.x) because the damage is calculated client side.
> 
> So, this whole change wouldn't be very useful.

I thought that was only for lava and drowning damage. Which was fixed btw.
Even if hackers can prevent the damage it's one more thing they'll have to figure out how to do. It would probably work on most of the hacked clients you can download off the internet rn too

LoneWolfHT:
> Just add damage_per_second = 3 to all types of stone and dirt_with_grass. No point in adding it to any other nodes

If wanted I could create a PR to this PR to change to the method I suggested above. Or I could create a completely new PR

AKryukov92:
I think it is better to close this PR.

PR #231, opened 2018-11-11
Add hacky fix for incorrect door colorization on_place

ClobberXD:
See https://github.com/rubenwardy/capturetheflag/issues/230#issuecomment-437655391

**Current colorization logic**: Append `tname` if `name == "doors:door_steel"`

**Proposed colorization logic**: Append `tname` if `name` **contains** `"doors:door_steel"`. This way, `doors:door_steel_red` and `doors:door_steel_blue` will also be considered for colorization on_place in addition to plain steel doors)

This is a hacky solution, meant to be a temporary fix until the root cause has been taken care of. Tested; Video Demonstration: https://youtu.be/efq8svmD57c

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -300,7 +300,12 @@ function doors.register(name, def)
 			end
 
 			-- Get placer's team
-			local tname = ctf.player(pn).team or ""
+			local tname = ctf.player(pn).team
+
+			-- If steel doors, append tname to place coloured team-doors instead
+			if name:find("doors:door_steel") then
+				name = "doors:door_steel_" .. tname -- e.g. "doors:door_steel_red"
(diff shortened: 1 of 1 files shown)


AKryukov92:
Your solution actually works because `"doors:door_steel_red":find("doors:door_steel")` will return 1 (=true) while `"doors:door_steel_red" == "doors:door_steel" will be false.
Though, you have violated variable `name` because its scope wider than function `on_place` has.
Two different calls of `on_place` function operate with the same variable `name`. Assignment to that variable affects all following calls to `on_place`.

ClobberXD:
Yes, I know. As I said this is a hacky solution. But now that there's #233, I'll close this. Thanks.

PR #234, opened 2018-11-11
Improve rankings formspec

ClobberXD:
- Renamed `render_per_team_stats` to `render_team_stats`, for the immaculate intention of shortening lines without adding a line-break.
- Adds fixed minimum width to columns to space them out.
- Displays a player's rankings at the bottom of the formspec without having to scroll down. (if the player isn't in the top 50)
  - This is achieved using a very hacky method: Another table is required, with the same column widths as the main table. This is the reason why I added fixed width to columns in the main table. :)

screenshot_20181111_223654
screenshot_20181111_224301

@Gael-de-Sailly You specifically wanted this feature. Please do take a look at it. I'm afraid this was the best solution I could come up with. :)

### DIFF
--- a/mods/ctf_stats/gui.lua
+++ b/mods/ctf_stats/gui.lua
@@ -1,13 +1,30 @@
 local storage = minetest.get_mod_storage()
 local prev_match_summary = storage:get_string("prev_match_summary")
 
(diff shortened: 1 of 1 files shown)


Coder12a:
I like this new rankings formspec.

PR #245, opened 2018-11-15
Fix EOM shown before winner_player is initialized

ClobberXD:
#### Issue

Players reported that the name of the flag-capturer wasn't shown in the End-of-Match formspec.
issue

#### Cause

The `register_on_winner` callbacks are called from a `register_on_capture` in `ctf_match/matches.lua` after checking that there are no other teams left. But `winner_player` is also initialized in another `register_on_capture` callback in `ctf_stats/init.lua`. I'm guessing the callback in `ctf_match/matches.lua` is always executed first, because `ctf_stats` has a hard dependency on `ctf_match`. So the EOM formspec is shown before the winner player can be initialized, resulting in a `-` being shown.

#### Solution (albeit a hacky one)

The fix is to just insert the callback in `ctf_stats/init.lua` at the beginning of `ctf_flag.registered_on_capture` instead of appending it to the end. Tested.

screenshot_20181115_144849

### DIFF
--- a/mods/ctf_stats/init.lua
+++ b/mods/ctf_stats/init.lua
@@ -142,7 +142,7 @@ end)
 local winner_team = "-"
 local winner_player = "-"
 
-ctf_flag.register_on_capture(function(name, flag)
+table.insert(ctf_flag.registered_on_capture, 1, function(name, flag)
 	local main, match = ctf_stats.player(name)
 	if main and match then
 		main.captures  = main.captures  + 1
@@ -162,7 +162,7 @@ ctf_match.register_on_winner(function(winner)
 
 	-- Show match summary
 	local fs = ctf_stats.get_formspec_match_summary(ctf_stats.current,
-					winner_team, winner_player, os.time()-ctf_stats.start)
+					winner_team, winner_player, os.time() - ctf_stats.start)
 	for _, player in pairs(minetest.get_connected_players()) do
 		minetest.show_formspec(player:get_player_name(), "ctf_stats:eom", fs)
 	end


mods/ctf_stats/init.lua
```diff
@@ -142,7 +142,7 @@ end)
 local winner_team = "-"
 local winner_player = "-"
 
-ctf_flag.register_on_capture(function(name, flag)
+table.insert(ctf_flag.registered_on_capture, 1, function(name, flag)
```
AKryukov92: I assume that you insert this function at specified index to make sure that this function will be called after some other function.
Are you sure, that *some other function* will always be on the same index?

I think it is better to initialize both functions now, but register them later?
something like
```
function show_summary(name,flag)
end
function something_else()
end
ctf_flag.register_on_capture(show_summary)
ctf_flag.register_on_capture(something_else)
```

ClobberXD: > I assume that you insert this function at specified index to make sure that this function will be called after some other function.

I'm inserting this into the _very first index_, so that it's called _before_ all other callbacks.

As I've explained in the first post, the callback that checks for winner and initializes `winner_player` is in another `register_on_capture` in `ctf_match/matches.lua`, which is loaded before `ctf_stats` because `ctf_stats` depends on `ctf_match`. This PR would ensure that `winner_player` is initialized before the EOM formspec is shown.

AKryukov92: I forgot that lua array are indexed ftom 1.

AKryukov92:
I tested it. Flag carrier was displayed in the match summary screen.

PR #247, opened 2018-11-16
ctf_map: Misc. improvements to map maker

ClobberXD:
- Add a "Reset" button that resets all fields in map maker formspec, and also removes all flags placed.
- Convert the individual working set of vars into a table `config`.
- Add another sister table `defaults` which helps with `reset`.
- Add a check to abort schem export if barriers not placed.
- Disallow placement of more than 2 flags.

### DIFF
--- a/mods/ctf_map/map_maker.lua
+++ b/mods/ctf_map/map_maker.lua
@@ -1,37 +1,51 @@
+
+local storage = minetest.get_mod_storage()
 local randint = math.random(100)
+local defaults = {
+	mapname = "ctf_" .. randint,
+	mapauthor = nil,
+	maptitle = "Untitled Map " .. randint,
+	mapinitial = "",
+	barrier_r = 110,
+	barrier_rot = 0,
+	center = { x = 0, y = 0, z = 0, r = 115, h = 140 },
+	flags = {}
+}
 
 -- Reload mapmaker context from mod_storage if it exists
-local storage = minetest.get_mod_storage()
-local mapname = storage:get_string("mapname")
-local maptitle = storage:get_string("maptitle")
-local mapauthor = storage:get_string("mapauthor")
-local mapinitial = storage:get_string("mapinitial")
-local center = storage:get_string("center")
-local flag_positions = storage:get_string("flags")
-local barrier_r = storage:get_int("barrier_r")
-local center_barrier_rot = storage:get_int("center_barrier_rot")
+local config = {
+	mapname = storage:get_string("mapname"),
+	maptitle = storage:get_string("maptitle"),
+	mapauthor = storage:get_string("mapauthor"),
+	mapinitial = storage:get_string("mapinitial"),
+	center = storage:get_string("center"),
+	flags = storage:get_string("flags"),
+	barrier_r = storage:get_int("barrier_r"),
+	barrier_rot = storage:get_int("barrier_rot"),
+	barriers_placed = storage:get_int("barriers_placed") == 1
+}
 
-if mapname == "" then
-	mapname = "ctf_" .. randint
+if config.mapname == "" then
+	config.mapname = defaults.mapname
 end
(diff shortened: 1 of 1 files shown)


mods/ctf_map/map_maker.lua
```diff
@@ -371,9 +390,20 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
...
 			end
 		end)
+
+		reset()
 		return
 	end
 
+	if fields.reset then
+		for _, pos in pairs(flag_positions) do
+			minetest.set_node(pos, {name = "air"})
```
SmallJoker: `remove_node` is better in this case because the map might use any kind of air

mods/ctf_map/map_maker.lua
```diff
@@ -7,6 +7,7 @@ local maptitle = storage:get("maptitle") or "Untitled Map " .. randint
 local mapauthor = storage:get("mapauthor")
 local mapinitial = storage:get_string("mapinitial")
 local center_barrier_rot = storage:get_int("center_barrier_rot")
+local barriers_placed = storage:get_int("barriers_placed") == 1 and true
```
SmallJoker: `and true` is superfluous

mods/ctf_map/map_maker.lua
```diff
@@ -219,6 +221,26 @@ local function emerge_progress(ctx)
...
 
+local function reset()
+	mapname = "ctf_" .. randint
+	maptitle = "Untitled Map " .. randint
+	mapinitial = ""
+	center_barrier_rot = 0
+	barriers_placed = false
+	barrier_r = 110
+	center = { x = 0, y = 0, z = 0, r = 115, h = 140 }
+	flag_positions = {}
```
SmallJoker: Duplicated values are not that nice, especially when they're 200 lines further up in the code

mods/ctf_map/map_maker.lua
ClobberXD: I was thinking of having this as a boolean var. Does a value of 0 evaluate `barriers_placed` to false in an `if` statement?

mods/ctf_map/map_maker.lua
ClobberXD: Ok. Added `remove_node` in place of `set_node` and `{name = "air"}`, thanks.

mods/ctf_map/map_maker.lua
AKryukov92: 0 == 1 -> false

ClobberXD: Just checked, `barriers_placed` (= 0) still evaluates to true. I guess I'll have to add an explicit check against zero...

ClobberXD: Oh I understand now. SmallJoker suggested the exact same thing...

mods/ctf_map/map_maker.lua
ClobberXD: Still wondering what should be done here...

SmallJoker: Best solution (and the one with the most changed lines) would be to create a table on the very top, containing the defaults for each setting. Then you could just read the storage value by the default's type into a new "config" table.
Here in the reset function it would basically be a table copy call.

ClobberXD: Finally got the time to work on this. All the variables in the working set are now a part of the `config` table, and the defaults are defined in another table called `defaults`. The function `reset` is very simple now, thanks to the new tables. Hopefully, I've also touched as many lines as humanely possible... 8)

ClobberXD:
Ready for testing.

ClobberXD:
Way too many changes. I'm splitting this PR into two separate PRs:

- One for general cleanup, adding `config` and `defaults` table.
- And then another one for the reset feature.

PR #256, opened 2018-11-19
Update `stairs` to latest MTG version

ClobberXD:
The 5.0 seems to add inner and outer stairs, which would be useful for maps and the ind_nodes PR.

Maps using stairs from 0.4-stable should still work, as the 5.0 version just adds more to the existing set of stairs and slabs.


AKryukov92:
Are stairs ever available in CTF? I didn't see them for a long time.

AKryukov92:
CTF is not survival. Why do we ever need to craft stairs?

ClobberXD:
For maps. More specifically, maps with artificial structures like buildings.

AKryukov92:
Do you have some specific map in mind?
It is easy to add all blocks from all mods to CTF "for maps", but if we don't make maps with them, then they will be useless.

ClobberXD:
CTF already has `stairs` mod, but just the version from MTG 0.4.x.

PR #260, opened 2018-11-20
Modify ammo recipe and add a new recipe that uses tin

LoneWolfHT:
5 steel is a bit too much IMO, using tin may not be realistic but it gives tin some value.


AKryukov92:
It will make them uncraftable on "the bridge" and "nostalgia" maps.

LoneWolfHT:
> It will make them uncraftable on "the bridge" and "nostalgia" maps.

Oh really :thinking: 
What if I added a recipe with 3 steel and 2 coal then?

LoneWolfHT:
> What if I added a recipe with 3 steel and 2 coal then?

Done

mods/ctf_crafting/init.lua
```diff
@@ -159,7 +159,15 @@ crafting.register_recipe({
...
 	level  = 1,
 })
+
+crafting.register_recipe({
+	type   = "inv",
+	output = "shooter:ammo 1",
+	items  = { "default:steel_ingot 3", "default:coal_lump 2" },
+	always_known = true,
+	level  = 1,
+})
```
ClobberXD: Please add a newline at the end of the file.

LoneWolfHT:
Huh, didn't seem to work. Will do tomorrow when I can get back on my computer

ClobberXD:
It worked. The Travis CI build passed too. The newline is just not displayed in the Github diff.

ClobberXD:
Maybe `shooter:ammo 2` as output instead of 1 for both the recipes?

AKryukov92:
I'd rather be careful and add underpowered recipes rather than overpowered. Thats why I support 3+2=1 recipe.

PR #263, opened 2018-11-21
Eliminate shooter's dependency on tnt_explode.ogg

ClobberXD:
Apparently, grenade explosions use `tnt`'s explosion sound, but since CTF does not have `tnt` yet, no sound is played. This PR copy-pastes the sound file into the `shooter` mod itself, to eliminate this missing dependency.


ClobberXD:
Clarification: This PR just adds a missing audio file :)

@rubenwardy

PR #273, opened 2018-11-26
Remove unused arg from ctf_colors.get_color and ctf_colors.get_irc_color

ClobberXD:
Follow-up PR to MT-CTF/ctf_pvp_engine#25


ClobberXD:
Once MT-CTF/ctf_pvp_engine#25 gets merged, I'll update the sub-module to the latest commit, after which this PR can be merged.

Issue #264, opened 2018-11-21
Add a grenades mod


LoneWolfHT:
Would be cool to have a few more grenades. Like a flash bang or smoke grenade. Maybe even a few made up ones.
I've started a mod I'll be releasing on the forums. But it would probably work here too. Thoughts?

AKryukov92:
I don't like proposal about "more of something".
What exactly do you have in mind?

LoneWolfHT:
> I don't like proposal about "more of something".
> What exactly do you have in mind?

Not sure yet. Was thinking people could post suggestions here. Popular ideas are a smoke and flash bang grenade. What I'm making is an API and one basic grenade. More grenades could be added

AKryukov92:
I think people always can post suggestion here. By "here" I mean list of issues in this repository.
Github is not forum. There is no need for "mega" threads about vague ideas.
I think it will be better if you create one issue about flashbang grenade and other - about smoke grenade. Of course these threads should not be empty. You should write why do you propose it and how do you imagine their mechanic in detail.

LoneWolfHT:
I'll close the issue once the API is done or if it gets a lot of negative feedback and create a PR. For now it'll be for suggestions

ClobberXD:
Please don't make one issue for each type of grenade. One issue would suffice.

LoneWolfHT:
>Please don't make one issue for each type of grenade. One issue would suffice.

I'll only have this issue and a PR later for grenades

AKryukov92:
> I'll only have this issue and a PR later for grenades

PRs might not be merged because rubenwardy might not like those features. I think it is better to discuss things first and get approve on idea before doing any work.

LoneWolfHT:
> > I'll only have this issue and a PR later for grenades
> 
> PRs might not be merged because rubenwardy might not like those features. I think it is better to discuss things first and get approve on idea before doing any work.

If PRs won't be merged then said PRs will be changed. This issue is for the discussion of grenades.

LoneWolfHT:
May leave the flashbang until I can sort out the viewing problems. How are the other two grenades though? https://github.com/LoneWolfHT/grenades

LoneWolfHT:
Yeah I was going to start a PR but I had some trouble with getting the game to work. I'm improving the code rn though. So thought I might as well wait until that's finished.

LoneWolfHT:
Maybe test the grenade and the API now. Improved it a little @ClobberXD

[...]

PR #279, opened 2018-11-27
Add the grenades mod

LoneWolfHT:
https://github.com/MT-CTF/capturetheflag/issues/264

Could probably use testing by someone other than me. These grenades are also craftable (5 steel ingots + 1-2 coal lumps)

**Todo:**
- [ ] Make Flashbang slow down players
- [x] Add sounds


LoneWolfHT:
I'm not sure how to fix the deathlist HUD bug. Could someone point me to the file where the shooter grenade was set to show up as a grenade img and not a sword?

ClobberXD:
Embarrassingly, this is hard-coded in `registered_on_killedplayer` callback in the `ctf_pvp_engine` sub-module. Fixing this would need another PR in https://github.com/MT-CTF/ctf_pvp_engine. Here's the relevant code:

https://github.com/MT-CTF/ctf_pvp_engine/blob/e2a59cef6eba80464af5414d89c4e1c6fbfd597c/ctf/teams.lua#L472-L492
```lua
		if player:get_hp() - damage <= 0 then
			dead_players[player:get_player_name()] = true
			local wielded = hitter:get_wielded_item()
			local wname = wielded:get_name()
			print(wname)
			local type = "sword"
			if wname:sub(1, 8) == "shooter:" then
				if wname == "shooter:grenade" then
					type = "grenade"
				else
					type = "bullet"
				end
			end
			if tool_capabilities.damage_groups.grenade then
				type = "grenade"
			end
			for i = 1, #ctf.registered_on_killedplayer do
				ctf.registered_on_killedplayer[i](player:get_player_name(), hitter:get_player_name(), type)
			end
			return false
		end
```

ClobberXD:
I've attempted at making the kill type a little more generic: MT-CTF/ctf_pvp_engine#26

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+                minsize = 7,
+                maxsize = 10,
+                collisiondetection = true,
+                collision_removal = false,
+                vertical = false,
+                texture = "grenades_smoke.png",
+            })
+
+            for k, v in ipairs(minetest.get_objects_inside_radius(pos, radius)) do
+                if v:is_player() and v:get_hp() > 0 then
```
ClobberXD: Players should not take damage if they're in the same team as the thrower. At least that's how grenades work currently. You could achieve this like so:

```lua
local vteam = ctf.player(v:get_player_name()).team
local pteam = ctf.player(name).team
if vteam ~= pteam then
	v:punch( ... )
end
```

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+                    local grenadedir = vector.round(vector.direction(v:get_pos(), pos))
+                    local pname = v:get_player_name()
+
+                    if playerdir.x == grenadedir.x and playerdir.z == grenadedir.z then
+                        for i = 1, 3, 1 do
+                            local key = v:hud_add({
+                                hud_elem_type = "image",
+                                position = {x=0, 0},
+                                name = "death_list_hud",
+                                scale = {x=1000, y=1000},
```
ClobberXD: `scale.x` and `scale.y` can also take negative values, in which case they are considered to be the percentage of screen width and screen height respectively. See https://github.com/minetest/minetest/blob/9519d57017cfa28eb4f2fbf3c780f826017bbb00/doc/lua_api.txt#L1255

So this should be `scale = {x = -100, y = -100},`

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+            end
+        end,
+    })
+end
+
+-- Smoke Grenade
+
+if not smoke or smoke == true then
+    grenades.register_grenade("smoke", {
+        description = "A smoke grenade (Generates a lot of smoke around the detonation area)",
```
ClobberXD: The item descriptions need to be simple and concise.

e.g. `Smoke grenade (generates a lot of smoke at blast area)` or something similar. A little bit of grammar could also be compromised here and there (if required).

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+
+    minetest.register_craftitem("grenades:gun_powder", {
+        description = "A dark powder used for crafting some grenades",
+        inventory_image = "grenades_gun_powder.png"
+    })
+
+    minetest.register_craft({
+        type = "shapeless",
+        output = "grenades:gun_powder",
+        recipe = {"default:coal_lump", "default:coal_lump", "default:coal_lump", "default:coal_lump"},
```
ClobberXD: Since the recipe type is shapeless, this can just be `recipe = {"default:coal_lump 4"}`

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+    })
+end
```
ClobberXD: Wherever this icon shows up, it means that a newline is missing at EOF. Please ensure that there is a newline at the end of all the files in the mod.

mods/grenades/init.lua
```diff
@@ -0,0 +1,127 @@
+grenades = {}
+
+local function throw_grenade(name, player)
+    local dir = player:get_look_dir()
+    local pos = player:get_pos()
+    local obj = minetest.add_entity({x=pos.x+dir.x, y=pos.y+1.3, z=pos.z+dir.z}, name)
+    local yaw = player:get_look_yaw()
+
+    obj:setvelocity({x=dir.x * 16, y=dir.y * 16, z=dir.z * 16})
+    obj:setacceleration({x=dir.x * -3, y=-17, z=dir.z * -3})
```
ClobberXD: Inconsistent spacing in the two lines above. Spaces around `=` is necessary too.

mods/grenades/init.lua
```diff
@@ -0,0 +1,127 @@
...
+    obj:setyaw(yaw + math.pi)
+
+    return(obj:get_luaentity())
+end
+
+function grenades.register_grenade(name, def)
+    if not def.timeout then def.timeout = 5 end
```
ClobberXD:
```lua
if not def.timeout then
    def.timeout = 5
end
```

mods/grenades/init.lua
```diff
@@ -0,0 +1,127 @@
...
+            groups = {oddly_breakable_by_hand = 2},
+            on_use = function(itemstack, user, pointed_thing)
+                local player_name = user:get_player_name()
+
+                if pointed_thing.type ~= "node" then
+                    local grenade = throw_grenade("grenades:grenade_"..name, user)
+                    grenade.timer = 0
+                    grenade.thrower_name = player_name
+
+                    if not minetest.setting_getbool("creative_mode") then
```
ClobberXD: Deprecated. Use `minetest.settings:get_bool("creative_mode")` instead.

mods/grenades/init.lua
```diff
@@ -0,0 +1,127 @@
...
+                        itemstack = ""
+                    end
+                end
+
+                return itemstack
+            end
+        })
+    end
+end
+
```
ClobberXD: The craftitem def and nodedef is largely the same, so what can be done here is to pre-assign the def to a local, and add/remove a couple of fields within the if-else blocks before registration.

mods/grenades/settingtypes.txt
```diff
@@ -0,0 +1,11 @@
...
+enable_regular_grenade (Add a regular grenade to the world) bool true
+
+#Add a flashbang grenade
+enable_flashbang_grenade (Add a flashbang grenade to the world) bool true
+
+#Add a smoke grenade
+enable_smoke_grenade (Add a smoke grenade to the world) bool true
+
+#Enable Recipes
+enable_grenade_recipes (Gives enabled grenades cafting recipes) bool true
```
ClobberXD: It's advisable to prefix mod-specific settings by mod name. e.g. `grenades.enable_flashbang`, `grenades.enable_recipes`

ClobberXD: Also, the last setting could be `grenades.enable_crafting (Enables crafting of grenades) bool true`

AKryukov92:
Please describe features, that you expect to introduce with grenades.
What exactly should happen on use of each grenade?
When I will do testing, I need to know, what is bug and what is intented behaviour.

ClobberXD:
> Please describe features, that you expect to introduce with grenades.

Good point. Preferably, this should be added to the first post...

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+        image = "grenades_flashbang.png",
+        on_explode = function(pos, name)
+            for k, v in ipairs(minetest.get_objects_inside_radius(pos, 15)) do
+                if v:is_player() and v:get_hp() > 0 then
+                    local playerdir = vector.round(v:get_look_dir())
+                    local grenadedir = vector.round(vector.direction(v:get_pos(), pos))
+                    local pname = v:get_player_name()
+
+                    if playerdir.x == grenadedir.x and playerdir.z == grenadedir.z then
+                        for i = 1, 3, 1 do
```
ClobberXD: Why are there 3 textures for this effect? Wouldn't one texture suffice?

mods/grenades/grenades.lua
ClobberXD: Or... does this invoke all the `registered_on_punchplayer` callbacks? In that case, no need for the above.

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+                    v:punch(player, 2, {damage_groups = {fleshy = 22-vector.distance(pos, v:get_pos())}}, nil)
+                end
+            end
+        end,
+    })
+end
+
+-- Flashbang Grenade
+
+if flash == true then
```
ClobberXD: `if not flash or flash == true then`

Or just replace all instances with `if flash ~= false then`

ClobberXD:
Tested. Here are my thoughts, from the most important  to the least important:

- Grenades follow the same trajectory irrespective of player's rotation in the y-axis.
  - Personally, this would be a huge game-breaker for me, as I won't be able to "aim" with grenades - they'll just fall a couple of nodes ahead of me, however high up I look.
  - Players would expect grenades to land further away when they look higher up.
  - Not sure, but I think `shooter` sets the velocity/acceleration based on how high up the player is looking.
  - Grenades should also be faster in general.
- Damage radius of regular grenade is too less, and does the same damage to all objects in range irrespective of their distance to it. Same applies to flashbangs.
  - Damage radius should be around 5-6 nodes, and the players should take lesser damage, the farther away they are.
- I'm quite happy with smoke grenades, but the "smoke" causes a huge hit in my FPS.
  - Particles in MT are currently poorly optimized, and lavish use of particles will heavily tax the graphics processor.
  - As a trial, the individual particles could be made much bigger, and the total number of particles being displayed at the same time could be decreased. This will provide the same effect, but will go a little easy on the GPU.
  - The particle textures could also be made a little rounder to look nicer.
- The fading-out effect of flashbangs could be improved.
  - It feels like there are only two overlays instead of three. This could _probably_ be a result of my high gamma.
  - `on_explode` of flashbang adds three white HUD overlays of various intensities, and gradually removes them one-by-one with the strongest one removed first.
  - I'm sure the same effect can be achieved using a single texture, with the help of texture modifiers.

mods/ctf_crafting/init.lua
```diff
@@ -171,3 +171,19 @@ crafting.register_recipe({
...
+	output = "grenades:grenade_regular 1",
+	items  = { "default:steel_ingot 5", "default:coal_lump 1" },
+	always_known = true,
+	level  = 1,
+})
+
+crafting.register_recipe({
+	type   = "inv",
+	output = "grenades:grenade_smoke 1",
+	items  = { "default:steel_ingot 5", "default:coal_lump 2" },
```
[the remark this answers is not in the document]
LoneWolfHT: 5 steel is worth 2 swords. One throw and you've wasted that much steel. The most steel I've seen in a match is about 40-50. That's only 7-9 grenades.
I could probably increase the coal amount by 2 though?

mods/grenades/.luacheckrc
```diff
@@ -0,0 +1,18 @@
+unused_args = false
```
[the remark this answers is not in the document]
LoneWolfHT: Ok. Will remove

mods/grenades/API.md
```diff
@@ -0,0 +1,24 @@
+#Grenades API
+
+Still WIP. Please suggest new features here: https://forum.minetest.net/viewtopic.php?f=9&t=21466
+
+##API
+
+	`grenades.register_grenade("name", { -- Name of the grenade (Like 'smoke' or 'flashbang')
```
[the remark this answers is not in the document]
LoneWolfHT: Will remove then

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
+local settings = minetest.settings
+
+local regular = settings:get_bool("enable_regular_grenade")
+local flash = settings:get_bool("enable_flashbang_grenade")
+local smoke = settings:get_bool("enable_smoke_grenade")
```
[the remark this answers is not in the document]
LoneWolfHT: They are enabled by default. So are the smoke grenades. It's defined elsewhere

mods/grenades/init.lua
LoneWolfHT: ok

mods/grenades/init.lua
LoneWolfHT: Will change

mods/grenades/init.lua
LoneWolfHT: lol. Not sure how I did that. Will fix

mods/grenades/init.lua
LoneWolfHT: Ok. Will look into that

mods/grenades/settingtypes.txt
LoneWolfHT: Ok. This was my first mod with settings. Will fix

mods/grenades/grenades.lua
LoneWolfHT: Adds a fading effect. I could probably improve the textures a little

mods/grenades/grenades.lua
LoneWolfHT: I was going to disable the flashbang until I could fix the code that checks to see if the player is looking in the flashbang's direction

mods/grenades/grenades.lua
LoneWolfHT: > Or... does this invoke all the `registered_on_punchplayer` callbacks? In that case, no need for the above.

90% sure it does invoke the callbacks

mods/grenades/grenades.lua
LoneWolfHT: Oh. I knew there was a better way to do it (Kind of sad too. I'm always consulting the lua_api.txt)

mods/grenades/grenades.lua
LoneWolfHT: I think I've fixed that now

mods/grenades/grenades.lua
LoneWolfHT: Oh? Didn't know that was supported

mods/grenades/grenades.lua
LoneWolfHT: Ok

mods/grenades/API.md
```diff
@@ -0,0 +1,24 @@
...
+		placeable = false, -- Optional, default is false
+		timeout = 5, -- Optional, default is 5
+		particle = { -- Adds particles in the grenade's trail
+		    image = "grenades_smoke.png", -- The particle's image
+		    life = 1, -- How long (seconds) it takes for the particle to disappear
+		    size = 4, -- Size of the particle
+		    glow = 0, -- Brightens the texture in darkness
+		    interval = 5, -- How long it takes before a particle can be added
+		}
+	})`
```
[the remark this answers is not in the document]
LoneWolfHT: Removing...

mods/grenades/grenades.lua
LoneWolfHT: I was disabling the flashbangs by default because they aren't perfect yet

mods/grenades/grenades.lua
LoneWolfHT: Changed it to use a texture modifier

LoneWolfHT:
> Tested. Here are my thoughts, from the most important to the least important:
> 
>     * Grenades follow the same trajectory irrespective of player's rotation in the y-axis.
>       
>       * Personally, this would be a huge game-breaker for me, as I won't be able to "aim" with grenades - they'll just fall a couple of nodes ahead of me, however high up I look.
>       * Players would expect grenades to land further away when they look higher up.
>       * Not sure, but I think `shooter` sets the velocity/acceleration based on how high up the player is looking.

My throwing code was basically copied from the shooter mod's grenade and modified.

>       * Grenades should also be faster in general.
> 
>     * Damage radius of regular grenade is too less, and does the same damage to all objects in range irrespective of their distance to it. Same applies to flashbangs.

I increased the radius of regular grenades. They already do one less hp point per node away the player is
    
>       * Damage radius should be around 5-6 nodes, and the players should take lesser damage, the farther away they are.
> 
>     * I'm quite happy with smoke grenades, but the "smoke" causes a huge hit in my FPS.
>       
>       * Particles in MT are currently poorly optimized, and lavish use of particles will heavily tax the graphics processor.
>       * As a trial, the individual particles could be made much bigger, and the total number of particles being displayed at the same time could be decreased. This will provide the same effect, but will go a little easy on the GPU.

You need lots of particles for it to do anything. I lessened the amount. (Now 15. IIRC I had put 100 before)
I, with my potato. Notice hardly any drop in FPS with multiple smoke grenades set off at once though

>       * The particle textures could also be made a little rounder to look nicer.
> 
>     * The fading-out effect of flashbangs could be improved.
>       
>       * It feels like there are only two overlays instead of three. This could _probably_ be a result of my high gamma.
>       * `on_explode` of flashbang adds three white HUD overlays of various intensities, and gradually removes them one-by-one with the strongest one removed first.
>       * I'm sure the same effect can be achieved using a single texture, with the help of texture modifiers.

I've made some changes to the flashbang. It uses a single texture now, opacity changes may still need some help, and I also set it up so it's enabled by default now

mods/grenades/grenades.lua
LoneWolfHT: flashbang is enabled by default now

LoneWolfHT:
> Please describe features, that you expect to introduce with grenades.
> What exactly should happen on use of each grenade?
> When I will do testing, I need to know, what is bug and what is intented behaviour.

Just mention anything you think I did wrong or you don't like. My 'intended behavior' is grenades that almost everyone likes

LoneWolfHT:
@ClobberXD Could you review/test again?

mods/grenades/API.md
```diff
@@ -4,7 +4,7 @@ Still WIP. Please suggest new features here: https://forum.minetest.net/viewtopi
 
 ##API
```
ClobberXD: These are rendered as plain-text without spaces between `#` and the first word. This, for example, should be `## API`

mods/grenades/init.lua
```diff
@@ -1,127 +1,112 @@
...
+	local dir = player:get_look_dir()
+	local pos = player:get_pos()
+	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.4, z = pos.z + dir.z}, name)
+	local yaw = player:get_look_yaw()
 
-    obj:setvelocity({x=dir.x * 16, y=dir.y * 16, z=dir.z * 16})
-    obj:setacceleration({x=dir.x * -3, y=-17, z=dir.z * -3})
-    obj:setyaw(yaw + math.pi)
+	obj:setvelocity({x = dir.x * 20, y = dir.y * 20, z = dir.z * 20})
+	obj:setacceleration({x = dir.x * -4, y = -20, z = dir.z * -4})
```
ClobberXD: Looks like `obj:setyaw()` has been omitted. Also, please use `set_yaw` instead.

mods/grenades/API.md
LoneWolfHT: ~~I love markdown~~. Will fix. Thanks!

mods/grenades/grenades.lua
```diff
@@ -1,184 +1,188 @@
...
-    })
-
-    minetest.register_craft({
-        type = "shapeless",
-        output = "grenades:gun_powder",
-        recipe = {"default:coal_lump", "default:coal_lump", "default:coal_lump", "default:coal_lump"},
-    })
-end
\ No newline at end of file
+if settings:get_bool("grenades.enable_crafting") == true then
```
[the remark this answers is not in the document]
LoneWolfHT: This will register the grenade recipes by default if it isn't disabled in the settings. Is that ok?

mods/grenades/init.lua
LoneWolfHT: Is setting the yaw needed for a sprite which always faces the player? (Or does it do something else?...)

mods/grenades/init.lua
```diff
@@ -0,0 +1,112 @@
...
+		def.timeout = 5
+	end
+
+	local grenade_entity = {
+		physical = true,
+		timer = 0,
+		visual = "sprite",
+		visual_size = {x = 1, y = 1, z = 1},
+		textures = {def.image},
+		collisionbox = {1, 1, 1, 1, 1, 1},
```
ClobberXD: `collide_with_objects = true` can also be added to allow grenades to not pass through players.

ClobberXD:
The latest commit _only_ fixes the markdown in API.md, and does not add flashbangs to treasure chests and crafting.lua.

Also, re-tested after incorporating the latest changes.
- The "smoke" from smoke grenades aren't messing with my FPS now. :heavy_check_mark:
- Flashbang's overlay works just as before, but using texture modifiers instead. :heavy_check_mark:
- But the issue of not being able to change the trajectory of grenades still persists. :x: 

For the issue of non-varying trajectories, I'll experiment with the code and see what's to be done.

ClobberXD:
A `settingtypes.txt` is only useful for singleplayer at best, and won't be of much use in server-only environments without a GUI. I suggest removing this file, and modifying setting-checks in the code accordingly.

ClobberXD:
Here are my deductions on acceleration and it's components:

- The y-component of acceleration refers to the grenade's acceleration along the vertical (up-down) axis. This is generally set to the value of gravity (~ `-9.81`), which is rounded to `-10` in `shooter`, with the negative signage referring to downwards acceleration. Setting this to `-20` means that the grenade will experience twice as much "gravity" as players do.
- The x-component and z-component together account for drag/air-resistance, and are generally equal unless the code wants to account for phenomenon like wind.

Velocity on the other hand is quite simple. All three components are set to `dir.<axis> * k`, where `k` is a constant. Increasing/Decreasing `k` is all that's required to tune the velocity of the grenade.

I just realized grenades actually do change trajectory depending on how high they're thrown, but the strong "gravity" and low velocity confused me into thinking that wasn't the case.

ClobberXD:
More testing: Thanks to `physical = true`, grenades bounce off vertical surfaces, and even bounces off the ground at times. Instead of checking if node at grenade pos ~= "air", I suggest keeping a tab on the velocity, and triggering the explosion on collision (i.e. when velocity steeply drops due to impact).

I also experimented with velocity and acceleration. The following seems to be one of the most optimal combinations:
```lua
obj:set_velocity({x = dir.x * 40, y = dir.y * 40, z = dir.z * 40})
obj:set_acceleration({x = dir.x * -4, y = -10, z = dir.z * -4})
```

LoneWolfHT:
> More testing: Thanks to `physical = true`, grenades bounce off vertical surfaces, and even bounces off the ground at times. Instead of checking if node at grenade pos ~= "air", I suggest keeping a tab on the velocity, and triggering the explosion on collision (i.e. when velocity steeply drops due to impact).

I think collision is still a little buggy. May not work very well. I'll set it up so it explodes when it hits a wall though

LoneWolfHT:
Ok that's just weird...
image

mods/grenades/init.lua
LoneWolfHT: Ok

mods/grenades/init.lua
LoneWolfHT: Done

LoneWolfHT:
We don't want the grenade's throwing range to be too great or players could 'bomb' the enemy base without risking confrontation

LoneWolfHT:
@ClobberXD I think it's ready for another 'checkup'

mods/grenades/grenades.lua
```diff
@@ -1,188 +1,114 @@
...
-				collisiondetection = true,
-				collision_removal = false,
-				vertical = false,
-				texture = "grenades_smoke.png",
-			})
+grenades.register_grenade("regular", {
+	description = "Regular grenade (Kills anyone near blast)",
+	image = "grenades_regular.png",
+	on_explode = function(pos, name)
+        if not name or not pos then
```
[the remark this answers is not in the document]
LoneWolfHT: huh. Sublime Text switched back to using spaces for indentation

mods/grenades/grenades.lua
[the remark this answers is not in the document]
LoneWolfHT: Removed the settings

ClobberXD:
@LoneWolfHT 

- I feel the grenades are still a tad too slow.
- They are also way too rare in the chests.
- Flashbang's effect wears off a little too quickly.

As of now, I don't see any other issues.

LoneWolfHT:
> @LoneWolfHT
> 
>     * I feel the grenades are still a tad too slow.
> 
>     * They are also way too rare in the chests.
> 
>     * Flashbang's effect wears off a little too quickly.
> 
> 
> As of now, I don't see any other issues.

Ok

Ok. I set them to have the same rarity as the current grenade. These grenades are craftable too. Unlike the current grenade. I will also need some help figuring out what all the arguments in the treasure registration do lol

Ok. I'll increase it slightly

ClobberXD:
Taken from `mods/treasurer/README.md`:

****

==== `treasurer.register_treasure` ====
Registers a new treasure (this means the treasure will be ready to be spawned by treasure spawning mods).

This function does some basic parameter checking to catch the most obvious
mistakes. If invalid parameters have been passed, the input is rejected and
the function returns false. However, it does not cover every possible
mistake, so some invalid treasures may slip through.
		
Rarity does not imply preciousness. A rare treasure may not neccessarily a
very precious one. A treasure chest with scorched stuff inside may be very
rare, but it’s certainly also very unprecious.

===== Parameters =====
* `name`: name of resulting `ItemStack`, e.g. “`mymod:item`”
* `rarity`: rarity of treasure on a scale from 0 to 1 (inclusive). lower = rarer
* `preciousness` : subjective preciousness on a scale from 0 to 10 (inclusive). higher = more precious.
* `count`: optional value which specifies the multiplicity of the item. Default is 1. See `count` syntax help in this file.
* `wear`: optional value which specifies the wear of the item. Default is 0, which disables the wear. See `wear` syntax help in this file.
* `treasurer_groups`: an optional table of group names to assign this treasure to. If omitted, the treasure is added to the default group.

===== Return value =====		
`true` on success, `false` on failure.

=== data formats ===
format of count type:
==== `count` ====
A `count` can be a number or a table 

* `number`: it’s always so many times
* `{min, max}`: it’s pseudorandomly between `min` and `max` times, `math.random` will be used to chose the value
* `{min, max, prob_func}`: it’s between `min` and `max` times, and the value is given by `prob_func` (see below)

==== `wear` ====
Completely analogous to `count`.

LoneWolfHT:
I think the flashbangs could still use a little touching up regarding the blindness. If someone could take a look at that when they test...

ClobberXD:
Sounds would make these grenades more enjoyable to use. Would you be able to add the appropriate sounds for each grenade? `register_grenade` could set a default sound if a grenade isn't registered with a custom one.

LoneWolfHT:
> Sounds would make these grenades more enjoyable to use. Would you be able to add the appropriate sounds for each grenade? `register_grenade` could set a default sound if a grenade isn't registered with a custom one.
> 
> Also, this PR needs to be rebased.

heh. I just finished those. I added them to the non-CTF mod though. So I'll have to copy them over.
While you wait you could use the mod to test out the sounds: https://github.com/LoneWolfHT/grenades

Will do

rubenwardy:
I like this. I would be great to see #356 added to this too, but this can be latter

> When you kill a player with a grenade a sword is shown instead of a grenade in the deathlist HUD

This is no longer hard coded, so it may just work

> Flashbang can be cheated by turning off HUD

Maybe decrease speed too? This would beg for a physics resolver too. I suggest just ignoring this issue for now

ClobberXD:
> This would beg for a physics resolver too.

#289 :)

LoneWolfHT:
Ok. I'll get to work on this again lol
Just realized you can make the smoke and flashbang effects invisible with a texturepack too. May have to apply the slowness for when in smoke too...

LoneWolfHT:
Ok. Using `[noalpha` I fixed the texturepack problem with the flashbang HUD. Not sure how I'll do the smoke...

LoneWolfHT:
heh. I may need to make a new PR. Messed up the rebase...

mods/grenades/grenades.lua
```diff
@@ -0,0 +1,114 @@
...
+	description = "Flashbang grenade (Blinds all who look at blast)",
+	image = "grenades_flashbang.png",
+	on_explode = function(pos, name)
+		for k, v in ipairs(minetest.get_objects_inside_radius(pos, 20)) do
+			if v:is_player() and v:get_hp() > 0 then
+				local playerdir = vector.round(v:get_look_dir())
+				local grenadedir = vector.round(vector.direction(v:get_pos(), pos))
+				local pname = v:get_player_name()
+
+				if vector.equals(playerdir, grenadedir) then
```
DS-Minetest: This is rather inaccurate. You could use instead the angle between the vectors. The lengths being 1 make this even easier:
```lua
-- playerdir and grenadedir are not rounded
if math.arcos(playerdir.x*grenadedir.x + playerdir.y*grenadedir.y + playerdir.z*grenadedir.z) <= math.pi/4 then
```

LoneWolfHT: Yeah. My vector knowledge is limited lol. Thanks

LoneWolfHT:
Moving to #357

PR #296, opened 2018-12-03
Improve vote_kick feature

ClobberXD:
- Moves all vote_kick related code to `vote_kick.lua`.
- Adds chat-command `/unblock` to unblock player before their cooldown time expires.
- Immunizes moderators to being vote-kicked.
- Logs every grant/revoke of `vote_kick` and every instance of a kick-vote

### DIFF
--- a/mods/vote/init.lua
+++ b/mods/vote/init.lua
@@ -1,10 +1,7 @@
 vote = {
 	active = {},
 	queue = {},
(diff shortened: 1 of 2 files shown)


Thomas--S:
How about not only logging the grant/revoke of `vote_kick` and every instance of a kick-vote, but also send it to the moderators like a report?

PR #300, opened 2018-12-06
Add /r in the same message of rankings

MinetestSam:
Adds /r in the random messages

### DIFF
--- a/mods/random_messages/init.lua
+++ b/mods/random_messages/init.lua
@@ -65,7 +65,7 @@ function random_messages.read_messages()
 		"Trapping team mates on purpose is strictly against the rules and you will be kicked immediately.",
 		"Help your team claim victory by storing extra weapons in the team chest, and never taking more than you need.",
 		"Excessive spawn-killing is a direct violation of the rules - appropriate punishments will be given.",
-		"Check your score and your rank in the league tables using /rankings"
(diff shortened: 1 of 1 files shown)


mods/random_messages/init.lua
```diff
@@ -65,7 +65,7 @@ function random_messages.read_messages()
 		"Trapping team mates on purpose is strictly against the rules and you will be kicked immediately.",
 		"Help your team claim victory by storing extra weapons in the team chest, and never taking more than you need.",
 		"Excessive spawn-killing is a direct violation of the rules - appropriate punishments will be given.",
-		"Check your score and your rank in the league tables using /rankings"
+		"Check your score using /r and your rank in the league tables using /rankings"
```
ClobberXD: Needs improvement. Currently it sounds as if the score can only be checked using `/r` and rank in league tables using `/rankings`. It'd be nice if this sounds more unambiguous.

MinetestSam: I will rather add it as a seperate message

ClobberXD: That'd be great. Please make sure you highlight the main advantage of `/r`, which is that it's a much quicker way to check one's score w/ respect to `/rankings`.

PR #240, opened 2018-11-13
Re-add HUD indication of kill score

ClobberXD:
Tested. Will fix merge conflict in a bit.

screenshot_20181113_075001

PR #301, opened 2018-12-11
Add `hud_score` mod API

### LINKED ISSUE #72: Give indication of score gained after a kill

ClobberXD:
Replaces #240, with a generic API usable by mods.

### API

- `hud_score.new`: This is the only one method exposed to mods. This method takes a HUD score element definition as param, see "HUD score element definition" section below.
  - It adds the HUD element def to the player's list of HUD score elements, and calls an internal API function `update(name)` which updates all the HUD score elements of `name`.
  - If the number of HUD score elements exceeds 6, the oldest one will be marked for deletion.
- A globalstep takes care of expired elements by "marking" them for deletion, and calling `update`.

### HUD Score element definition

Here's an example table (taken from README.md):

```lua
{
    name  = "ctf_stats:kill_score",
    color = "0x00FF00",
    value = 50,   
    -- Field `time` is automatically added by `hud_score.new`
    -- to keep track of element expiry
}
```

### Testing code

```lua
minetest.register_chatcommand("test_hud", {
	func = function(name)
		local colors = {
			0x00FF00,
			0x0000FF,
			0xFF0000,
			0xFFFF00
		}

		hud_score.new(name, {
			name = "hud_score:debug_" .. math.random(65535),
			color = colors[math.random(1, 4)],
			value = math.random(50, 500)
		})
	end
})
```

****

Passed tests; everything works as it should. Closes #72

### DIFF
--- a/mods/pvp/hud_score/init.lua
+++ b/mods/pvp/hud_score/init.lua
@@ -1,0 +1,102 @@
+hud_score = {}
+local hud = hudkit()
+
+local players = {}
+local duration = 5
+local max = 6
+
+local function update(name)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
+	end
+
+	-- Handle all elements marked for deletion
+	-- and rebuild table
+	local temp = {}
+	for _, def in ipairs(players[name]) do
+		if def.delete then
+			if hud:exists(player, def.name) then
+				hud:remove(player, def.name)
+			end
+		else
+			table.insert(temp, def)
+		end
+	end
+
+	for i, def in ipairs(temp) do
+		-- If not the top-most element, prefix with "+ "
+		local text = def.value
+		if i > 1 then
+			text = "+ " .. text
+		end
+
+		if hud:exists(player, def.name) then
+			hud:change(player, def.name, "text", text)
+			hud:change(player, def.name, "offset", {x = 0, y = i * 20})
+		else
+			hud:add(player, def.name, {
+				hud_elem_type = "text",
+				alignment = {x = 0, y = 0},
+				position = {x = 0.5, y = 0.6},
+				offset = {x = 0, y = i * 20},
+				number = def.color,
+				text = text
+			})
+		end
+	end
+	players[name] = temp
+end
+
+function hud_score.new(name, def)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
+	end
+
+	-- Verify HUD score element def
+	if not name or not def or type(def) ~= "table" or
+			not def.name or not def.value or not def.color then
+		error("hud_score: Invalid HUD score element definition", 2)
+	end
+
+	-- Store element expiration time in def.time
+	-- and append score element def to players[name]
+	def.time = os.time() + duration
+	table.insert(players[name], def)
+
+	-- If more than `max` active elements, mark oldest element for deletion
+	if #players[name] > max then
+		players[name][1].delete = true
+	end
+
+	update(name)
+end
+
+local modified
+minetest.register_globalstep(function(dtime)
+	-- Loop through HUD score elements of all players
+	-- and remove them if they've expired
+	for name, hudset in pairs(players) do
+		modified = false
+		for i, def in pairs(hudset) do
+			if def.time < os.time() then
+				def.delete = true
+				modified = true
+			end
+		end
+
+		-- If a player's hudset was modified, update player's HUD
+		if modified then
+			update(name)
+		end
+	end
+end)
+
+minetest.register_on_joinplayer(function(player)
+	players[player:get_player_name()] = {}
+end)
+
+minetest.register_on_leaveplayer(function(player)
+	players[player:get_player_name()] = nil
+end)
--- a/mods/ctf/ctf_bounties/depends.txt
+++ b/mods/ctf/ctf_bounties/depends.txt
@@ -1,3 +1,3 @@
 ctf
 ctf_stats
-irc?
+hud_score
--- a/mods/ctf/ctf_bounties/init.lua
+++ b/mods/ctf/ctf_bounties/init.lua
@@ -86,26 +86,27 @@ minetest.register_on_joinplayer(function(player)
 end)
 
 ctf.register_on_killedplayer(function(victim, killer)
-	-- Suicide is not encouraged here at CTF
-	if victim == killer then
+	if victim ~= bountied_player or victim == killer then
 		return
 	end
-	if victim == bountied_player then
-		local main, match = ctf_stats.player(killer)
-		if main and match then
-			main.score  = main.score  + bounty_score
-			match.score = match.score + bounty_score
-			ctf.needs_save = true
-		end
-		bountied_player = nil
 
-		local msg = killer .. " has killed " .. victim .. " and received the prize!"
-		minetest.chat_send_all(msg)
-
-		local pstats, mstats = ctf_stats.player(killer)
-		pstats.bounty_kills = pstats.bounty_kills + 1
-		mstats.bounty_kills = mstats.bounty_kills + 1
+	local main, match = ctf_stats.player(killer)
+	if main and match then
+		main.score  = main.score  + bounty_score
+		match.score = match.score + bounty_score
+		main.bounty_kills = main.bounty_kills + 1
+		match.bounty_kills = match.bounty_kills + 1
+		ctf.needs_save = true
 	end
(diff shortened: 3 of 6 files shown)


ClobberXD:
Crashes when there are more than 6 active elements. WIP.

ClobberXD:
Crash fixed. Tested, works as expected. Could definitely do with some more testing. I've added some testing code to the first post.

Will also let `ctf_bounties` make use of this.

ClobberXD:
Hello world! :3

mods/hud_score/README.md
```diff
@@ -0,0 +1,47 @@
+# `hud_score`
+
+`hud_score` provides an API to display a HUD score elements which can be used to
```
DS-Minetest: "**a** HUD score element**s**" I think, this is wrong.

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+	-- and remove them if they've expired
+	for name, hudset in pairs(players) do
+		modified = false
+		for i, def in pairs(hudset) do
+			if def.time < os.time() then
+				def.delete = true
+				players[name][i] = def
+				if not modified then
+					modified = true
+				end
```
DS-Minetest: Why not simply `modified = true`?

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+local modified
+minetest.register_globalstep(function(dtime)
+	-- Loop through HUD score elements of all players
+	-- and remove them if they've expired
+	for name, hudset in pairs(players) do
+		modified = false
+		for i, def in pairs(hudset) do
+			if def.time < os.time() then
+				def.delete = true
+				players[name][i] = def
```
DS-Minetest: Is this needed? Does `pairs` give you a copy of the table `def`?

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+function hud_score.new(name, def)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
+	end
+
+	-- Verify HUD score element def
+	if not name or not def or not type(def) == "table" or
+			not def.name or not def.value or not def.color then
+		error("hud_score: Invalid HUD score element definition")
```
DS-Minetest: You could set the [error level](http://www.lua.org/manual/5.1/manual.html#pdf-error) to 2.

mods/hud_score/README.md
ClobberXD: oops, good catch. :)

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+local function update(name)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
+	end
+
+	-- Handle all elements marked for deletion
+	-- and rebuild table
+	local temp = {}
+	for _, def in pairs(players[name]) do
```
DS-Minetest: Is there a reason why you use pairs instead of `ipairs`?

mods/hud_score/init.lua
ClobberXD: Oh nice!

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+	-- Verify HUD score element def
+	if not name or not def or not type(def) == "table" or
+			not def.name or not def.value or not def.color then
+		error("hud_score: Invalid HUD score element definition")
+	end
+
+	-- Store element expiration time in def.time
+	-- and append score element def to players[name]
+	def.time = os.time() + duration
+	table.insert(players[name], def)
```
DS-Minetest: `players[name]` is `nil` at first, does `table.insert` make in this case a new table?

mods/hud_score/init.lua
```diff
@@ -0,0 +1,105 @@
...
+local hud = hudkit()
+
+local players = {}
+local duration = 5
+local max = 6
+
+local function update(name)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
```
DS-Minetest:
```suggestion
		players[name] = nil
		return
```

mods/hud_score/init.lua
ClobberXD: I guess that isn't much of an optimization, if at all...

mods/hud_score/init.lua
ClobberXD: The `players` table and its sub-tables all contain `nil`s in the middle of the table. AFAIK, `ipairs` works only with contiguous indices, right?

mods/hud_score/init.lua
ClobberXD: The `on_joinplayer` callback at L96 initializes `players[name]` when a player joins the game.

mods/hud_score/init.lua
ClobberXD: The `on_leaveplayer` callback at L100 sets `players[name]` to `nil`

mods/hud_score/init.lua
DS-Minetest: The sub-tables don't contain `nil`s in the middle. You are filling `temp` without gaps and set `players[name] = temp`.

mods/hud_score/init.lua
DS-Minetest: Oh I haven't seen it, sorry.

mods/hud_score/init.lua
DS-Minetest: Oh I haven't seen it, sorry. This can be marked as resolved.

mods/hud_score/init.lua
ClobberXD: That's what I thought. Are there references in Lua?

mods/hud_score/init.lua
ClobberXD: Good point.

mods/hud_score/init.lua
ClobberXD: https://stackoverflow.com/a/6078882

DS-Minetest: Yes, there are references in lua. I think, only strings, numbers and bools are "copied when they are function parameters".
In the linked example, the top `editColor` function modifies `col`, the bottom function assigns a new color to `col` so `col` does no longer point to the same as `color`.

mods/pvp/hud_score/init.lua
```diff
@@ -0,0 +1,103 @@
...
+end
+
+function hud_score.new(name, def)
+	local player = minetest.get_player_by_name(name)
+	if not player then
+		return
+	end
+
+	-- Verify HUD score element def
+	if not name or not def or not type(def) == "table" or
```
DS-Minetest: http://www.lua.org/manual/5.1/manual.html#2.5.6
`not type(def)` will be `false` and `false == "table"` is `false`. So `type(def) == "table"` is always `false` because `type` gives back strings and `not` has a higher precedence than `==`.
Use `~=`!
```suggestion
	if not name or not def or type(def) ~= "table" or
```

mods/hud_score/init.lua
ClobberXD: Tried it out myself - it doesn't seem to be the case. Modifying the _copied_ table using `for` and `pairs` doesn't reflect in the original table:

```lua
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> t = {1, 2, 3, 4, 5, 6, 7, 8}
> do
>> for i, n in pairs(t) do
>> if i == 5 then
>> n = 9
>> end
>> end
>> end
> print(t[5])
5
> 
```

DS-Minetest: This is because you used numbers.
At first `n` is assigned to `5` (which is in `t[5]`). `9` is a constant. You assign `n` to a new value, you do not modify `t`.
Try something like:
```lua
local t = {{1}, {2}, {3}, {4}, {5}, {6}, {7}}
for i, n in pairs(t) do
	if i == 5 then
		n[1] = 9
	end
end
print(t[5][1])
```

ClobberXD: That does work...

Will perform more tests, and push.

ClobberXD: Works perfectly, thanks for the tip.

ClobberXD:
Thanks for your review, @DS-Minetest.

ClobberXD:
Deployed this PR on my public testing server, and had no issues, warnings, or errors whatsoever.

PR #302, opened 2018-12-12
Traps

MinetestSam:
Adds various traps for CTF. Namely,

- Ghost stone

- Ghost cobblestone

- Ghost dirt

- Spikes

- Fake/Damage Cobblestone

Related to #35


mods/traps/init.lua
```diff
@@ -0,0 +1,135 @@
+minetest.register_node("traps:dirt", {
+    description = "Fake Dirt",
```
ClobberXD: Elaborate "Fake". Maybe `"Unwalkable Dirt"`?

mods/traps/init.lua
```diff
@@ -0,0 +1,135 @@
+minetest.register_node("traps:dirt", {
+    description = "Fake Dirt",
+    tiles = {"traps_dirt.png"},
```
ClobberXD: The textures which are just lighter/darker overall, can use texture modifiers on the default textures instead of having to use a separate one.

mods/traps/init.lua
```diff
@@ -0,0 +1,135 @@
...
+	type   = "inv",
+	output = "traps:damage_cobble",
+	items  = { "default:cobblestone", "default:coal 4", "default:steel_ingot 4" },
+	always_known = true,
+    level  = 1,
+})
+]]--
+
+minetest.register_node("traps:damage_cobble", {
+    description = "Cobblestone that does damage",
```
ClobberXD: Ambiguous description. Please replace with `"Cobblestone that damages digger"` or something similar instead.

mods/traps/init.lua
```diff
@@ -0,0 +1,95 @@
...
+		local pos
+		local node = minetest.get_node(pointed_thing.under)
+		local pdef = minetest.registered_nodes[node.name]
+		if pdef and pdef.on_rightclick and
+			not placer:get_player_control().sneak then
+			return pdef.on_rightclick(pointed_thing.under,
+			node, placer, itemstack, pointed_thing)
+		end
+
+		if pdef and pdef.buildable_to then
```
ClobberXD: Empty `if` block

mods/traps/init.lua
```diff
@@ -0,0 +1,92 @@
...
+	walkable = true,
+	groups = {cracky=3, stone=2},
+	on_place = function(itemstack, placer, pointed_thing)
+		local pos
+		local node = minetest.get_node(pointed_thing.under)
+		local pdef = minetest.registered_nodes[node.name]
+		if pdef and pdef.on_rightclick and
+			not placer:get_player_control().sneak then
+			return pdef.on_rightclick(pointed_thing.under,
+			node, placer, itemstack, pointed_thing)
```
ClobberXD: Needs a 2-tab indentation

Thomas--S:
:-1: except for spikes because all other traps will probably be mainly used by griefers.

:+1: for spikes.

MinetestSam:
The damaging cobble cannot be used by griefers as it will only damage the opponent team

mods/traps/depends.txt
```diff
@@ -1,3 +1,2 @@
 default
-ctf_crafting?
-crafting?
\ No newline at end of file
+crafting?
```
ClobberXD: No need for this either, as this is only relevant to `ctf_crafting`

MinetestSam:
I don't think I should add these traps to ctf_treasure as it will be much more easier for the griefers ( generally noobs) to get it.

ClobberXD:
This mod depends on `ctf` mod for retrieving the damaging-cobble's placer's team.

Please create a `depends.txt` for `traps`, and add `ctf` in it. Since the mod depends on `ctf`, please rename it to `ctf_traps`.

mods/traps/init.lua
```diff
@@ -0,0 +1,92 @@
...
+
+minetest.register_node("traps:stone", {
+	description = "Unwalkable Stone",
+	tiles = {"traps_stone.png"},
+	is_ground_content = false,
+	walkable = false,
+	groups = {cracky=3, stone=1}
+})
+
+minetest.register_node("traps:cobblestone", {
```
ClobberXD: `"traps:cobble"` to be consistent with `default`

ClobberXD:
I performed my first round of testing, and here are my thoughts:

- The fall-through nodes are very noticeable due to their yellow-green border. They could be subtler.
- The spikes could be made a bit taller, to be more noticeable.

mods/traps/init.lua
```diff
@@ -0,0 +1,92 @@
...
+		end
+
+		local digger_team = ctf.player(name).team
+		local meta = minetest.get_meta(pos)
+		local placer_team = meta:get_string("placer") or "missing"
+		if digger_team ~= placer_team then
+			local hp = digger:get_hp()
+			digger:set_hp(hp - 7)
+			minetest.remove_node(pos)
+			return
```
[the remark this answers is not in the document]
MinetestSam: It needs to remove the node when the damage is done. If minetest.remove_node is remove then it will not remove the node and the member of the opposite team could keep on collecting these damage inflicting cobblestones

mods/ctf_traps/init.lua
```diff
@@ -38,9 +38,9 @@ minetest.register_node("traps:spike", {
 	},
 })
 
-minetest.register_node("traps:damage_cobble", {
+minetest.register_node("ctf_traps:damage_cobble", {
 	description = "Cobblestone that damages digger",
-	tiles = {"traps_dmg_cobble.png"},
+	tiles = {"ctf_traps_dmg_cobble.png"},
```
ClobberXD: Texture-names should follow same convention as node-name:
`<mod_name>:<node_name>` --> `<mod_name>_<node_name>`
So this texture should be renamed to `ctf_traps_damage_cobble`.

ClobberXD:
Will test again soon.

ClobberXD:
Tested once again:

- No change to damaging cobble.
- Spikes look better. :heavy_check_mark:
- Textures of fake nodes look too subtle now. :x: 
  - I suggest overlaying the original textures with a mild tint. The textures will be subtle, and yet noticeable if the player is careful enough.

mods/ctf_traps/init.lua
```diff
@@ -0,0 +1,71 @@
...
+	damage_per_second = 5,
+	groups = {cracky=1, level=2},
+	selection_box = {
+		type = "fixed",
+		fixed = {-0.5, -0.5, -0.5, 0.5, 0, 0.5},
+	},
+})
+
+minetest.register_node("ctf_traps:damage_cobble", {
+	description = "Cobblestone that damages digger",
```
ClobberXD: Some sort of a clarification would be appreciated here, as this node only damages the digger if they're on the enemy team.

ClobberXD:
I still think a slight tint would be better than just making them darker, as they still are harder to identify in the dark.

screenshot_20181220_151216
screenshot_20181220_151205

Nodes in screenshot from left-to-right:
- `ctf_traps:dirt`
- `default:dirt`
- `ctf_traps:stone`
- `default:stone`
- `ctf_traps:cobble`
- `ctf_traps:damage_cobble`
- `default:cobble`

ClobberXD:
screenshot_20181221_202955
screenshot_20181221_203002

Textures are well-balanced; subtle yet noticeable. I notice some sort of a pattern on the textures of all three fake nodes, intentional?

MinetestSam:
Yup

ClobberXD:
Textures look good to me. :+1:

I suspect the damage cobble might be abused by tunnellers. On the other hand, that's alright, because the defending team should've protected their flag against tunnellers anyway. I feel this needs to be given more thought.

MinetestSam:
Well same can be said for the defending team because you can make a hole below and cover it all with dmg cobble

ClobberXD:
Alright, makes sense (sorta). :+1: from me.

ClobberXD:
However it's worth gathering more opinions about the textures of the fake nodes before merging. I'm finding them well-balanced with respect to subtlety and ability to differentiate them in the dark as well (using default `display_gamma`)

bewlay:
I can only make Spikes, the ghost and damage blocks seem to have misspelled ingredients.

MinetestSam:
@bewlay Thanks for that, did some silly mistakes and it should be fixed now

bewlay:
The fact that damage cobble looks always the same (regardless which team placed it) could be a problem. In some ways, that would be similar as if all blocks damaged you, as you don't want to dig a damage cobble block of which you don't know the origin (except maybe directly at the base, where you have time to heal anyway).

Good work otherwise, btw.

ClobberXD:
> The fact that damage cobble looks always the same (regardless which team placed it) could be a problem.

I'd consider that a part of the challenge. Traps can be compared to a double-edged sword, which can damage the wielder as well.

MinetestSam:
> The fact that damage cobble looks always the same (regardless which team placed it) could be a problem. In some ways, that would be similar as if all blocks damaged you, as you don't want to dig a damage cobble block of which you don't know the origin (except maybe directly at the base, where you have time to heal anyway).
 
If players would be able to differentiate then they would just mine around that cobble and there would be no point in having it.

MinetestSam:
I think the mod is complete but if anyone wants to test it and send any feedbacks then that would be helpful

ClobberXD:
Needs to be rebased to fix merge conflicts.

```sh
git checkout traps
git pull upstream master --rebase
# where upstream is the MT-CTF/capturetheflag base repo
```

MinetestSam:
Doesn't work :| . It can't find the 'MT-CTF/capturetheflag' git repository

ClobberXD:
Ok... can you paste the output of `git remote -v`?

MinetestSam:
Here
> origin  https://github.com/MinetestSam/capturetheflag (fetch)
origin  https://github.com/MinetestSam/capturetheflag (push)

rubenwardy:
I can rebase this before merging

MinetestSam:
Thanks

rubenwardy:
20daeebead5bc13a96a2dd323b5b66775d97920b

PR #311, opened 2018-12-21
Add ctf_anticombatlog

ClobberXD:
In the wake of the recent episodes of chaos caused by players combat-logging, I've worked on adding a heavily modified and optimized version of Elkien3's [anticombatlog](https://github.com/Elkien3/anticombatlog). This mod depends on `ctf` in order to engage the anti-combatlogging mechanism only when player engages in PvP with the enemy team, rather than engaging _every_ time a player is punched, even by team-mates.

As this mod uses a custom version of `dropondie` there exists a lot of duplicated code. To mend this, our existing `dropondie` mod would extend its `drop_all(player)` function into `dropondie.drop_all(inv, pos)`. Notice that the newer version of `drop_all` takes different arguments. This is to allow detached inventories to make use of the same code.


AKryukov92:
What combat logging is?

ClobberXD:
Sorry for not explaining that. Combat-logging is the act of intentionally quitting the game when another player starts attacking; in this case, to preserve their K/D or to not award the attacker with their score.

AKryukov92:
I think I have seen such thing in World of Warcraft.

When blue player attacks red player and red quits, then red player replaced with some "ghost" entity.
If blue player kills ghost, then you update K/D and score like blue player kills real red player.
Ghost disappears after `timeout` seconds.
When ghost dies or disappears, then all inventory of red player get dropped on the ground.

Why do you need to track combat state of players?
You can just create ghosts for every left player. It doesn't happen so often to cause any performance problems. This way you will be able to simplify your code a bit.

Will you display combat state of player in some kind of HUD element?
(This question actually should be asked in issue, but there is no accosiated issue)
I think that players should know when they can quit without penalty.

I don't see where you update K/D of players on interaction with ghosts. Please point me where is it happens.

ClobberXD:
> When blue player attacks red player and red quits, then red player replaced with some "ghost" entity.
If blue player kills ghost, then you update K/D and score like blue player kills real red player.

Yes, that exactly how this is supposed to work.

> Why do you need to track combat state of players?
You can just create ghosts for every left player. It doesn't happen so often to cause any performance problems. This way you will be able to simplify your code a bit.

Why should this apply to all players? The whole point of this mod is to penalize players only if they leave in the middle of combat.

> Will you display combat state of player in some kind of HUD element?
I think that players should know when they can quit without penalty.

I initially did not plan on doing that, but now that you provide a very valid reason, I'll re-add it (a HUD element is shown in the original mod). Thanks for pointing that out.

> I don't see where you update K/D of players on interaction with ghosts. Please point me where is it happens.

I haven't added that part yet. As you can see, this PR is still WIP, and will be finished in a day or two.

mods/ctf/ctf_anticombatlog/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = ctf_anticombatlog
+depends = ctf,dropondie,default
```
Panquesito7: Better understanding.
```suggestion
depends = ctf, dropondie, default
```

mods/pvp/dropondie/init.lua
```diff
@@ -12,17 +14,18 @@ local function drop(pos, itemstack)
 	end
 
 	local obj = minetest.add_item(pos, it)
-
 	if obj then
-		obj:setvelocity({x=math.random(-1,1), y=5, z=math.random(-1,1)})
+		obj:setvelocity({
```
Panquesito7: Please use `set_velocity`.

mods/ctf/ctf_anticombatlog/mod.conf
ClobberXD: I don't think spaces are allowed here.

Panquesito7: They're allowed, and it works.

ClobberXD: I prefer to not add spaces, sorry.

mods/pvp/dropondie/init.lua
ClobberXD: This was already fixed in a recent PR. I just had to rebase this PR.

PR #321, opened 2019-01-07
Add alternate furnace recipe

MinetestSam:
Trivial PR
Adds alternate furnace recipe that requires desert cobblestone so as to fit with the maps that contain desert cobblestone such as The Bridge and Desert Spikes or those that may contain desert cobblestone in the future


ClobberXD:
Order of craft registration matters when using the `crafting` mod. It'd be nice if both the furnace recipes are next to each other.

Issue #313, opened 2018-12-23
ctf_stats requires a major refactor/cleanup and a proper API


ClobberXD:
There are at least 4 PRs that modify various parts of `ctf_stats`, and even if we don't take them into consideration, both `gui.lua` and `init.lua` are getting very cluttered. `ctf_stats` needs some substantial amount of refactoring and cleanup.

The mod should also provide a proper API, which can be used by mods that adds new stats like bounty kills and kill assists. e.g. to add a new stat `"kill_assists"`, the new code should just have to

- Register the new stat. e.g. `ctf_stats.register_stat("kill_assists", "Kill assists")`
- Modify the new stat whenever required like the following example:
```lua
    local assists = ctf_stats.get_stat(name, "kill_assists")
    ctf_stats.set_stat(name, "kill_assists", assists + 1)
```

Please share your thoughts on this. I'll start working on this once those 4 PRs are merged.

ClobberXD:
I'd say the very first step is to split the two existing files into 4 files as given below:

- chatcommands.lua: Exclusively for chat-commands.
- init.lua: All the callbacks (`on_dieplayer`, `on_killedplayer`, `on_capture`, `on_winner`, etc.)
- api.lua: Existing API, `ctf_stats.players`, etc.
- gui.lua: End-of-match, league tables formspecs

AKryukov92:
I think that new API worth implementation for things that are supposed to be added in dozens with some common logic. Something like blocks, tools and entites.
How many similar stat of player can you imagine? I don't think that their quantity will be high.
Thats why I'm not sure that API for action "add new stat" is something that worth doing.
You will get additional abstraction layer and increase complexity of code.

You showed code to add new stat in case of making API.
Can you write similar snippet to illustrate amount of code, required to do it now?
Maybe it is trivial enough to keep things as they are.

I think that any refactoring should be done with [Clean Architecture](https://rubenwardy.com/minetest_modding_book/en/quality/clean_arch.html) in mind.
In addition, I think that code should be covered with [unit-test](https://rubenwardy.com/minetest_modding_book/en/quality/unit_testing.html)s before you do any major transformations.
It will help you to find bugs on development stage. Which will be much faster than usual.

AKryukov92:
>  I'll start working on this once those 4 PRs are merged.

Have you considered option to make 1 PR instead of 4 and do refactoring there?

ClobberXD:
[...]

PR #327, opened 2019-01-10
Extend /summary to show scores of current match as well

### LINKED ISSUE #310: Add support for displaying scores of the current match
Preferably using the same end-of-match formspec; or as I'd now like to call it, the match scores formspec. This probably isn't as simple as the "previous match summary" feature, as the data has to be calculated every time the formspec is requested. I also suggest combining both these features into the same command, and incorporating `<` and `>` buttons to switch between scores of the current match and the previous match.

ClobberXD:
screenshot_20190109_215649
screenshot_20190109_215652

- `/summary` can now display the scores of the current match in addition to the previous match summary.
  - The button to the left of `Close` can be used to switch between the two pages.
- `prev_match_summary` can be retrieved using the `ctf_stats.get_prev_match_summary` (to allow other files to access this var. See https://github.com/MT-CTF/capturetheflag/issues/313#issuecomment-452619913)

Closes #310. Tested.

### DIFF
--- a/mods/ctf/ctf_stats/init.lua
+++ b/mods/ctf/ctf_stats/init.lua
@@ -1,7 +1,12 @@
 ctf_stats = {}
 
 local storage = minetest.get_mod_storage()
+local prev_match_summary = storage:get_string("prev_match_summary")
 local data_to_persist = { "matches", "players" }
+
+function ctf_stats.get_prev_match_summary()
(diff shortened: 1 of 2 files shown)


ClobberXD:
Rebased to fix merge conflict.

ClobberXD:
Ran without any problems on my test server.

PR #335, opened 2019-01-16
Add killnode

archfan7411:
This is my first PR to a major repository, and thus while it is trivial I may have messed something up. Please bear with me.

This PR adds a node, `ctf_map:killnode`, which kills players that come within 1 node of it. It is intended for use in maps where lava would otherwise be used to force players to respawn.

While I have done some testing myself, more testing of this feature is needed.


mods/ctf/ctf_map/nodes.lua
```diff
@@ -735,6 +735,30 @@ minetest.register_node("ctf_map:meselamp", {
...
+	sunlight_propogates = true,
+	is_ground_content = false,
+	groups = {immortal = 1},
+	sounds = default.node_sound_stone_defaults(),
+	after_place_node = function(pos, placer, itemstack, pointed_thing)
+		 local timer = minetest.get_node_timer(pos)
+		 timer:start(0.5)
+	end,
+	on_timer = function(pos, elapsed)
+		local objects = minetest.get_objects_inside_radius(pos, 1)
```
ClobberXD: IIRC, a radius of 1 covers only the node itself. Is this what you intended to do?

mods/ctf/ctf_map/nodes.lua
```diff
@@ -735,6 +735,30 @@ minetest.register_node("ctf_map:meselamp", {
...
+		for i, object in pairs(objects) do
+		    if minetest.is_player(object) then
+		       object:set_hp(0)
+		    end
```
ClobberXD: L754-L756 use a mix of tab-spaces and white-spaces for indents, please convert them to pure tab-spaces.

mods/ctf/ctf_map/nodes.lua
```diff
@@ -735,6 +735,30 @@ minetest.register_node("ctf_map:meselamp", {
...
 
+minetest.register_node("ctf_map:killnode", {
+	description = "Kill Node",
+	drawtype = "glasslike",
+	tiles = {"ctf_map_killnode.png"},
+	paramtype = "light",
+	sunlight_propogates = true,
+	is_ground_content = false,
+	groups = {immortal = 1},
+	sounds = default.node_sound_stone_defaults(),
```
ClobberXD: `glasslike` drawtype with stone sounds?

mods/ctf/ctf_map/nodes.lua
```diff
@@ -735,6 +735,30 @@ minetest.register_node("ctf_map:meselamp", {
 	light_source = default.LIGHT_MAX,
 })
 
+minetest.register_node("ctf_map:killnode", {
```
ClobberXD: Is it a good idea to add `pointable = false`?

Thomas--S:
I think using a `walkable = false` node with `damage_per_second` would introduce less lag.

MinetestSam:
Hm introducing a node which does 20 damage per second should do the job

```
 minetest.register_node("xyz:x",
 {
    description = "X",
    tiles = {"x.png"},
    damage_per_second = 20,
    walkable = false,
    groups = {immortal =1,},
    pointable = false
})
```

archfan7411:
The PR is updated to use `walkable = false` and `damage_per_second = 20`. To be honest, I had no idea the latter even existed - thanks!

mods/ctf/ctf_map/nodes.lua
archfan7411: Then map creators would have to use WorldEdit to remove misplaced nodes.

MinetestSam:
`damage_per_second` is only used when the node is unwalkable the player won't be killed if they can walk on it

archfan7411:
Yep, my latest commit also made the node unwalkable.

PR #338, opened 2019-01-17
Track furnace locations and remove all placed furnaces on new match

ClobberXD:
This PR fixes the bug of normal stones having the meta of a furnace, as schem placement doesn't touch node metas.

Note: Instead of `table.insert(table, pos)`, I'm using the furnace pos itself as index, to avoid having to loop through the whole table when a furnace is manually dug.

Tested.


ClobberXD:
I'm also updating maps sub-module in this PR.

PR #343, opened 2019-01-19
Prevent arrows from colliding with players

ClobberXD:
Also fixes usage of deprecated API methods in `grenade.lua`.

Tested, works.

### DIFF
--- a/mods/pvp/shooter/crossbow.lua
+++ b/mods/pvp/shooter/crossbow.lua
@@ -32,6 +32,7 @@ end
 
 minetest.register_entity("shooter:arrow_entity", {
 	physical = false,
+	collide_with_objects = false,
 	visual = "mesh",
 	mesh = "shooter_arrow.b3d",
 	visual_size = {x=1, y=1},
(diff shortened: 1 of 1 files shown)


AKryukov92:
Why?
With this commit, arrows will pierce all players in their way and deal damage to each?

AKryukov92:
Have you considered other ways of solving this problem?
What about:
1. Drop arrows as item when they hit the ground or player.
2. Remove arrow when they hit the ground or player.

Is there a simple callback on collision between objects with `physical = true` and players?
Is there a simple callback on collision between objects with `collide_with_objects = true` and players?

ClobberXD:
> Have you considered other ways of solving this problem?
> What about:
> 
> 1. Drop arrows as item when they hit the ground or player.
> 2. Remove arrow when they hit the ground or player.

Both 1 and 2 are viable, but aren't as realistic.

> Is there a simple callback on collision between objects with `physical = true` and players?
> Is there a simple callback on collision between objects with `collide_with_objects = true` and players?

Neither callbacks exist, even in 5.0.0-dev, AFAIK.

PR #344, opened 2019-01-19
Reduce width of name column in scores formspec

ClobberXD:
The current width results in the last 2-3 columns being clipped out on low DPI displays. This PR reduces the width of the name column by 4 slots, from 20 to 16.

Tested. The largest possible name with 19 full-width chars, is displayed fully, even on a fairly low DPI display.

screenshot_20190119_133144

### DIFF
--- a/mods/ctf/ctf_stats/gui.lua
+++ b/mods/ctf/ctf_stats/gui.lua
@@ -2,7 +2,7 @@
 local tablecolumns = {
 	"tablecolumns[color;",
 	"text;",
-	"text,width=20;",
+	"text,width=16;",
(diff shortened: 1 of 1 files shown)


bewlay:
Works for `/rankings` and end-of-game summary, but not for `/summary`.

Thanks for making this, as I'm one of the people who aren't currently able to see the scores in the rankings!

ClobberXD:
> but not for `/summary`

That shouldn't be possible. All scores-related formspec is handled by the same function. Could you retest, if possible?

bewlay:
You are right, I must have confused something, sorry. Everything works for me.

PR #345, opened 2019-01-19
Add bandages

MinetestSam:
Bandages can be used to heal teammates when necessary. This should promote more of sharing and team work which is the ultimate goal of CTF


ClobberXD:
Requires a rebase on top of latest master.

mods/ctf_bandages/init.lua
```diff
@@ -0,0 +1,31 @@
...
+				return
+			end
+
+			if ctf.player(object:get_player_name()).team == team then
+				local hp = object:get_hp()
+				if hp > 0 and hp <= 15 then
+					object:set_hp(hp + math.random(3,4))
+					itemstack:take_item()
+				else
+					minetest.chat_send_player(object:get_player_name(), "Your HP is above 15!")
```
ClobberXD: I don't see a need for this, why would the player to be healed receive this message? The "can't heal" message only concerns the player who uses bandages, and not the player whom it's used on.

mods/ctf_bandages/init.lua
```diff
@@ -0,0 +1,31 @@
+--Inspired from Andrey's bandages mod
+
+minetest.register_craftitem("ctf_bandages:bandage", {
+	description = "Bandage, heals teammates for 15HP",
```
AKryukov92: These bandages doesn't heal `for15 hp`. According to line 19, they heal only for 3-4.

AKryukov92:
Bandages should also give an ability to send message "I want to heal you" to players, who are further than bandage distance. In the same time, patient should not get spammed by this message.
In CTF when you help flag carrier to get out of base, you don't really have time to chat.

I didn't see any cooldown in these bandages. Does it allow to heal with speed of clicking?

MinetestSam:
I think there are many things that can be done with this PR. I hurried up this due to some project works but I will work on it more. For now I am closing it.

ClobberXD:
Feel free to keep this open if you want. I could just label this PR as WIP until you say you're done with it.

PR #346, opened 2019-01-20
Use mod.conf and modpack.conf; remove .txt counterparts

ClobberXD:
Haven't updated the individual mods in `mtg` modpack, as we'll probably switch to the 5.0.0-dev version of those mods anyway.

"Forward-ports" other commits from master as well.


ClobberXD:
Argh, based the branch on master. \*sigh\*

PR #349, opened 2019-01-25
Improve grenades

ClobberXD:
- Reduce collision-checking interval. Fixes grenades occasionally passing right through nodes without exploding.
- Increase initial velocity of grenade entity. Players can enjoy improved accuracy and larger range due to the increased initial velocity.
- Make grenade blast damage scaling more realistic. See table below for comparison.
- ~~Allow stacking upto 10 grenades~~

#### Blast damage scaling comparison

Current formula: `fleshy * 0.5 ^ dist * 3`
Proposed formula: `fleshy * 0.5 / dist * 3`

Player's distance from grenade in `m` | Damage (current formula) in `HP` | Damage (proposed formula) in `HP`
--- | --- | ---
0.1 | 69.97 | 375
0.5 | 53.03 | 75
1 | 37.5 | 37.5
2 | 18.75 | 18.75
3 | 9.375 | 12.5
4 | 4.69 | 9.375
5 | 2.34 | 7.5

Notice how the damage is extremely low and unrealistic as the distance from the grenade increases as per the current formula. The new formula makes damage a little more realistic and consistent, even at higher distances.

### DIFF
--- a/mods/pvp/shooter/grenade.lua
+++ b/mods/pvp/shooter/grenade.lua
@@ -21,7 +21,7 @@ minetest.register_entity("shooter:grenade_entity", {
 	end,
 	on_step = function(self, dtime)
 		self.timer = self.timer + dtime
-		if self.timer > 0.2 then
+		if self.timer > 0.1 then
 			local pos = self.object:getpos()
 			local above = {x=pos.x, y=pos.y + 1, z=pos.z}
 			if minetest.get_node(pos).name ~= "air" then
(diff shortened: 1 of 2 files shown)


AKryukov92:
I don't think that stacking grenades is good idea.
Stacking will allow throwing them one by another in quick sequence. Assuming their damage, it is overpowered.
Then they stay in inventory as separate entities, then they clutter hotbar which is good for their damage and player have to switch from one to another by wheel or numbers.

archfan7411:
I share the same concerns about stacking grenades, but it's unlikely for a team to have more than 3 grenades every match so it probably isn't a pressing issue. However, I am more concerned about the increased spawnkilling possibility, due to the the huge increase in damage for distances under a node. Is spawnkilling already taken into account?

ClobberXD:
Stacking could lead to some serious issues, I agree. This could be partly fixed by not allowing the player to throw grenades within an interval of 1s or so. Note that grenades can be thrown in quick succession, even if they aren't stackable.

@MinetestSam We still have to put up with the existing crappy grenades until 279 is merged. We might as well make them a little more reliable.

ClobberXD:
I'll remove the stacking part.

rubenwardy:
The old formula made it work as inverse square law, which was good. Also, do grenades raycast? Too large a radius and you risk walls being in the way

ClobberXD:
The old formula is weak, but more importantly inconsistent - the damage reduction with increasing distance is too extreme, and you can survive even if you're just a couple of nodes away from the blast, which is very unrealistic.

ClobberXD:
It might become too late to merge this, if delayed any more. This is just a temporary fix to `shooter`'s grenade until it's replaced by #279.

rubenwardy:
Still not happy with this. The long tail of your function is way too long.

damage

I suggest using `fleshy * (0.707106 ^ dist) * 3` instead. This is equivalent to `fleshy * 0.5 ^ (dist * 0.5) * 3`

PR #351, opened 2019-01-31
Allow maps to optionally specify custom treasures

ClobberXD:
See https://github.com/MT-CTF/capturetheflag/issues/243#issuecomment-457525901

****

- Allow maps to optionally define custom treasures, by adding a `treasures` field to the conf file, which takes a serialized string in the following format:

```conf
treasures = item1,rarity,preciousness,count;item2,rarity,preciousness,count;item3,rarity,preciousness,{min_count,max_count}
# count can either be a number or a table containing a min and max value
```

- If `treasures` isn't defined, the default treasures from `ctf_treasures` are registered.
- ctf_map's README has been updated with info relating to map meta and all the available fields.

Tested; works exactly as it should.


ClobberXD:
To test:

- Add the following to the .conf file of any map:
```
treasures = default:pick_steel,0.5,5,1,10;shooter:shotgun,0.04,2,1;shooter:grenade,0.08,2,1
```
- Startup CTF.
- `/set_next <map_name>`, to select the next map. e.g. `/set_next cavern` will select `03_cavern`.
- `/ctf_next`, to immediately skip to the next match.

This will register only steel pick, shotgun, and grenade as the map's treasures.

bewlay:
Tested with altered parameters, seems to work.

ClobberXD:
Fixed merge conflict.

PR #357, opened 2019-03-05
Add grenades mod

LoneWolfHT:
I broke #279 when rebasing it so here's the updated PR

This PR only adds smoke/regular grenades. It'll be a bit of a hassle to do a flashbang grenade. So I decided to leave that for later

**Todo:**
Find bugs pls


ClobberXD:
I'm okay with not adding flashbang - it's not necessary. Will test soon. Is there anything I should test more carefully?

LoneWolfHT:
> I'm okay with not adding flashbang - it's not necessary. Will test soon. Is there anything I should look out for? Anything I should test more carefully?

Mainly physics. I'm not sure what looks fine and what doesn't. Or how far it should be possible to throw them

ClobberXD:
- Sound files don't seem to follow MT's nomenclature for media files. Please prefix them with the mod name.
- Could `grenades_basic` be merged into `grenades`? While API separation is good for standalone mods, that isn't really required when bundling it into a game, and it just adds to clutter.
  - `grenades_basic/init.lua` could be added as `grenades/grenades.lua`, and could be `dofile`-d at the end of `grenades/init.lua`.
  - Media from `grenades_basic` could be moved to their respective sub-dirs in `grenades`, and renamed accordingly.

ClobberXD:
- Grenades fly a little too fast to be realistic. Maybe reduce their initial velocity?
- Their cooking time (`clock` in the API) is too less. 5 seconds would be more realistic, IMO.
- The duration of the smoke grenade is too less. An ideal duration would be between 30-60 secs.

Thomas--S:
> Could you also remove grenades from shooter in this PR?

Didn't we want to convert shooter into a submodule?

I suggest to simply unregister shooter's grenade in this mod here.

ClobberXD:
> Didn't we want to convert shooter into a submodule?

That doesn't seem to be happening anytime soon, unfortunately. Until that PR is merged, we'd end up with two grenades.

But on second thought, this PR removes grenade's treasure registration, which means that players won't be able to obtain it anyway. That's fine too.

Thomas--S:
> I suggest to simply unregister shooter's grenade in this mod here.

I did actually mean to completely remove the shooter mod grenade entirely by using `minetest.unregister_item`.

LoneWolfHT:
>     * Grenades fly a little too fast to be realistic. Maybe reduce their initial velocity?

I was basing it off of the speed a rock flies when I throw it

>     * Their cooking time (`clock` acc. to the API) is too less. 5 seconds would be more realistic, IMO.

I thought 3 seconds was what it was for real life. But upon googling I got this:
```
Approximately 75,000,000 grenades were manufactured during World War I, used in the war and remaining in use through to the Second World War. At first, the grenade was fitted with a seven-second fuze, but during combat in the Battle of France in 1940, this delay proved too long, giving defenders time to escape the explosion or to throw the grenade back, so the delay was reduced to four seconds. 
```
Changed to 4 seconds for the same reason lol

>     * The duration of the smoke grenade is too less. An ideal duration would be between 30-60 secs.

Changed to 45

ClobberXD:
Awesome, will test again soon :tm:.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,133 @@
+grenades = {}
+
+local function throw_grenade(name, player)
+	local dir = player:get_look_dir()
+	local pos = player:get_pos()
+	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.6, z = pos.z + dir.z}, name)
+	local self = obj:get_luaentity()
+
+	obj:set_velocity({x = dir.x * 40, y = dir.y * 30, z = dir.z * 40})
+	obj:set_acceleration({x = 0, y = -30, z = 0})
```
ClobberXD: `y = -30`

I was wondering why the grenades fell so quickly. This should only be 10 or 9.8, lol. :)

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,133 @@
+grenades = {}
+
+local function throw_grenade(name, player)
+	local dir = player:get_look_dir()
+	local pos = player:get_pos()
+	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.6, z = pos.z + dir.z}, name)
+	local self = obj:get_luaentity()
+
+	obj:set_velocity({x = dir.x * 40, y = dir.y * 30, z = dir.z * 40})
```
ClobberXD: All components of the `dir` vector should be multiplied by the same constant. Only then will the entity point along the thrower's look dir (initially, at least). I suggest creating the multiplier as a local var. which will make modifying it much easier and cleaner. e.g.

```lua
local mult = 40
obj:set_velocity({ x = dir.x * mult, y = dir.y * mult, z = dir.z * mult })
```

mods/pvp/grenades/init.lua
LoneWolfHT: I wasn't able to get 9.8 to work the way I wanted it to. In the latest commit I've made them fall 2/3 as fast. which seems ok to me. If I decrease it too much it starts to act weird

mods/pvp/grenades/init.lua
LoneWolfHT: How's the change?

LoneWolfHT:
```lua
(-m/2) - (dir.y*2)
```

This little bit of code reduces/increases the weight of the grenades depending on how far up/down the player is looking. Might remove if you don't think it works well

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,134 @@
+grenades = {}
+
+local function throw_grenade(name, player)
+	local dir = player:get_look_dir()
+	local pos = player:get_pos()
+	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.5, z = pos.z + dir.z}, name)
+	local self = obj:get_luaentity()
+
+	local m = 40
+	obj:set_velocity({x = dir.x * m, y = dir.y * 30, z = dir.z * m})
```
ClobberXD:
```suggestion
	obj:set_velocity({x = dir.x * m, y = dir.y * m, z = dir.z * m})
```

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,134 @@
...
+	obj:set_acceleration({x = 0, y = (-m/2) - (dir.y*2), z = 0})
```
ClobberXD: `y = (-m/2) - (dir.y*2)`

What does this do?

ClobberXD:
> ```lua
> (-m/2) - (dir.y*2)
> ```
> 
> This little bit of code reduces/increases the weight of the grenades depending on how far up/down the player is looking. I can remove if you don't like the effect

Interesting, will test.

ClobberXD:
I'm sorry, but the new behaviour seems to be inconsistent. Ideally, the initial velocity should be very high, and the acceleration, simulating air resistance and gravity, should be updated every step.

The x and the z components of acceleration should be set to `-k * dir`, where `dir` is the direction of the grenade at that point in time. So `-k * dir` is basically a force acting against the grenade's movement direction, a.k.a air resistance; the constant factor `k` needs to be decided upon by extensive testing. The y-component should just be -9.8.

mods/ctf/ctf_crafting/init.lua
```diff
@@ -206,3 +206,31 @@ for _, orex in pairs(full_ores) do
...
 	})
 end
+
+--
+--- Grenade Crafts
+--
+
+crafting.register_recipe({
+	type   = "inv",
+	output = "grenades:regular 1",
```
ClobberXD: This should probably either be `grenades:grenade` or `grenades:regular_grenade`, bc just `regular` isn't as intuitive.

mods/pvp/grenades/grenades.lua
```diff
@@ -0,0 +1,184 @@
...
+	local pos2 = vector.add(pos, radius)
+
+	for _, p in ipairs(minetest.find_nodes_in_area(pos1, pos2, "group:flora")) do
+		if vector.distance(pos, p) <= radius then
+			minetest.remove_node(p)
+		end
+	end
+end
+
+grenades.register_grenade("grenades:regular", {
```
ClobberXD: This should probably either be `grenades:grenade` or `grenades:regular_grenade`, bc just `regular` isn't as intuitive.

mods/pvp/grenades/init.lua
LoneWolfHT: Once the grenade's velocity reaches 0 the grenade will begin to rocket 'backwards'

LoneWolfHT:
I'll make those changes once I get on my computer. How does 'grenades:frag' sound?

LoneWolfHT:
The code for slowing the grenade velocity is on line 65+

mods/pvp/grenades/init.lua
ClobberXD: > Once the grenade's velocity reaches 0 the grenade will begin to rocket 'backwards'

If the suggestion is applied? I don't see how that's possible. This is just the initial velocity.

LoneWolfHT: For some reason I saw `set_acceleration` in there :thinking:
I guess just take my answer and apply it to
> The x and the z components of acceleration should be set to -k * dir, where dir is the direction of the grenade at that point in time. So -k * dir is basically a force acting against the grenade's movement direction, a.k.a air resistance

ClobberXD: I suggested moving `set_acceleration` to `on_step`. +y acceleration can be prevented by checking if y-velocity is > 0 before setting air resistance.

LoneWolfHT: > I suggested moving `set_acceleration` to `on_step`. +y acceleration can be prevented by checking if y-velocity is > 0 before setting air resistance.

I just did a `-math.abs()`

LoneWolfHT: Changes coming soon...

mods/pvp/grenades/grenades.lua
LoneWolfHT: `grenades:frag` in the incoming commits...

LoneWolfHT:
@ClobberXD I've mostly forgotten how you wanted the grenade physics. I went and watched some videos of grenades being thrown and used that to base how the grenade physics work
Ready for testing though

mods/pvp/grenades/init.lua
```diff
@@ -62,9 +67,8 @@ function grenades.register_grenade(name, def)
 
 			-- Fix accel bug
```
ClobberXD: What is the accel bug, exactly?

mods/pvp/grenades/init.lua
```diff
@@ -6,9 +6,14 @@ local function throw_grenade(name, player)
...
-	obj:set_velocity({x = dir.x * m, y = dir.y * 30, z = dir.z * m})
-	obj:set_acceleration({x = 0, y = (-m/2) - (dir.y*2), z = 0})
+	local dm = 1
+	if dir.y < 0 then
+		dm = -1
+	end
+
+	local m = 33
+	obj:set_velocity({x = dir.x * m, y = dir.y * m, z = dir.z * m})
+	obj:set_acceleration({x = 0, y = -math.abs(dm) * 30, z = 0})
```
ClobberXD: Use `-(dir.y / math.abs(dir.y))` instead of using a separate variable to get the sign. Tad less readable, but the extra variable can then go poof.

mods/pvp/grenades/init.lua
LoneWolfHT: The one where the grenade will shoot backwards after the acceleration brings the velocity of the obj to 0, I'll improve that comment...

mods/pvp/grenades/init.lua
LoneWolfHT: Ok

LoneWolfHT: I actually have no clue what I was doing there. Changed to -30

mods/pvp/grenades/init.lua
ClobberXD: Ah, ok

mods/pvp/grenades/init.lua
```diff
@@ -65,8 +60,7 @@ function grenades.register_grenade(name, def)
 				vel = obj:get_velocity()
 			end
 
-			-- Fix accel bug
-
+			-- Can't use set_acceleration() because the grenade will shoot backwards once the velocity reaches 0
 			vel.x = vel.x / 1.03
 			vel.z = vel.z / 1.03
```
ClobberXD: You can use [`vector.multiply`](https://github.com/minetest/minetest/blob/37923920a07f06d201662d8f1f4e5821efcc7b09/doc/lua_api.txt#L2628) instead of manually scaling each component, to make this code nice and more readable.

LoneWolfHT: `vector.multiply(vel, vector.new(1.03, 0, 1.03))`
This good?

ClobberXD: Oh wait, I thought you were modifying y as well. Hmm... how would this actually help fix the nade zooming backwards?

LoneWolfHT: When you're dividing it'll never pass 0

ClobberXD:
The frag grenade's sound can still be improved. Currently, it's somewhat softer, more like a cannon firing, and there's also another sound being played in the background some where at the middle.

LoneWolfHT:
> The frag grenade's sound can still be improved. Currently, it's somewhat softer, more like a cannon firing, and there's also another sound being played in the background some where at the middle. (Sounds to me like an empty shell falling on a hard floor)

Is it possible to use the tnt sound or shall I look into making a new sound?

LoneWolfHT:
It looks like the license for the tnt sound isn't known. So I'll have to find a new sound, make a new sound, or stick with this one

LoneWolfHT:
@ClobberXD How is this new sound?

mods/pvp/grenades/grenades.lua
```diff
@@ -0,0 +1,198 @@
...
+		})
+
+		remove_flora(pos, radius/2)
+
+		for _, v in ipairs(minetest.get_objects_inside_radius(pos, radius)) do
+			local hit = minetest.raycast(pos, v:get_pos(), true, true):next()
+
+			if hit and v:is_player() and v:get_hp() > 0 and hit.type == "object" and hit.ref:is_player() and
+			hit.ref:get_player_name() == v:get_player_name() then
+				v:punch(player, 2, {damage_groups = {grenade = 1, fleshy = 26 - (vector.distance(pos, v:get_pos()) * 2)}}, nil)
```
ClobberXD: I suggest calculating damage using the formula `25 * (0.707106 ^ dist) * 3`, suggested in https://github.com/MT-CTF/capturetheflag/pull/349#issuecomment-474963363, because linear damage is unrealistic. Please also increase the radius to at least 10 - the damage is >1 up to a radius of 12 using the suggested formula.

mods/pvp/grenades/grenades.lua
```diff
@@ -0,0 +1,198 @@
...
+			object_collision = false,
+			vertical = false,
+			texture = "grenades_boom.png",
+			glow = 10
+		})
+
+		minetest.sound_play("grenades_explode", {
+			pos = pos,
+			gain = 1.0,
+			max_hear_distance = 32,
```
ClobberXD: AFAIK, `max_hear_distance` is 32 by default. But I suggest increasing it to 64 here - it'd be nice if they can be heard from afar, however feeble they might sound.

mods/pvp/grenades/grenades.lua
LoneWolfHT: With the damage at 30 I'm brought to ~1 hp at a 3-4 node distance from the grenade
image

I may actually raise it a bit, because IMO the grenades should have a slightly larger radius due to them not detonating on collision

mods/pvp/grenades/grenades.lua
LoneWolfHT: Done in upcoming changes

mods/pvp/grenades/grenades.lua
LoneWolfHT: On second thought this is pretty good as it is...

mods/pvp/grenades/grenades.lua
```diff
@@ -67,7 +67,7 @@ grenades.register_grenade("grenades:frag", {
 
 			if hit and v:is_player() and v:get_hp() > 0 and hit.type == "object" and hit.ref:is_player() and
 			hit.ref:get_player_name() == v:get_player_name() then
-				v:punch(player, 2, {damage_groups = {grenade = 1, fleshy = 26 - (vector.distance(pos, v:get_pos()) * 2)}}, nil)
+				v:punch(player, 2, {damage_groups = {grenade = 1, fleshy = 30 * (0.707106 ^ vector.distance(pos, v:get_pos()) * 3)}}, nil)
```
ClobberXD: - `30 * (0.707106 ^ dist) * 3` might be a _little_ too much. Have you tested this?
- Should've suggested this earlier (sorry), but you might as well just combine the two numbers together:

  i.e. `90 * 0.707106 ^ dist` (The inner pair of parenthesis isn't required due to the order of operators - I checked)

LoneWolfHT: I tested and it worked fine IMO

ClobberXD:
@LoneWolfHT It'd be nice if you could rebase this PR on top of latest master - directly checking out this branch creates issues due to the recent `ctf_pvp_engine` merge.

olliy1or:
this looks complicated but it looks like its done. Good work

LoneWolfHT:
@ClobberXD Should be done now

mods/pvp/grenades/grenades.lua
```diff
@@ -0,0 +1,198 @@
...
+			hit.ref:get_player_name() == v:get_player_name() then
+				v:punch(player, 2, {damage_groups = {grenade = 1, fleshy = 90 * 0.707106 ^ vector.distance(pos, v:get_pos())}}, nil)
+			end
+		end
+	end,
+})
+
+-- Flashbang Grenade
+
+-- local flash_huds = {}
```
Panquesito7: Why not remove all these comments?

ClobberXD: Because they may be re-added at a later date after making a couple of changes.

LoneWolfHT: The flashbangs are a pretty big thing to add. I decided to try to get them working after this PR is merged

mods/pvp/grenades/init.lua
```diff
@@ -41,28 +41,35 @@ function grenades.register_grenade(name, def)
...
 
 				obj:set_velocity(self.last_vel)
 				vel = obj:get_velocity()
 			end
 
 			-- Can't use set_acceleration() because the grenade will shoot backwards once the velocity reaches 0
-			vel.x = vel.x / 1.04
-			vel.z = vel.z / 1.04
+			vel.x = vel.x / (1.04 + (self.hitcount/100))
+			vel.z = vel.z / (1.04 + (self.hitcount/100))
```
ClobberXD: Another reason why I feel grenades glide a little too much, is that the velocity never reaches 0. It'd be nice to directly set the velocity to 0 if it's below x (`0.1` would be a good value for starters).

mods/pvp/grenades/init.lua
```diff
@@ -41,28 +41,35 @@ function grenades.register_grenade(name, def)
 				self.last_vel = vel
 			end
 
+			if not self.hitcount then
+				self.hitcount = 0
```
ClobberXD: No need for this `if` block, _if_ (ha!) `hitcount` is directly initialised in the entity def.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,139 @@
...
+			end
+
+			-- Explode when clock is up
+
+			if self.timer > def.clock or not self.thrower_name then
+				if self.thrower_name then
+					minetest.log("[Grenades] A grenade thrown by "..self.thrower_name..
+					" is exploding at "..minetest.pos_to_string(vector.round(pos)))
+                    def.on_explode(pos, self.thrower_name)
+                end
```
ClobberXD: Whitespaces indents in lines 104 and 105. Please also surround `..` operators with spaces.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,139 @@
...
+			elseif def.particle and self.particle < def.particle.interval then
+				self.particle = self.particle + 1
+			end
+
+			-- Explode when clock is up
+
+			if self.timer > def.clock or not self.thrower_name then
+				if self.thrower_name then
+					minetest.log("[Grenades] A grenade thrown by "..self.thrower_name..
+					" is exploding at "..minetest.pos_to_string(vector.round(pos)))
```
ClobberXD: It's recommended to use simple present tense for consistency. i.e. "explodes", and not "is exploding"

ClobberXD:
The newly-added velocity modifiers make the 'nade bounce weirder than before. I think this hitcount-based velocity modifier should apply only to the y-component. Also, have you tried increasing the magnitude of the negative x and z components of acceleration, instead of directly modifying velocity?

mods/pvp/grenades/init.lua
LoneWolfHT: Eh, well I did that because it hasn't really exploded yet at the time of logging. Shall I add a 'Exploded!' log after the on_explode() call? Or just change the wording to present tense?

LoneWolfHT:
Hah, I pushed before testing. Lemme fix these buggy physics real qik...

mods/pvp/grenades/init.lua
ClobberXD: > it hasn't really exploded yet at the time of logging

I didn't suggest using past tense :P

And neither is the grenade in the process of exploding, when the message is being logged. If present continuous tense applies, then so would simple present tense. I suggested simple present tense because

- it's less verbose compared to simple present, in almost all cases.
- it's consistent with CTF, and MT as well.

ClobberXD:
Pro-tip: Don't directly modify velocity in `on_step`. Doing so will make the entity move in a *very* jerky manner, and it'd seem as if the grenade has a life of its own.

mods/pvp/grenades/init.lua
LoneWolfHT: Ok. Doing that then

LoneWolfHT:
@ClobberXD How's this?

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,140 @@
...
+			end
+		end
+
+		return itemstack
+	end
+
+	minetest.register_craftitem(name, newdef)
+end
+
+dofile(minetest.get_modpath("grenades").."/grenades.lua")
```
ClobberXD: Did you actually remove the spaces around the `..`? o_O

```suggestion
dofile(minetest.get_modpath("grenades") .. "/grenades.lua")
```

LoneWolfHT: I have no idea

LoneWolfHT: Fixed now

MinetestSam:
From my initial tests, everything seems to work well but eh smoke grenade inside water works quite weirdly.
screenshot_20191025_182931
screenshot_20191025_182937
 Oh and that smoke grenade sound gets a +1 from me :P

ClobberXD:
Unfortunately, particles spawning underwater that aren't visible from above is a known MT rendering bug. All we can do to prevent it from surfacing, is to avoid throwing grenades into the water, I suppose.

LoneWolfHT:
Did I forget to remove the shooter grenade? @MinetestSam

MinetestSam:
If you mean removing it from the treasure chests then yes it was removed but if you mean removing the whole code for shooter grenade then no.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,140 @@
...
+
+local function throw_grenade(name, player)
+	local dir = player:get_look_dir()
+	local pos = player:get_pos()
+	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.5, z = pos.z + dir.z}, name)
+	local self = obj:get_luaentity()
+
+	local m = 33
+	obj:set_velocity({x = dir.x * m, y = dir.y * m, z = dir.z * m})
+	obj:set_acceleration({x = 0, y = -27, z = 0})
```
ClobberXD: We, um... haven't yet migrated to Jupiter. We're still experiencing only 9.8 m/s^2 of gravity on Earth.

ClobberXD:
Grenades finally seem to be obeying the laws of Physics. :+1: 

But the grenade doesn't travel much before falling back. i.e. Resistance is too aggressive. Also, directly modifying velocity makes the grenade jerky. It's better to modify acceleration instead.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,140 @@
...
+				vel = obj:get_velocity()
+			end
+
+			if self.sliding == 0 and vel.y == 0 then
+				self.sliding = 1
+			elseif self.sliding == 1 and vel.y ~= 0 then
+				self.sliding = 0
+			end
+
+			-- Can't use set_acceleration() because the grenade will shoot backwards once the velocity reaches 0
```
ClobberXD: A simpler solution would be to check whether `velocity <= 0`, and nullify acceleration and velocity.

MinetestSam:
While testing earlier, I also found that you can't right click chests while having grenade as the wielded item

ClobberXD:
Yes, I noticed that too.

LoneWolfHT:
I set the reach to 0 so that in the heat of battle you wouldn't accidentally punch a node instead of throwing your grenade. It might be possible to change that. Shall I look into it?

mods/pvp/grenades/init.lua
LoneWolfHT: I would need to store the initial thrown direction and update that every time the grenade bounces off a wall. Since velocity is **direction x velocity**

mods/pvp/grenades/init.lua
LoneWolfHT: I couldn't get that to look realistic for some reason. object velocity may not be in m/s

LoneWolfHT: I couldn't get that to look realistic for some reason. object velocity may not be in m/s

ClobberXD:
Yes, please.

mods/pvp/grenades/init.lua
ClobberXD: Right, I was thinking of something else. I'll try to make the code cleaner, and if I succeed, I'll make a PR.

LoneWolfHT: Okie

LoneWolfHT: Okie

mods/pvp/grenades/init.lua
ClobberXD: That's because the initial velocity is pretty high, and there's 0 negative acceleration in both the `x` and `z` axes.

mods/pvp/grenades/init.lua
```diff
@@ -0,0 +1,140 @@
...
+					self.last_vel.x = self.last_vel.x * -0.5
+				end
+
+				if math.abs(self.last_vel.y - vel.y) > 6 then
+					self.last_vel.y = self.last_vel.y * -0.3
+				end
+
+				obj:set_velocity(self.last_vel)
+				vel = obj:get_velocity()
+			end
```
ClobberXD: Some more detailed comments would be useful to understand what this part of the code does.

LoneWolfHT:
Whoopsie. This might take a little bit.

ClobberXD:
Uh oh... I could help you fix the rebase, if you like.

LoneWolfHT:
I think I finally did it lol

ClobberXD:
Yaaay! :tada: 

You could also squash the commits. That way, rebasing would be a piece of cake (it still isn't particularly tasty though).

mods/ctf/ctf_treasure/init.lua
```diff
@@ -28,5 +28,16 @@ function ctf_treasure.get_default_treasures()
 
 		{ "medkits:medkit",       0.8, 5, 2 },
 		{ "ctf_bandages:bandage", 0.8, 2, { 2, 4 } }
```
ClobberXD: Missing `,`. Might've probably been botched during rebase.

https://travis-ci.org/MT-CTF/capturetheflag/builds/608845244#L310

mods/ctf/ctf_treasure/init.lua
```diff
@@ -28,5 +28,16 @@ function ctf_treasure.get_default_treasures()
...
+		{"grenades_basic:regular",0.1,2,1},
+		-- {"grenades_basic:flashbang",0.1,2,1},
+		{"grenades_basic:smoke",0.1,2,1},
+
+		{"shooter:machine_gun",0.02,2,1},
+		{"shooter:ammo",0.3,2,{1,10}},
+		{"shooter:arrow_white",0.5,2,{2,18}},
+
+		{"ctf_bandages:bandage",0.8,2,{2,4}},
+		{"medkits:medkit",0.8,5,2}
```
ClobberXD: Lines 36-41 are duplicate registrations, and should be removed.

mods/ctf/ctf_treasure/init.lua
```diff
@@ -27,17 +27,10 @@ function ctf_treasure.get_default_treasures()
 		{ "shooter:arrow_white", 0.5,  2, { 2, 18 } },
 
 		{ "medkits:medkit",       0.8, 5, 2 },
-		{ "ctf_bandages:bandage", 0.8, 2, { 2, 4 } }
+		{ "ctf_bandages:bandage", 0.8, 2, { 2, 4 } },
 
 		{"grenades_basic:regular",0.1,2,1},
 		-- {"grenades_basic:flashbang",0.1,2,1},
 		{"grenades_basic:smoke",0.1,2,1},
```
ClobberXD: Erm... spacing and alignment for these lines? They'd look cleaner.

LoneWolfHT: Not sure how that's possible if you're talking about the bandage/medkit lines

LoneWolfHT:
Guess I could have done the flashbang line too...
I'll just leave that for when I get around to doing the flashbangs

LoneWolfHT:
Currently working on fixing the fact that I don't use dtime in my velocity reduction calculations

mods/ctf/ctf_treasure/init.lua
ClobberXD: No, the grenade registrations. Maybe also move smoke grenade above the commented-out line, and align both the registrations using spaces. They're much easier to read that way.

LoneWolfHT: okie, already done lol

ClobberXD:
> Currently working on fixing the fact that I don't use dtime in my velocity reduction calculations. R.I.P if the server lags

Good idea! But server lag shouldn't matter if you're using `dtime`. In fact, it's quite the opposite - if you don't use `dtime` in your calculations, the code will not take into account the possibly varying delta-times and the end result would be messed up.

LoneWolfHT:
> > Currently working on fixing the fact that I don't use dtime in my velocity reduction calculations. R.I.P if the server lags
> 
> Good idea! But server lag shouldn't matter if you're using `dtime`. In fact, it's quite the opposite - if you don't use `dtime` in your calculations, the code will not take into account the possibly varying delta-times and the end result would be messed up.

Whoops, I guess I should have said 'Until then: R.I.P if the server lags' lol

ClobberXD:
..?! I rebased this PR, and it automatically closed

ClobberXD:
@LoneWolfHT Would you be able to run the following?

```
git fetch LoneWolfHT-master
git rebase LoneWolfHT-master master
```

P.S. Please create PRs from their own separate branches, and not from `master`. ;)

LoneWolfHT:
@ClobberXD I think you need to reopen

LoneWolfHT:
Alrighty nvm then

LoneWolfHT:
Something just broke lol

ClobberXD:
@LoneWolfHT Here's what I recommend:

- Decelerate entity using negative acceleration (`{ x = -m * dir.x, y = -9.8, z = -m * dir.z }`)
- Update acceleration every step, because the direction might change after bouncing.
- Obtain current dir, by storing a `last_pos`, and then `vector.direction(last_pos, obj:get_pos())`, and normalise it (`vector.normalize`).

LoneWolfHT:
@ClobberXD I think acceleration already takes dtime into account. So I'm not touching that.
Seems to work fine now. (Although I still think -9.8 acceleration looks terrible, especially when you throw a grenade straight upwards)
If you're wondering about that extra code in there: Without it the grenade will suddenly blast off in a random direction after coming to a full stop

ClobberXD:
Cool, will test

ClobberXD:
Grenade trajectory and collision has been improved by leaps and bounds, nice work!

All that remains now is to fine-tune some constants, which I'll get back to with, once I do more testing.

BTW, you might want to fix the treasure registration entries - the PR currently registers `grenades_basic:regular` and `grenades_basic:smoke`. ;)

ClobberXD:
> Tested- little problem with physics - grenades slide farther than expected

Agreed. I'm trying to arrive at the ideal velocity and deceleration constants that will make the grenades behave in realistic manner.

> the smoke takes longer than expected to disappear.

The smoke from smoke grenades, or the smoke from the blast of frag grenades? If it's the former, the smoke grenade lasts for 45s, and even that is a little on the lesser side, given that an M18 smoke grenade lasts from 50 to 90 seconds ([source](https://en.wikipedia.org/wiki/M18_smoke_grenade))

LoneWolfHT:
I think in the heat of battle the duration of the smoke will be a lot less noticeable

LoneWolfHT:
Don't mind the spam. Just a rebase

LoneWolfHT:
> I think in the heat of battle the duration of the smoke will be a lot less noticeable

After a little testing I decided to reduce the time by 10 seconds. The particles will last for a little over half a minute now. CTF is a little more fast-paced than real-life and there will be a good abundance of grenades too

LoneWolfHT:
Shall I just make grenades explode upon collision for now?

ClobberXD:
Physics seems a little better. I only noticed one oddity - grenades don't cover much lateral distance, when I'm throwing them at a steep angle. This happens IRL too, but the difference in-game is too high to be realistic.

ClobberXD:
@rubenwardy This was one of the key PRs to be a part of 2.0.0. This PR is 99% done, and a couple of physics tweaks is all that remains. I hope you don't mind if we merge this for 2.0.

rubenwardy:
That's fine

LoneWolfHT:
Ton of commits lol. Should be rebased though

ClobberXD:
Thanks for putting in your blood and tears to rebase this PR. ;)

@LoneWolfHT Please also note the following:

> Physics seems a little better. I only noticed one oddity - grenades don't cover much lateral distance, when I'm throwing them at a steep angle. This happens IRL too, but the difference in-game is too high to be realistic.

LoneWolfHT:
Had some old code to account for increased grenade speed

PR #359, opened 2019-03-06
Tweak medkits settings - Part 2

ClobberXD:
- Tweak medkits: Increase `regen_interval` to 0.5
  - This is because medkits are currently way too OP, since a player can now heal upto 20 HP within just a couple of seconds.
- Optimize medkits: Remove unwanted variable(s), since the HP target is now the same as `regen_max`.
- Update maps sub-module to the [latest commit](4665c084c6f9ea317e122898585c110e9abd3c3c)

DON'T SQUASH ME! `>:|`

### DIFF
--- a/mods/pvp/medkits/init.lua
+++ b/mods/pvp/medkits/init.lua
@@ -1,16 +1,19 @@
 -- Table of tables indexed by player names, each having three fields:
--- * HP at time of left-click + regen_max,
--- * Max HP of player (generally 20)
--- * ID of "healing effect" HUD element
+-- {
+--     hp:  HP at time of left-click + regen_max,
+--     pos: Position of player while initiating the healing process
+--     hud: ID of "healing effect" HUD element
+-- }
 local players = {}
 
-local regen_max = 20        -- Max HP provided by one medkit
-local regen_interval = 0.3 -- Time in seconds between each iteration
+local regen_max = 20       -- Max HP provided by one medkit
+local regen_interval = 0.5 -- Time in seconds between each iteration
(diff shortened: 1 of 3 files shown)


mods/pvp/medkits/init.lua
```diff
@@ -22,14 +25,14 @@ local function start_healing(stack, player)
 	end
 	local name = player:get_player_name()
 
-	if players[name] or player:get_hp() >= 20 then
+	if players[name] or player:get_hp() >= regen_max then
```
Thomas--S: Is this really correct if `regen_max` is less than 20?

Doesn't this mean that if for example `regen_max = 9`, players with 10 HP can't use the medkit to heal?

ClobberXD: Yes, that's how this is supposed to work.

Thomas--S: So `regen_max` is an absolute maximum? I thought it is the maximum amount of HP added to your current HP.

ClobberXD: It's taken to be an absolute maximum since 2789d141c369d373c9ed5262d19813b373d48fff

ClobberXD:
Updated the maps sub-module commit to include MT-CTF/maps@1595c6532532daf6745e42865db33f1b7dc1bfd8

mods/pvp/medkits/init.lua
Thomas--S: And how does 2789d14 affect the mod mechanics?

As far as I can see, it is only a settings change and in my opinion, we should keep the mod usable with different settings, too.

PR #365, opened 2019-03-08
Display actual kill weapons in kills list

ClobberXD:
screenshot_20190712_104651

### Use-cases

- Bragging rights: Imagine how awesome it'd be to kill an enemy with a sapling and have it show up in the kills list. :)

(Take the above screenshot for example - I pwned ClobberXD using furniture and bathroom supplies)

#### Checking for grenade kill

- If stack name is empty:
  - If `damage_groups.grenade` then kill weapon = `shooter:grenade`
  - Else, kill weapon = `""` (i.e. bare fists)

#### Retrieving image of kill weapon

- If kill weapon = `""`, image = `wieldhand.png`
- Else, select image from item def with the following order of preference:
  - If `inventory_image` and `inventory_image ~= ""`, image = `inventory_image`
  - Elseif `tiles[1]`
    - If `tiles[1]` is a table, image = `tiles[1].name`
    - Else, image = `tiles[1]`
  - Else image = `ctf_events_fallback.png` (fallback texture)

### To test

- Open up two instances of MT.
- Kill the other player with a variety of items (craftitems, tools, nodes, nodeboxes, etc.), _and_ with bare fists.
- Check the kills list to see if all the kill weapons are displayed correctly.


LoneWolfHT:
For the grenades I cound put something to hold the texture in the tool capabilities. You could then check for that

ClobberXD:
That's more of a workaround IMO. I think we should aim to completely fix this instead.

rubenwardy:
What happens if you punch someone to death?

ClobberXD:
> What happens if you punch someone to death?

I was thinking of showing a fist if the wielded item doesn't exist. But then, grenade kills have to be fixed first...

ClobberXD:
The grenades not showing up in the kills list actually deserves an issue of its own, as it's been that way since MT-CTF/ctf_pvp_engine@1c373b663b362498e035962ca6765b9e63873f29 - grenade kills have been showing up as sword kills since that commit. I don't understand how it was properly detected before that commit, as that commit basically just moved the kill weapon detection code down the call-stack.

ClobberXD:
Grenade kill detection has been fixed. Only issue no.2 remains to be fixed:

> Items without the `inventory_image` property doesn't work properly.

Meanwhile, how should punch-kills be displayed in the kills list? Is there a way to display the wieldhand? Should we add a fist texture?

ClobberXD:
Rebased.

LoneWolfHT:
> Grenade kill detection has been fixed. Only issue no.2 remains to be fixed:
> 
> > Items without the `inventory_image` property doesn't work properly.
> 
> Meanwhile, how should punch-kills be displayed in the kills list? Is there a way to display the wieldhand? Should we add a fist texture?

`wieldhand.png` (`minetest_game/default/textures/`)

ClobberXD:
Yes, but it isn't in the form of an inventory image, is it? It's _used_ to construct the inventory image. But I suppose we have to support displaying items that don't use `inventory_image` property. We could just display the raw texture used to create the tiles, for all items which don't have `inventory_image`, incl. the wieldhand itself.

LoneWolfHT:
> Yes, but it isn't in the form of an inventory image, is it? It's _used_ to construct the inventory image. But I suppose we have to support displaying items that don't use `inventory_image` property. We could just display the raw texture used to create the tiles, for all items which don't have `inventory_image`, incl. the wieldhand itself.

I think the wieldhand texture would look fine in the kill list

ClobberXD:
Added some magick to allow displaying items which don't have `inventory_image` defined.

I've updated the first post to replace the screenshot with a new one, and add implementation details. This PR is ready for review! :)

The wieldhand texture isn't very intuitive, btw. We should probably add a higher-res texture which better resembles a fist.

screenshot_20190712_105754

ClobberXD:
Thanks to @MinetestSam for testing this - nodeboxes and meshes are displayed as squares (as the code uses `tiles[1]`). I don't see any way to display these as HUD elements.

ClobberXD:
> Engine change needed

This label ain't gonna go any time soon...

LoneWolfHT:
What nodeboxes/meshes are there that don't have an inv image already?

LoneWolfHT:
Could always just default to the player's hand. Since that's technically what they're using anyway

ClobberXD:
> What nodeboxes/meshes are there that don't have an inv image already?

Good question... They *should* be having an inventory image. But I wonder why they didn't show up properly then - the code scans for `inventory_image` first, and only then resorts to `tiles[1]`.

> Could always just default to the player's hand. Since that's technically what they're using anyway

That'd be misleading. Better not to show it than to show the wrong item and mislead players.

mods/ctf/ctf_events/init.lua
```diff
@@ -122,16 +122,32 @@ end
...
+
+	-- Get inventory image of kill weapon
+	local image
 	if sname == "" then
-		if tool_caps.damage_groups.grenade then
-			type = "grenade"
+		image = "wieldhand.png"
+	else
+		local def = minetest.registered_items[sname]
+		image = (def.inventory_image ~= "") and def.inventory_image
```
ClobberXD: `inv_image ~= ""` would be `true` if `inv_image` is `nil` too, lol. I'll give this PR another shot.

ClobberXD:
> What nodeboxes/meshes are there that don't have an inv image already?

Welp, what was I thinking... These nodes not having an `inventory_image` was the original problem, lol. These nodes use `tiles`, and the inv-slot rendering is done on the engine-side. This obviously make it a lot more convenient for modders as they don't have to create images for each nodebox and meshnode, but there's also no way to render the images from within a mod.

PR #366, opened 2019-03-09
Fix normal -> unbreakable bug

Thomas--S:
There seems to be a critical bug regarding slabs/stairs and wool: When a player digs a normal slab/stair/wool and places it, it turns into a unbreakable one after placing.

This fix is tested, seems to work.

@rubenwardy, @ClobberXD Please review and merge as soon as possible.


ClobberXD:
Wool isn't supposed to be breakable in the first place. It's only meant to serve the purpose of decorating the map, and the player shouldn't be able to dig it at all. However, that's out of the scope of this PR. 

LGTM :+1:

Thomas--S:
Thanks for merging!

> Wool isn't supposed to be breakable in the first place

In my wall map, it is actually supposed to be breakable.

ClobberXD:
Yes, but I meant that it shouldn't be so - see #367

PR #370, opened 2019-03-11
Increase marker visibility time to 30 seconds

Thomas--S:
Seems like this was intended anyway:
Quote from https://github.com/MT-CTF/capturetheflag/pull/360#issue-258746976:

> Each marker is visible for 30s.

The current time seems way too short.

### DIFF
--- a/mods/ctf/ctf_marker/init.lua
+++ b/mods/ctf/ctf_marker/init.lua
@@ -2,7 +2,7 @@
 -- marker, because the members in the team needn't
 -- be the same within an extended duration of time
 local teams = {}
-local visibility_time = 10
+local visibility_time = 30
 
 -- Convenience function that returns passed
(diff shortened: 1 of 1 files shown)


ClobberXD:
The reason why I reduced the time from 30 to 10 is that 30 seemed way too long during testing. I guess 10 maybe a bit too less, but do you really want 30?

Thomas--S:
I think 30 is really necessary. It may seem long during singleplayer testing, but in a real match, you often have to finish your current fight then you have to orientate yourself to find the marker and then you have to follow it. This is nearly impossible in 10 seconds.

I actually even thought about 60 seconds (but this is too long, I agree) :smile:

Thomas--S:
Spamming is not possible, as there is only one marker per team.

PR #385, opened 2019-03-20
Exclude map's initial_stuff from list of treasures

ClobberXD:
Treasures that are already a part of map's `initial_stuff` aren't registered.

Screenshot from 2019-03-21 13-11-15

(Note: Coloured chat output was just for testing, and has been replaced by calls to `minetest.log` in the PR)

Tested. Works.

### DIFF
--- a/mods/ctf/ctf_map/schem_map.lua
+++ b/mods/ctf/ctf_map/schem_map.lua
@@ -207,9 +207,9 @@ function ctf_match.load_map_meta(idx, name)
 		i = i + 1
 	end
 
-	-- Register per-map treasures or the default set of treasures
+	-- Register per-map treasures, or the default set of treasures
 	-- if treasures field hasn't been defined in map meta
-	if treasurer and ctf_treasure then
+	if ctf_treasure then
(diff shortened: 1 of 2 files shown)


ClobberXD:
PR is ready for review. See OP for screenshot.

Thomas--S:
I think that stuff given in a greater amount than 1 in initial item shouldn't be omitted from the treasure chest. (e.g. torches, ammo, cobble etc.)

ClobberXD:
I had half the mind to fix that but thought I'd wait for more opinions. This is a very valid issue, and I'll fix this.

ClobberXD:
Done. A treasure is omitted from registration only if it's blacklisted _and_ its stack count is 1.

Thomas--S:
It might also be worth considering to whitelist steel pickaxes as they are very useful e.g. in the caves map to mine large tunnels etc. I often need more than one pickaxe there.

ClobberXD:
They can be crafted if required. That isn't something worth whitelisting, IMHO.

PR #390, opened 2019-03-22
Slow down players on flag pickup to 95% of normal speed

ClobberXD:
PR also takes other ways of losing flag into consideration while restoring original speed, like deaths and flag timeouts.

Semi-trivial. Tested - works.

### DIFF
--- a/mods/ctf/ctf_match/matches.lua
+++ b/mods/ctf/ctf_match/matches.lua
@@ -82,10 +82,20 @@ function ctf_match.create_teams()
 	error("Error! Unimplemented")
 end
 
+ctf_flag.register_on_pick_up(function(name)
+	physics.set(name, "ctf_match:flag_mult", { speed = 0.95 })
+end)
+
+ctf_flag.register_on_drop(function(name)
+	physics.remove(name, "ctf_match:flag_mult")
+end)
+
 ctf_flag.register_on_capture(function(attname, flag)
 	if not ctf.setting("match.destroy_team") then
 		return
 	end
+
+	physics.remove(attname, "ctf_match:flag_mult")
 
 	local fl_team = ctf.team(flag.team)
 	if fl_team and #fl_team.flags == 0 then
--- a/mods/ctf/ctf_match/mod.conf
+++ b/mods/ctf/ctf_match/mod.conf
@@ -1,3 +1,3 @@
 name = ctf_match
-depends = ctf, ctf_flag, ctf_inventory, ctf_alloc, vote, hudkit
+depends = ctf, ctf_flag, ctf_inventory, ctf_alloc, vote, hudkit, physics
 optional_depends = irc


ClobberXD:
I'm splitting the changes in `physics` mod to a separate PR. This can't be merged until those changes are in `master`

AKryukov92:
How long are the matches now?
How long they should be?

This feature will increase time to play one match because it will be easier to return the flag.
It is physically exchausting to play matches for several hours (like it was when I have played Caves last time).

ClobberXD:
This PR adds realism to the gameplay by making flag carriers marginally slower than normal players, as they should be. But nothing prevents players from guarding and protecting their flag carrier. This feature can't be blamed for the lack of teamwork and togetherness in the team. ;)

ClobberXD:
Bump!

LoneWolfHT:
I'll support this once the player speeds are a little faster

ClobberXD:
Is 90% a bit too slow? Should it be 95% instead?

ClobberXD:
Unfortunately I don't get notified on +1's, sorry for the delay ;)

Changed speed differential to 95%, and rebased the PR.

xenonca:
Tested, 95% seems good

Thomas--S:
To be honest, I'm not sure anymore if this concept is really helpful.

On some maps (Tunnel, Ice Age etc.) there are some extremly long matches that will nearly be unendable if this PR gets merged.

Instead, I suggest to make the relative speed change of the flag carrier configureable in the map configuration files.

This would allow to add the 95% slow-down on "easy" maps, but it could also allow to make flag carriers faster (~105%) on some very hard maps (e.g. maps similiar to Tunnel).

ClobberXD:
> Tested, 95% seems good

Glad to know, thanks for testing.

> Instead, I suggest to make the relative speed change of the flag carrier configureable in the map configuration files.

I really don't know what to say here... per-map flag carrier speeds doesn't make any sense to me. :)

95% is a sweet spot where the flag carrier is a wee bit slower than normal players, wherein someone tailing _right_ behind them can catch up, but is fast enough to escape unscathed almost every other time. I've tested this myself.

Thomas--S:
I agree with you that 95% seems to be a perfect value for "traditional" maps (2H2L, Bridge, Coast, Plains, ...).

However, I fear that this change will make captureing the flag on some difficult new maps with other concepts (e.g. Tunnel) nearly impossible.

If you are opposed to making the percentage configurable (which I don't think wouldn't be a bad idea necessarily, just think of different distances between the bases), I at least suggest to allow to disable the carrier-slowdown in the map configuration files.

ClobberXD:
> However, I fear that this change will make captureing the flag on some difficult new maps with other concepts (e.g. Tunnel) nearly impossible.

I can understand why you're hesitant about this. Have you actually tested the speed? Per-map settings are just not worth it for this simple feature, IMO. If you still feel that the speed delta is too high, I can increase the value up to 97% or 98%.

The whole point of this PR is to slow down the flag carrier by a wee infinitesimal bit, so less that most players won't even notice the difference in speed unless they're chasing the flag carrier in a straight line for more than 10 seconds.

Thomas--S:
Even small speed disadvantages will make the games last longer.

Given that these speed adjustments are very gentle and most players won't notice them anyway, I think they could be used very well to balance the match length, i.e. boost the flag carrier on maps with far too long matches and slow the flag carrier down on rather fast maps.

I think that these fine-tuning capabilities of such a feature could even be used to a much greater extent:
 - There could be a global flag carrier speed modificator that allows to adjust match durations in general. (This part is already implemented in this PR; although I very strongly suggest to not hardcode the value, but to make it configurable in `minetest.conf`.)
 - There could be a player-count or time dependent speed modificator to take account of the different match durations depending on how many players are online.
 - And there could also be the already suggested per-map modification to balance the different types of maps and prevent the game from becoming boring.

All of these three values can then be multiplied and applied to the flag carrier.


Regarding the implementation of such a feature: Without studying the relevant parts of code in detail, I'd suggest to allow to register callback functions for speed modifications, something like this (pseudo code):

```lua
ctf_flag.mod_functions = {}

function ctf_flag.register_flag_carrier_speed_modificator(mod)
    table.insert(ctf_flag.mod_functions, mod)
end

function ctf_flag.get_flag_carrier_speed(name)
    local speed = 1
    for k,v in pairs(ctf_flag.mod_functions) do
        speed = speed * v(name)
    end
    return speed
end

ctf_flag.register_on_pick_up(function(name)
	physics.set(name, "ctf_match:flag_mult", { speed = ctf_flag.get_flag_carrier_speed(name) })
end)

ctf_flag.register_flag_carrier_speed_modificator(name) -- Global
    return minetest_settings.get('flag_mult')
end

ctf_flag.register_flag_carrier_speed_modificator(name) -- Player count
    return some_fancy_formula(player_count)
end

ctf_flag.register_flag_carrier_speed_modificator(name) -- Map
    return ctf_map.current.setting_get('flag_mult')
end

```

Further possible use cases (I don't suggest to implement them now) would be to give a better speed modificator to disadvantaged teams or to disadvantage pro players under certain conditions etc.

ClobberXD:
Closing due to lack of time. Sorry.

PR #407, opened 2019-05-09
Fix server crash when vote kicking an already left player

### LINKED ISSUE #405: Crash due to missing player check before a vote-kick

Thomas--S:
Also fixes the /unblock command which didn't work previously

Tested and seems to work.

Fixes #405

### DIFF
--- a/mods/other/vote/vote_kick.lua
+++ b/mods/other/vote/vote_kick.lua
@@ -37,6 +37,15 @@ minetest.register_chatcommand("vote_kick", {
 
 		minetest.log("warning", "Player '" .. name .. "' started a vote" ..
 							" to kick '" .. param .. "'")
+
+
+		if not vlist[param] then
+			vlist[param] = {
+				ip = minetest.get_player_ip(param),
+				locked = false
+			}
+		end
+
 		return vote.new_vote(name, {
 			description = "Kick " .. param,
 			help = "/yes,  /no  or  /abstain",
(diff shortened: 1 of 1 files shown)


Thomas--S:
@ClobberXD PR updated according to your suggestions. Tested and seems to work. Thanks for your review.

mods/other/vote/vote_kick.lua
```diff
@@ -37,6 +37,15 @@ minetest.register_chatcommand("vote_kick", {
 
 		minetest.log("warning", "Player '" .. name .. "' started a vote" ..
 							" to kick '" .. param .. "'")
+
+
+		if not vlist[param] then
+			vlist[param] = {
+				ip = minetest.get_player_ip(param),
+				locked = false
```
ClobberXD: Is `locked` now required? To check if the player's kick vote is still active, you could do

```lua
if vlist[name] and minetest.get_player_by_name(name) then
    -- Player is a target of a kick vote, and the vote still hasn't ended (because the player is still in-game).
end
```

Or, you could define `vlist[name]` after the player has been kicked.

Thomas--S: I think that the current situation is more obvious and extendable.
The IP address can also be stored in a more understandable way with the current approach.

So, I'd prefer not to change it.

ClobberXD: Alright, that's fine by me. :+1:

Thomas--S:
**UPDATE:** Added a second commit to fix a server crash on priv grant.

PR #409, opened 2019-05-10
Add more items to blacklist

Thomas-AC-Dev:
Adds more definitions to team chests blacklist, which include dirt, the stone pickaxe and the stone sword.


mods/ctf/ctf_map/chest.lua
```diff
@@ -3,7 +3,10 @@ local blacklist = {
 	"default:jungleleaves",
 	"default:pine_needles",
 	"default:acacia_leaves",
-	"default:aspen_leaves"
+	"default:aspen_leaves",
+        "default:dirt",
```
Thomas--S: Wrong indentation.

Thomas--S:
I don't think that adding `default:dirt` to the blacklist is a good idea. It can be used for quick tower constructions and for terraforming.

Thomas-AC-Dev:
Will fix. Thank you.

Thomas-AC-Dev:
Removed dirt from blacklist.

mods/ctf/ctf_map/chest.lua
Thomas-AC-Dev: Fixed a while ago.

mods/other/report/init.lua
```diff
@@ -85,6 +85,6 @@ minetest.register_chatcommand("report", {
 			end
 		end
 
-		return true, "Reported. We'll get back to you."
+		return true, "Report has been sent."
```
ClobberXD: Ahem....

Thomas-AC-Dev: That has already been merged, I don't know why it still says that it is different from the official build, the official build should already have that changed.

ClobberXD: That's because you're using the same branch for three separate PRs. You must always use one branch for one PR only, and also leave the `master` branch untouched, so that it can be used to sync with the upstream repo.

You can manually undo the other changes in this PR, as the changes are very small. In case of larger PRs, it's way more difficult to separate commits into different PRs.

mods/ctf/ctf_map/chest.lua
```diff
@@ -3,7 +3,9 @@ local blacklist = {
 	"default:jungleleaves",
 	"default:pine_needles",
 	"default:acacia_leaves",
-	"default:aspen_leaves"
+	"default:aspen_leaves",
+        "default:sword_stone",
```
Thomas--S: Still incorrect indentation.

It seems like tabs are used in the rest of the file, while you are using spaces.

Thomas-AC-Dev:
Alright, I will close this pull request, I know there will be a conflict with torches in some maps. I do agree that stone tools may be useful, but I always hate it when noobs put their stone swords and picks in the team chest, filling spaces that could be occupied with other things.

PR #410, opened 2019-05-12
Move duplicated code to ctf_map.get_idx_and_map_by_input(param)

Thomas--S:
Tested and seems to work.

### DIFF
--- a/mods/ctf/ctf_map/schem_map.lua
+++ b/mods/ctf/ctf_map/schem_map.lua
@@ -55,19 +55,27 @@ function minetest.get_server_status(name, joined)
 end
 
 
+function ctf_map.get_idx_and_map(param)
+	param = param:lower():trim()
+	for i, map in pairs(ctf_map.available_maps) do
+		if map.name:lower():find(param, 1, true) or
+				map.path:lower():find(param, 1, true) then
+			return i, map
+		end
+	end
+end
+
 local next_idx
 minetest.register_chatcommand("set_next", {
 	privs = { ctf_admin = true },
 	func = function(name, param)
-		for i, map in pairs(ctf_map.available_maps) do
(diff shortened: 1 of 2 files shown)


mods/ctf/ctf_map/schem_map.lua
```diff
@@ -55,19 +55,28 @@ function minetest.get_server_status(name, joined)
...
 
+function ctf_map.get_idx_and_map_by_input(param)
+	param = param:lower()
+	for i, map in pairs(ctf_map.available_maps) do
+		if map.name:lower():find(param, 1, true) or
+				map.path:lower():find(param, 1, true) then
+			return i, map
+		end
+	end
+	return nil
```
ClobberXD: This line isn't required as all Lua functions return `nil` by default.

mods/ctf/ctf_map/schem_map.lua
```diff
@@ -55,19 +55,28 @@ function minetest.get_server_status(name, joined)
 end
 
 
+function ctf_map.get_idx_and_map_by_input(param)
+	param = param:lower()
```
ClobberXD:
```suggestion
	param = param:lower():trim()
```

mods/ctf/ctf_map/schem_map.lua
```diff
@@ -55,19 +55,28 @@ function minetest.get_server_status(name, joined)
 end
 
 
+function ctf_map.get_idx_and_map_by_input(param)
```
ClobberXD: It'd be nice to shorten this to `ctf_map.get_idx_and_map(param)`. "By input" is implied, as there's no other straightforward way to select the map.

Thomas--S:
Thanks for merging :)

PR #412, opened 2019-05-12
Fix server crash caused by /maps in some edge cases

Thomas--S:


mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -83,6 +83,9 @@ end)
 minetest.register_chatcommand("maps", {
 	privs = {interact = true},
 	func = function(name, param)
+		if #ctf_map.available_maps == 0 then
+			return false, "No maps are loaded!"
```
ClobberXD:
```suggestion
			return false, "No maps are available!"
```

Thomas--S:
Thanks for merging :)

PR #413, opened 2019-05-12
/maps: Add support for longer hints and licenses

Thomas--S:
The X and Y positions of the textarea may seem wrong, but I carefully adjusted those to look correctly.

The longer texts are especially important for the license strings because some licenses require the changes to be stated.

### DIFF
--- a/mods/ctf/ctf_map/maps_catalog.lua
+++ b/mods/ctf/ctf_map/maps_catalog.lua
@@ -8,7 +8,7 @@ local function show_catalog(name, idx)
 	-- Select map to be displayed
 	local map = ctf_map.available_maps[idx]
 
-	local fs = "size[10,8]"
+	local fs = "size[10,9]"
 
 	fs = fs .. "container[0,0]"
 	fs = fs .. "box[0,0;9.8,1;#111]"
@@ -20,12 +20,14 @@ local function show_catalog(name, idx)
 	end
 
 	-- Map name and author
-	fs = fs .. "label[3,0;" .. minetest.formspec_escape(map.name) .. "]"
-	fs = fs .. "label[5,0.5;" .. "(by " .. minetest.formspec_escape(map.author) .. ")]"
+	fs = fs .. "label[1.75,0;" ..
+			minetest.colorize("#ffff00", minetest.formspec_escape(map.name)) .. "]"
+	fs = fs .. "label[1.75,0.5;" .. minetest.colorize("#cccccc",
+					"by " .. minetest.formspec_escape(map.author)) .. "]"
 	fs = fs .. "container_end[]"
 
 	-- List of maps
-	fs = fs .. "textlist[0,1.2;3.5,6.8;maps_list;"
+	fs = fs .. "textlist[0,1.2;3.5,7.8;maps_list;"
 	for i, v in pairs(ctf_map.available_maps) do
 		local mname = v.name
 
@@ -50,12 +52,27 @@ local function show_catalog(name, idx)
 
 	-- Other fields
 	fs = fs .. "container[3.5," .. y + 0.5 .. "]"
-	fs = fs .. "label[0.5,0;" .. minetest.colorize("#FFFF00", "HINT:") .. "]"
-	fs = fs .. "textarea[0.8,0.5;5.5,1;;;" ..
-			minetest.formspec_escape(map.hint or "---") .. "]"
-	fs = fs .. "label[0.5,1.5;" .. minetest.colorize("#FFFF00", "LICENSE:") .. "]"
-	fs = fs .. "textarea[0.8,2;5.5,1;;;" ..
-			minetest.formspec_escape(map.license or "---") .. "]"
+	y = 0
+	if map.hint then
+		fs = fs .. "label[0.5," .. y .. ";" ..
+				minetest.colorize("#FFFF00", "HINT:") .. "]"
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.hint) .. "]"
+		y = y + 1.375
+	end
+	if map.license then
+		fs = fs .. "label[0.5," .. y .. ";" ..
+				minetest.colorize("#FFFF00", "LICENSE:") .. "]"
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.license) .. "]"
+		y = y + 1.375
+	end
+	if map.others then
+		fs = fs .. "label[0.5," .. y .. ";" ..
+				minetest.colorize("#FFFF00", "MORE INFORMATION:") .. "]"
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.others) .. "]"
+	end
 	fs = fs .. "container_end[]"
 
 	minetest.show_formspec(name, "ctf_map:maps_catalog", fs)
--- a/mods/ctf/ctf_map/maps_catalog.lua
+++ b/mods/ctf/ctf_map/maps_catalog.lua
@@ -45,14 +45,16 @@ local function show_catalog(name, idx)
 	local y = 1
 	if map.screenshot then
 		fs = fs .. "image[4,1.5;6.5,3.5;" .. map.screenshot .. "]"
-		y = y + 4
+		y = y + 3.5
 	end
 
 	-- Other fields
 	fs = fs .. "container[3.5," .. y + 0.5 .. "]"
-	fs = fs .. "label[0.5,0;HINT: " ..
+	fs = fs .. "label[0.5,0;" .. minetest.colorize("#FFFF00", "HINT:") .. "]"
+	fs = fs .. "textarea[0.8,0.5;5.5,1;;;" ..
 			minetest.formspec_escape(map.hint or "---") .. "]"
-	fs = fs .. "label[0.5,0.5;LICENSE: " ..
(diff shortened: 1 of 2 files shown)


mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -45,14 +45,16 @@ local function show_catalog(name, idx)
...
 	if map.screenshot then
 		fs = fs .. "image[4,1.5;6.5,3.5;" .. map.screenshot .. "]"
-		y = y + 4
+		y = y + 3.5
 	end
 
 	-- Other fields
 	fs = fs .. "container[3.5," .. y + 0.5 .. "]"
-	fs = fs .. "label[0.5,0;HINT: " ..
+	fs = fs .. "label[0.5,0;" .. minetest.colorize("#FFFF00", "HINT:") .. "]"
```
ClobberXD: Formspecs support `minetest.colorize`? I never knew! :O

ClobberXD:
Screenshot(s) please :)

Thomas--S:
screenshot_20190513_193317
screenshot_20190513_193328

ClobberXD:
`hint` is a very short 1-line hint for the players to know about the quirks of that map. I don't think it requires anything more than a label.

Thomas--S:
I'd prefer the textarea for hint for two reasons:
- consistency with the other fields
- Especially on small devices, the label will be cut off very soon, so that the second part of the line is hidden. Textareas allow the text to wrap; so the information will be better visible across all devices.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -45,15 +45,20 @@ local function show_catalog(name, idx)
...
+	fs = fs .. "textarea[0.8,0.5;5.5,1;;;" ..
 			minetest.formspec_escape(map.hint or "---") .. "]"
-	fs = fs .. "label[0.5,0.5;LICENSE: " ..
+	fs = fs .. "label[0.5,1.375;" .. minetest.colorize("#FFFF00", "LICENSE:") .. "]"
+	fs = fs .. "textarea[0.8,1.875;5.5,1;;;" ..
```
ClobberXD: Since textareas are substantially larger compared to labels, it's better if they're displayed only if the respective field exists, and omitted otherwise.

ClobberXD:
Both the reasons are valid, I have no issues with displaying all the meta fields as textareas.

Is `modifications` really needed? Do they _have_ to be listed out? Wouldn't attribution suffice? We can't afford to keep adding more textareas for every small detail. Maybe add an `others` textarea that can be used as a general purpose text display?

ClobberXD:
See
- https://creativecommons.org/licenses/by-sa/4.0/
- https://creativecommons.org/faq/#what-is-an-adaptation
- https://wiki.creativecommons.org/wiki/best_practices_for_attribution

Listing modifications isn't required, it seems.

Thomas--S:
I like the idea of using a `others` field for such things. I'll update the PR accordingly.

I'll also add the other suggestion to omit unused fields.

Thomas--S:
PR updated according to your suggestions. Thanks!

screenshot_20190514_090129
screenshot_20190514_090134
screenshot_20190514_090145
screenshot_20190514_090148
screenshot_20190514_090151
screenshot_20190514_090153

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -45,15 +45,29 @@ local function show_catalog(name, idx)
...
+		y = y + 1.375
+	end
+	if map.license then
+		fs = fs .. "label[0.5," .. y .. ";" .. minetest.colorize("#FFFF00", "LICENSE:") .. "]"
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.license) .. "]"
+		y = y + 1.375
+	end
+	if map.others then
+		fs = fs .. "label[0.5," .. y .. ";" .. minetest.colorize("#FFFF00", "MORE INFORMATION:") .. "]"
```
ClobberXD: Lines 55, 61, and 67 are a little too long. Please add a line-break to each of them.

JostP:
on small devices long `author` label is cut off:

gh

ClobberXD:
Which is why I asked you to keep it small ;)

Thomas--S:
I'll do this later this day.

Thomas--S:
screenshot_20190514_215037

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -20,12 +20,12 @@ local function show_catalog(name, idx)
 	end
 
 	-- Map name and author
-	fs = fs .. "label[3,0;" .. minetest.formspec_escape(map.name) .. "]"
-	fs = fs .. "label[5,0.5;" .. "(by " .. minetest.formspec_escape(map.author) .. ")]"
+	fs = fs .. "label[1.75,0;" .. minetest.colorize("#ffff00", minetest.formspec_escape(map.name)) .. "]"
+	fs = fs .. "label[1.75,0.5;" .. minetest.colorize("#cccccc", "by " .. minetest.formspec_escape(map.author)) .. "]"
```
ClobberXD: Lines 23 and 24 exceed 100 chars. Please add line-breaks. An 80-char limit is ideal, but it can be extended up to 90 if breaking at 80 or below causes readability issues.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -45,15 +45,32 @@ local function show_catalog(name, idx)
...
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.hint) .. "]"
+		y = y + 1.375
+	end
+	if map.license then
+		fs = fs .. "label[0.5," .. y .. ";" ..
+				minetest.colorize("#FFFF00", "LICENSE:") .. "]"
+		fs = fs .. "textarea[0.8," .. y + 0.5 .. ";5.5,1;;;" ..
+				minetest.formspec_escape(map.license) .. "]"
+		y = y + 1.375
```
ClobberXD: So much precision is unfortunately a waste of effort, as formspec coordinates are horribly inconsistent across various resolutions and DPI values. I suggest incrementing `y` by 1.5 to be on the safer side - formspec elements are sorta squashed together on Android, and more separation is better.

Thomas--S: If we use 1.5 steps, we need a larger size, which will cause even more problems as far as I know.

The precision actually works well, I think. Those numbers are multiples of 1/8.

ClobberXD: Have you tested this on a phone? Do the elements overlap?

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -20,12 +20,17 @@ local function show_catalog(name, idx)
...
 
 	-- Map name and author
-	fs = fs .. "label[3,0;" .. minetest.formspec_escape(map.name) .. "]"
-	fs = fs .. "label[5,0.5;" .. "(by " .. minetest.formspec_escape(map.author) .. ")]"
+	fs = fs .. "label[1.75,0;" ..
+			minetest.colorize("#ffff00", minetest.formspec_escape(map.name)) .. "]"
+	fs = fs .. "label[1.75,0.5;" ..
+			minetest.colorize(
+					"#cccccc",
+					"by " .. minetest.formspec_escape(map.author)
```
ClobberXD: Way too many line-breaks! :O

Something like this would be better:

```lua
		fs = fs .. "label[1.75,0.5;" .. minetest.colorize("#cccccc",
				"by " .. minetest.formspec_escape(map.author)) .. "]"
```

Both lines are still well under 80 chars.

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: We can always add a scrollbar instead of increasing formspec height, so that's not a problem.

mods/ctf/ctf_map/maps_catalog.lua
Thomas--S: I also thought about this version, but I didn't choose it, as the linebreak is inside of a function and thus not on a logical location.

It will make reading and understanding the code more difficult when using you suggestion, I think.

mods/ctf/ctf_map/maps_catalog.lua
Thomas--S: Screenshot_20190515-072209

It seems to work fine on Android in the current state.

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: Well, the whole of CTF's code-base uses this style of line-breaks, and so does Minetest (both Lua and C++). Too many line-breaks are even worse for readability. This looks less readable, IMHO.

Thomas--S: I still think that
```
[code]
[function call]
[code]
```
is more readable than
```
[code][function call part 1]
[function call part 2][code]
```
.

But I will update the PR according to your suggestion soon to match the criteria :)


BTW, thanks for your reviews!

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: That's not bad. We'll leave it at 1.375 for the present then. :+1:

mods/ctf/ctf_map/maps_catalog.lua
Thomas--S: Fixed. :)

Thomas--S:
Thanks for reviewing and mergin :+1:

PR #414, opened 2019-05-13
Add random message for maps catalog

Thomas--S:


mods/other/random_messages/init.lua
```diff
@@ -71,7 +71,8 @@ function random_messages.read_messages()
 		"Use /r to check your score and rank, and /rankings to see the league tables.",
 		"Use bandages on team-mates to heal them by 3-4 HP if their health is below 15 HP.",
 		"Use /m to add a team marker at pointed location, that's visible only to team-mates.",
-		"Use /summary to check scores of the current match and the previous match."
+		"Use /summary to check scores of the current match and the previous match.",
+		"Use /maps to view the maps catalog. You can also see license information there."
```
ClobberXD: Instead of specifically pointing out license info, you could say
	`"Use /maps to view the maps catalog, which contains information about every map."`
or smth.

Thomas--S: I'd prefer to mention the license information explicitely, so that legal problems can be avoided.

Please note: I am not a lawyer.

ClobberXD: Understood, that makes sense.

```suggestion
		"Use /maps to view the maps catalog. It also contains license info and attribution details."
```

Thomas--S:
Thanks for merging :)

PR #415, opened 2019-05-15
Add IRC support for /maps command

Thomas--S:
grafik

### DIFF
--- a/mods/ctf/ctf_map/maps_catalog.lua
+++ b/mods/ctf/ctf_map/maps_catalog.lua
@@ -78,6 +78,25 @@ local function show_catalog(name, idx)
 	minetest.show_formspec(name, "ctf_map:maps_catalog", fs)
 end
 
+local function send_irc_catalog(name, idx)
+	-- Select map to be displayed
+	local map = ctf_map.available_maps[idx]
+	local red = string.char(3) .. "4"
+	local normal = string.char(3)
+	minetest.chat_send_player(name, red .. "Map: " .. normal .. map.name)
+	minetest.chat_send_player(name, red .. "Author: " .. normal .. map.author)
+	if map.hint then
+		minetest.chat_send_player(name, red .. "Hint: " .. normal .. map.hint)
+	end
+	if map.license then
+		minetest.chat_send_player(name, red .. "License: " .. normal .. map.license)
+	end
+	if map.others then
+		minetest.chat_send_player(name,
+				red .. "More Information: " .. normal .. map.others)
+	end
+end
+
 minetest.register_on_player_receive_fields(function(player, formname, fields)
 	if not player or formname ~= "ctf_map:maps_catalog" then
 		return
@@ -105,15 +124,13 @@ minetest.register_chatcommand("maps", {
 		if #ctf_map.available_maps == 0 then
 			return false, "No maps are available!"
 		end
-		if not minetest.get_player_by_name(name) then
-			return false, "You must be online to view the maps catalog!"
-		end
 
 		-- Set param to nil if it's empty
 		if param and param:trim() == "" then
 			param = nil
 		end
 
+		local player = minetest.get_player_by_name(name)
 		local idx
 
 		-- If arg. supplied, set idx to index of the matching map name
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -116,6 +133,19 @@ minetest.register_chatcommand("maps", {
 
 		local idx
 
+		if not minetest.get_player_by_name(name) then
```
ClobberXD:
```suggestion
		if not minetest.get_player_by_name(name) and
				minetest.global_exists("irc") then
```

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -116,6 +133,19 @@ minetest.register_chatcommand("maps", {
...
+			minetest.chat_send_player(name, " *** CTF Map Catalog for IRC *** ")
+			if param then
+				idx = ctf_map.get_idx_and_map(param)
+			else
+				minetest.chat_send_player(name,
+						"No param supplied, showing information for current map.")
+				idx = ctf_map.map and ctf_map.map.idx or 1
+			end
```
ClobberXD: This exact `if-else` block already exists further down.

L157 can be changed like so:

```lua
if not minetest.get_player_by_name(name)
	show_irc_catalog
else
	show_catalog
end
```

Thomas--S: It is not exactly the same one because I decided to not use the `indices` value because it doesn't really make sense for IRC.

Additionally, the player gets notified about the omitted param, which would need another `if` block if we move it to the position suggested by you.

ClobberXD: That line can be changed to 

```lua
idx = (player and indices[name]) or ctf_map.map and ctf_map.map.idx or 1
```

Notifying the player about omitted param can have it's own small `if` block, that's fine. I'm just trying to avoid code duplication. :)

mods/ctf/ctf_map/maps_catalog.lua
Thomas--S: I thought about this again and I think it might be better to not use this check to make this more generic and usable with other, similiar mods.

If no such mod exists, nothing bad happens anyway, afaik.

ClobberXD: It's always better to be specific in cases where there's a chance for unexpected things to happen. A player executing commands while not online is an indication that they're on IRC, but things can change, and there would be many more situations where the player can execute commands while not being in-game. And it's also not like there's going to be another IRC mod anyway. :)

Thomas--S: I don't think that anything bad/unexpected can happen here. The worst case is that nobody reads the messages.

As far as I know, there are helper programs, which allow commands to be entered via terminal. If those programs implement everything well, it can also work there.

ClobberXD: Fine then, we'll leave it as it is.

Thomas--S: Thank you! :+1:

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -78,6 +78,26 @@ local function show_catalog(name, idx)
...
+	if map.hint then
+		minetest.chat_send_player(name, red .. "Hint: " .. normal .. map.hint)
+	end
+	if map.license then
+		minetest.chat_send_player(name, red .. "License: " .. normal .. map.license)
+	end
+	if map.others then
+		minetest.chat_send_player(name,
+				red .. "More Information: " .. normal .. map.others)
+
```
ClobberXD: Empty line at the end of a block of code; can be removed.

Thomas--S:
I removed the empty line, but I didn't add the tab-space.

I tested the tab-space, but some IRC clients (including mine) don't seem to be able to display it correctly. It'd look very weird with a tab-space.

I don't think it's necessary anyway because the colored keywords should be enough to ensure legibility.

ClobberXD:
I see, no need for it then.

PR #420, opened 2019-07-03
Fix furnace crash when dug

LoneWolfHT:
Tested and it seems to work

### DIFF
--- a/mods/mtg/furnace/init.lua
+++ b/mods/mtg/furnace/init.lua
@@ -242,6 +242,9 @@ local furnaces = {}
 
 local function on_destruct(pos)
 	local inv = minetest.get_inventory({ type = "node", pos = pos })
+
+	if not inv then return end
+
 	for _, list in pairs(inv:get_lists()) do
 		for _, item in pairs(list) do
 			minetest.add_item(pos, item)


ClobberXD:
Why would this even happen in the first place? :thinking: 

`on_destruct` of `furnace:furnace` is called only because a furnace exists. How can a furnace exist without an inventory?

Thomas--S:
Furnances placed by schematics don't have any meta. There are probably a few furnances in the new maps.

ClobberXD:
Quite likely. Then the correct fix would be to alias `furnace:furnace` to `default:stone`.

LoneWolfHT:
> Quite likely. Then the correct fix would be to alias `furnace:furnace` to `default:stone`.

That won't cause problems with the crafting/placing of furnaces? Not very familiar with how aliasing works...

ClobberXD:
Tested. Apparently, mods can't alias names of existing items to other names since 5.0.0 - I get a truck-load of warnings. This explains why we're seeing other removed items like grass and gold ore back in-game. The existing node-defs has to be removed first, before registering their respective aliases.

```
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:grass_1 -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:grass_2 -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:grass_3 -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:grass_4 -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:sand_with_kelp -> default:sand
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:grass_5 -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:bush_leaves -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:bush_stem -> air
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: default:stone_with_gold -> default:stone
2019-07-04 21:05:52: WARNING[Main]: Not registering alias, item with same name is already defined: furnace:furnace -> default:stone
```

So aliasing would not work, in this case. Then I feel that the correct thing to do would be to validate all nodes _during_ map export. While this PR would fix the crash, furnaces aren't supposed to be a part of map schems in the first place.

LoneWolfHT:
So close this or is there a chance of this being merged?

ClobberXD:
I don't think this is the right thing to do. It feels like a hack, IMO. Sorry.

As I said in my previous comment, node validation should take place during map export itself. The existing maps with furnaces chould be fixed manually.

Thomas--S:
I strongly suggest to submit this bugfix to https://github.com/minetest/minetest_game because the game must not crash when a furnance without meta is dug.

LoneWolfHT:
> I strongly suggest to submit this bugfix to https://github.com/minetest/minetest_game because the game must not crash when a furnance without meta is dug.

Ok

LoneWolfHT:
Actually, @Thomas--S this is only a bug for CTF. There was some code added which forgot to make sure the furnace node had an inventory

Thomas--S:
Under these conditions, I strongly recommend to reopen and merge this PR.

LoneWolfHT:
heh, I shouldn't have deleted the fork

ClobberXD:
You deleted the fork?! Don't you like CTF? How DARE you! :rage:

LoneWolfHT:
Whoops, 'branch'
I must have been eating at the time or something lol

Thomas--S:
Isn't there a "Restore branch" button?

LoneWolfHT:
Not sure. I'll look into this later...

PR #422, opened 2019-07-04
Improve log messages

ClobberXD:
- Add log messages where required/useful
- Remove unnecessary log messages
- Convert `print` statements to `minetest.log`
- Fix incorrect log level for some messages
- Trivial code-style fixes

****

I've tested this PR, and nothing is unintentionally broken.


ClobberXD:
I've also added `minetest.log` calls to all chat-commands, save a few unimportant ones.

ClobberXD:
This PR only touches logging calls, and fixes code-style here and there. MT doesn't crash, even after executing a ton of chat-commands. Extensive testing is not strictly necessary.

ClobberXD:
Rebased

ClobberXD:
Marking as high-priority, as it clears so much `print` junk from the terminal, which is a pain to skim through while testing stuff. Please review ASAP.

PR #423, opened 2019-07-04
Fix aliases

LoneWolfHT:
Tested and it seems to work.


ClobberXD:
Tested, works. No warning in console, and the aliased nodes don't appear in-game.

There's some other warnings about aliasing wool, but that's not related to aliasing itself.

PR #430, opened 2019-07-07
Make axes a decent weapon

MinetestSam:
I have improved the damage output on axes while also keeping in mind that it doesn't over power the swords. I think we require some variety and making axes a weapon with high damage output but also high punch interval will make it balanced enough. The aim is to encourage new strategies rather than the players using the same sword and shotgun combo. 

I haven't tested it yet and I am open to suggestions.


Thomas--S:
I'd prefer a `minetest.override_item()`in a CTF-related mod to avoid the modifications in Minetest Game (for easier updating).

ClobberXD:
I agree with Thomas--S. Updating MTG is a great pain already, and every time MTG's default/tools.lua is updated, we'd have to add these custom values back. I suggest creating a separate mod that depends on `default`, and overrides the tool caps like so:

```lua
local axes = {
	-- list of all axes
}

for _, name in pairs (axes) do
	local tool_caps = minetest.registered_tools[name].tool_capabilities
	tool_caps.full_punch_interval = 10
	tool_caps.damage_groups = {fleshy = 1000}
	minetest.override_item(name, { tool_capabilities = tool_caps })
end
```
WARNING: Untested code. Might get your PR rejected. Use at your own risk. :)

MinetestSam:
Dunno how but it made an unnecessary commit which I had reverted. But anyhow, I did test it and it seems that the damage output is a bit high. I will wait until you guys test it and what do you think about it.

mods/pvp/axe/depends.txt
```diff
@@ -0,0 +1 @@
+default
```
ClobberXD: `depends.txt` is legacy and won't work with 5.x. Replace this file with `mod.conf`, containing the following:

```
name = axe
depends = default
```

MinetestSam: Done.

MinetestSam:
I want to see how #223 and #258 are implemented and feel like this PR is just similar to those PRs

LoneWolfHT:
@MinetestSam Those other PRs need a rebase and some players were requesting for axes to be buffed a bit. This PR would probably get merged if you started working on it again

PR #434, opened 2019-07-11
Move crafting and ctf_pvp_engine into other/ and ctf/ respectively

ClobberXD:
Submodules `crafting` and `ctf_pvp_engine` are moved into `other/` and `ctf/`modpacks respectively.

Trivial PR. Tested; works.


ClobberXD:
Juggling submodules around the repo doesn't deserve a PR - it breaks git every time I change branches. I'll redo this change, and commit directly.

PR #436, opened 2019-07-12
Trivial changes to bandages

MinetestSam:
Spawn rate has been increased and now the person getting healed will get to know who healed them.

### DIFF
--- a/mods/ctf/ctf_treasure/init.lua
+++ b/mods/ctf/ctf_treasure/init.lua
@@ -23,7 +23,7 @@ function ctf_treasure.get_default_treasures()
 		{"shooter:ammo",0.3,2,{1,10}},
 		{"shooter:arrow_white",0.5,2,{2,18}},
 
-		{"ctf_bandages:bandage",0.3,5,{1,6}},
+		{"ctf_bandages:bandage",0.85,5,{2,4}},
 		{"medkits:medkit",0.8,5,2}
 	}
 end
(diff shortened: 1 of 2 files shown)


ClobberXD:
I'm also concerned about the target's chat being spammed with this message, when their team-mate heals them after a fight. I'll test this out, but I think a HUD alert would be a better implementation.

MinetestSam:
The chat can't be spammed more than 3-4 times since after the healing limit the bandages won't be of any use.

ClobberXD:
screenshot_20190712_204818

Seems fine to me (although we're running out of colours, lol).

It also makes logical sense, that this alert can't be spammed more than 4-5 times, as also confirmed by OP. So, :+1:

mods/ctf/ctf_treasure/init.lua
```diff
@@ -23,7 +23,7 @@ function ctf_treasure.get_default_treasures()
 		{"shooter:ammo",0.3,2,{1,10}},
 		{"shooter:arrow_white",0.5,2,{2,18}},
 
-		{"ctf_bandages:bandage",0.3,5,{1,6}},
+		{"ctf_bandages:bandage",0.85,5,{2,4}},
```
ClobberXD:
```suggestion
		{"ctf_bandages:bandage",0.8,2,{2,4}},
```

I've tested this, and there are a lot of bandages being spawned. The spawn rate is also a tad higher than medkits, as it should be.

Issue #131, opened 2018-03-09
Discussion regarding separation of team spawn from flag


ClobberXD:
Players could be made to spawn on the outer edge of their flag's protective 3-block radius. This prevents the players' selection boxes from interfering with the flag. Maybe randomize the same by having 4 spawn-points per team (one every corner of team base's square floor as shown below)

**`@  !  !  !  @`**
**`!  !  !  !  !`**
**`!  !  #  !  !`**
**`!  !  !  !  !`**
**`@  !  !  !  @`**

**`@`** Spawn-point
**`#`** Team-base / Flag position
**`!`** `ctf_team_base:ind_cobble`

afshaan4:
I agree on this idea, it would make the base less cluttered.

ClobberXD:
The four-corner implementation still makes the base a little cramped, and those who don't know the game will still remain in their bases throughout the match, making no difference whatsoever. What could be done to alleviate this is to randomize player spawn, so that players would spawn anywhere on their side of the map.

The only issue with this approach seems to be team-chest access. The team-chest formspec could be added as an sfinv page so that it can be accessed from anywhere by pressing `I`, but this ends up being a bit too OP.

Please express your opinions as to how the team-chest access issue can be resolved.

ClobberXD:
I'm planning to try out a multi-storey team-base implementation, with the flag on the top floor, and spawn + team-chest below (or vice-versa). Instead of giving 2 flags to a map-maker, two place-holder items should be given which can only be placed if there's enough space for the multi-storey team-base.

ClobberXD:
Separating the static spawn-point from team-base (a.k.a flag pos) requires a change in `ctf_pvp_engine`: rubenwardy/ctf_pvp_engine#21

ClobberXD:
Any thoughts on how team spawn can be separated from the flag?

- The 4-corner spawns idea in the OP isn't really feasible at the moment, as `ctf_pvp_engine` only supports a fixed offset from the flag.
- Random spawns isn't going to be a thing any time soon, as the spawnpoint selection needs to be very robust - we can't afford to spawn players in an underground cave because the algo thought it was the surface due to the presence of air. This can't be fixed by defining a static y-height as the surface, because of undulated terrain.

Thomas--S:
How about spawning players randomly in the 5x5 base area?

ClobberXD:
Best of both worlds, I like this simple yet effective solution. :)

[...]

PR #437, opened 2019-07-16
Implement random spawn within team base

ClobberXD:
Overrides `ctf.get_spawn(team_name)` to return a randomized position within the 3x3 area around the flag.

Idea provided by @Thomas--S in https://github.com/MT-CTF/capturetheflag/issues/131#issuecomment-511699910

#### To test

- Spam `/killme`

### DIFF
--- a/mods/ctf/ctf_map/base.lua
+++ b/mods/ctf/ctf_map/base.lua
@@ -41,3 +41,18 @@ function ctf_map.place_base(color, pos)
 	inv:add_item("main", ItemStack("default:glass 5"))
 	inv:add_item("main", ItemStack("default:torch 10"))
 end
+
+-- Override ctf.get_spawn to implement random spawns
+local old_spawn = ctf.get_spawn
+function ctf.get_spawn(team)
+	local spawn = old_spawn(team)
+	if not spawn then
+		return
+	end
+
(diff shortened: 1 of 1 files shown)


Thomas--S:
It seems like the `old_spawn` function can theoretically return nil.

This PR would introduce a crash for this scenario.
It might be best to check if old_spawn returns a vector, if yes, return the vector plus the random offset, else return the unmodified return value.

Thomas--S:
After thinking about this again, I suggest to only use the 3x3 area around the flag, i.e. using `math.random(-1, 1)` to avoid being trapped in structures in the corners (see screenshot; building is allowed there) or spawning inside the chest.

screenshot_20190716_214136

ClobberXD:
Players won't be trapped when their head is in a node, but I agree that the corners shouldn't be used.

ClobberXD:
Corners aren't selected any more now.

However I haven't checked for team chest, as it could be intensive. This is bad as respawning should be instantaneous, and players respawn at high rates. Such high frequency of `get_node` calls could potentially make the server lag.

ClobberXD:
Here's a map of how spawn positions are selected within the team-base:
```
  x x x  
x x x x x
x x x x x
x x x x x
  x x x  
```

PR #448, opened 2019-09-01
Remove shooter.conf from shooter's .gitignore, add basic shooter.conf

### LINKED ISSUE #447: Delete MTG tnt sounds if a mod uses them
If a mod is using sounds from the `tnt` mod (included in minetest_game), delete them immediately.
See [the Forums](https://forum.minetest.net/viewtopic.php?f=18&t=23154) for more information.

ClobberXD:
This PR:

- Removes `shooter.conf` from `.gitignore`

  The reason being that making the mod's configuration values public would allow people to recreate the exact same server experience. Another benefit of exposing config values is that developers and contributors can collaborate and improve performance-related values to improve the mod's fidelity and accuracy.

- Adds a basic `shooter.conf` that disables node blasting, crafting and also disables components like flare gun, turret gun, and rocket launchers, that are also disabled in the server.

This PR requires confirmation from @rubenwardy that the contents of the proposed `shooter.conf` is at least roughly equivalent to what's on the server's version of the file.

Tested, works.

### DIFF
--- a/mods/pvp/shooter/shooter.conf
+++ b/mods/pvp/shooter/shooter.conf
@@ -1,0 +1,15 @@
+
+# Disable node destruction with explosives
+SHOOTER_ENABLE_BLASTING = false
+
+# Disable node destruction
+SHOOTER_ALLOW_NODES = false
+
+# Disable extra components
+SHOOTER_ENABLE_FLARES   = false
+SHOOTER_ENABLE_HOOK     = false
+SHOOTER_ENABLE_ROCKETS  = false
+SHOOTER_ENABLE_TURRETS  = false
+
+# Disable crafting
+SHOOTER_ENABLE_CRAFTING = true
--- a/mods/pvp/shooter/.gitignore
+++ b/mods/pvp/shooter/.gitignore
@@ -4,5 +4,4 @@
 *bak*
 tags
 *.vim
-shooter.conf
 crosshair.png


Thomas--S:
Please split both changes into two PRs.

Replacing the sound has very high priority (please don't forget to update license information in readme files etc. if necessary), while the shooter.conf is not so high priority an controversial.

I think that shooter.conf shouldn't be included in the repo, as it has a very  similiar function to minetest.conf which is also not included.

I think that configuration and source code should be separated.

However, I think it might be a good idea to publish the contents of shooter.conf in a different place.

ClobberXD:
> Please split both changes into two PRs.
>
> Replacing the sound has very high priority (please don't forget to update license information in readme files etc. if necessary), while the shooter.conf is not so high priority an controversial.

Done - the sound file replacement change has been removed from this PR. Now, this PR is only about `shooter.conf`.

> I think that shooter.conf shouldn't be included in the repo, as it has a very similiar function to minetest.conf which is also not included.

The server's `minetest.conf` with server-related config values isn't a part of this repo, but the (sub-)game's `minetest.conf` containing game-related and mod-related config **is** a part of this repo. The reason is fairly simple - a (sub-)game's primary function includes dictating the style of gameplay, and that is also done by setting some config values of mods and modpacks. While a server could change stuff to suit their own style, CTF enforces a certain style of gameplay by default. See the contents of the included `minetest.conf`, for example - it contains values that are critical to the style of gameplay that CTF offers, like diplomacy, friendly fire, teams, flags, team-switching, etc.

Unfortunately, shooter doesn't parse `minetest.conf`, and only looks for `shooter.conf` in the modpath and worldpath. I consider that sub-optimal design, but I wouldn't bother changing that behaviour as it's going to be replaced by the submodule soon. But similar to `minetest.conf`, this file contains config values that shape the gameplay style, like node blasting, crafting, etc.

While I'm not at all in favour of adding server-specific settings to the repo, this is a completely different story, and the file in question adds settings that define the gameplay style of CTF, which I think should be clearly defined in the repo.

mods/pvp/shooter/shooter.conf
```diff
@@ -0,0 +1,15 @@
+
+-- Disable node destruction with explosives
```
Panquesito7: `.conf` files shouldn't have `--`.
Use `#` instead.

ClobberXD: Haha, good catch

ClobberXD: Fixed, thanks

ClobberXD:
shooter has been integrated as a sub-module into CTF. This PR is invalid now.

PR #449, opened 2019-09-02
ctf_chat: Add colored PMs

ClobberXD:
Continued from MT-CTF/ctf_pvp_engine#37

****

- Sender name is colorized according to their team color
- Message body is colorized according to the setting `ctf_chat.message_color`; defaults to `#E043FF`

screenshot_20190905_150832

screenshot_20190905_150652

### DIFF
--- a/mods/ctf/ctf_chat/init.lua
+++ b/mods/ctf/ctf_chat/init.lua
@@ -11,6 +11,34 @@ function minetest.is_player_name_valid(name)
 	return name:match("^[%a%d_-]+$")
 end
 
+-- Implement coloured PMs by overriding /msg
+-- The following code has been adapted from the chat-command of the same name in
+-- builtin/game/chat.lua of Minetest <https://github.com/minetest/minetest> licensed
+-- under the GNU LGPLv2.1+ license
+minetest.override_chatcommand("msg", {
+	func = function(name, param)
+		local sendto, message = param:match("^(%S+)%s(.+)$")
+		if not sendto then
+			return false, "Invalid usage, see /help msg."
+		end
+		if not minetest.get_player_by_name(sendto) then
+			return false, "The player " .. sendto .. " is not online."
+		end
+
+		-- Message color
+		local color = minetest.settings:get("ctf_chat.message_color") or "#E043FF"
+
+		-- Colorized sender name and message
+		local str =  minetest.colorize(color, "PM from ")
+		str = str .. minetest.colorize(ctf_colors.get_color(ctf.player(name)).css, name)
+		str = str .. minetest.colorize(color, ": " .. message)
+		minetest.chat_send_player(sendto, str)
+
+		minetest.log("action", "PM from " .. name .. " to " .. sendto .. ": " .. message)
+		return true, "Message sent."
+	end
+})
+
 local function team_console_help(name)
 	minetest.chat_send_player(name, "Try:")
 	minetest.chat_send_player(name, "/team - show team panel")


Thomas--S:
It seems like this function is copied from https://github.com/minetest/minetest/blob/master/builtin/game/chat.lua#L991in large parts.

I'd strongly suggest to add license information accordingly.

Please note: I am not a lawyer.

ClobberXD:
Hm, it didn't occur to me to add attribution to MT. Any suggestions as to how I'd go about doing that?

ClobberXD:
I've improved this PR a bit. See the first post for more details. I'm yet to add the attribution though.

Panquesito7:
One question: how can I colorize it to a color to my like? Let's say I want it to be blue.

ClobberXD:
> how can I colorize it to a color to my like? Let's say I want it to be blue.

Add the following setting to the server's `minetest.conf`: `ctf_chat.message_color = #3333FF`

That would produce a medium blue colour like so: `#3333FF`

Panquesito7:
> > how can I colorize it to a color to my like? Let's say I want it to be blue.
> 
> Add the following setting to the server's `minetest.conf`: `ctf_chat.message_color = #3333FF`
> 
> That would produce a medium blue colour like so: `#3333FF`

Cool! Thank you. 🙂

One more question: let's say I want to message you with purple, and you will message me with red.
E.g.: if I type `/msg ClobberXD purple Hi!`, then it'll send you a purple message.

Is that possible?

Thomas--S:
`ctf_chat.message_color` is a server side setting, as far as I can see.

So, it is the same color for all the players on a server, only the server admin can change the color for the whole server.

ClobberXD:
> One more question: let's say I want to message you with purple, and you will message me with red.
> E.g.: if I type /msg ClobberXD purple Hi!, then it'll send you a purple message.
>
> Is that possible?

That's very much possible, but it requires some extra code, a way to detect the colour from the chat command params, and a fallback colour if it isn't specified in the param. The purpose of `/msg` is to enable players to send private/one-to-one messages. Allowing players to specify a colour of the message is needlessly complicated, and to put it bluntly, a waste of dev time and effort.

OTOH, I'm not very sure about this, but I think you can achieve the desired effect using a CSM that colorizes the text before sending it. It'd likely be overridden when the mod colorizes the text, but you can give it a try, FWIW. For more info, check out https://github.com/minetest/minetest/blob/master/doc/client_lua_api.txt

Panquesito7:
> That's very much possible, but it requires some extra code, a way to detect the colour from the chat command params, and a fallback colour if it isn't specified in the param. The purpose of `/msg` is to enable players to send private/one-to-one messages. Allowing players to specify a colour of the message is needlessly complicated, and to put it bluntly, a waste of dev time and effort.

It must be a bit hard, but it'll be super easy if you set various colors (e.g.: blue, red, pink, ...).
Then, the player chooses one of those colors.

Anyway, as you said, it's a bit of waste of time, and will take you time to do it.
I like this PR BTW, coloured messages...love coloured messages. 🙂

ClobberXD:
I've added a comment within the code as attribution, see [diff](https://github.com/MT-CTF/capturetheflag/compare/12c50b892d12e1909dce3f6a9120862ea254248e..bc86562c945a3741c8ce7b8a3fb27c84bc23106f). This approach was finalised upon after [this conversation on IRC](http://irc.minetest.net/minetest-hub/2019-09-13#i_5590022)

Panquesito7:
> Anything else left to be addressed?

IMHO, it should be `LGPLv2.1+`.

ClobberXD:
Good catch

mods/ctf/ctf_chat/init.lua
```diff
@@ -11,6 +11,34 @@ function minetest.is_player_name_valid(name)
 	return name:match("^[%a%d_-]+$")
 end
 
+-- Implement coloured PMs by overriding /msg
```
Panquesito7: Just a small suggestion, but it'll short it a bit, IMHO.
```suggestion
--[[
Implement coloured PMs by overriding /msg
```
And remove all `-` to make it look better.

mods/ctf/ctf_chat/init.lua
```diff
@@ -11,6 +11,34 @@ function minetest.is_player_name_valid(name)
...
+-- The following code has been adapted from the chat-command of the same name in
+-- builtin/game/chat.lua of Minetest <https://github.com/minetest/minetest> licensed
+-- under the GNU LGPLv2.1+ license
```
Panquesito7: Same here.
```suggestion
under the GNU LGPLv2.1+ license.
--]]
```

mods/ctf/ctf_chat/init.lua
ClobberXD: 4 lines is too short to warrant comment blocks instead of line comments, IMO

Panquesito7: Well, IMO it looks weird and doesn't looks good using `--` each line.
But this is up to you: if you want to add it or not, anyway.

PR #450, opened 2019-09-02
Move suicide check to on_punchplayer callback in ctf

ClobberXD:
Continued from MT-CTF/ctf_pvp_engine#39

****

I've added suicide checks to different `on_killedplayer` callbacks on different occasions, the reason being that all `on_killedplayer` callbacks are designed in a way that require suicide checks. But adding a suicide check to every single callback is far from ideal. Ideally, this should be checked even before the `on_killedplayer` callbacks are run, and that's exactly what this PR does.

Needs testing to ensure nothing is broken.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -15,7 +15,7 @@ globals = {
 	"crafting", "vector", "table", "minetest", "worldedit", "ctf", "ctf_flag",
 	"ctf_colors", "hudkit", "default", "treasurer", "ChatCmdBuilder", "ctf_map",
 	"ctf_match", "ctf_stats", "ctf_treasure", "ctf_playertag", "chatplus", "irc",
-	"armor", "vote", "give_initial_stuff"
+	"armor", "vote", "give_initial_stuff", "hud_score"
 }
 
 read_globals = {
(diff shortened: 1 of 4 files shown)


ClobberXD:
`on_killedplayer` callbacks are only being registered in 4 places:

```
mods/ctf/ctf_metrics/init.lua:6:        ctf.register_on_killedplayer(function(victim, killer, type)
mods/ctf/ctf/teams.lua:454:        function ctf.register_on_killedplayer(func)
mods/ctf/ctf_events/init.lua:123:        ctf.register_on_killedplayer(function(victim, killer, stack, tool_caps)
mods/ctf/ctf_bounties/init.lua:87:        ctf.register_on_killedplayer(function(victim, killer)
mods/ctf/ctf_stats/init.lua:311:        ctf.register_on_killedplayer(function(victim, killer)
```

Will test today, although there's really not much to test here...

ClobberXD:
Pro-tip: Enabling "Hide whitespace changes" makes the diff much more simpler.

ClobberXD:
But sometimes suicide is a valid tactic and is allowed in the gameplay. While having to check for suicide in every callback is bad, ignoring suicides outright is just as bad. Maybe a more sensible solution is to pass an extra boolean to each callback, indicating whether the kill was a suicide. Or... maybe just leave things as-is.

PR #456, opened 2019-09-13
Parameterized /ctf_next; "Skip to map" and "Set as next map" buttons in maps catalog

ClobberXD:
This PR comprises of two commits, I've combined these commits because they kinda go hand-in-hand:

- The first commit adds support for directly passing a map name to `/ctf_next` for immediately skipping to the specified map. Earlier, this had to be done by first invoking `/set_next <map_name>` and then calling `/ctf_next`.
- The second commit adds "Skip to map" and "Set as next map" buttons to the maps catalog, which are the GUI-equivalent of `/ctf_next <map_name>` and `/set_next <map_name>` respectively.

screenshot_20190913_145322

#### First commit (parameterized `/ctf_next`)

- Check if `/ctf_next` skips to the next match and selects a map on its own. This behaviour should be no different from `master`.
- Test skipping to a certain map by invoking `/set_next <map>` and `/ctf_next`, and ensure that the process and outcome is no different from `master`.
- Now skip to a certain map by directly invoking `/ctf_next <map>`. The outcome should be identical to the previous step.

#### Second commit ("Skip to map" and "Set as next map" buttons)

- Test "Set as next map" by pressing the button and invoking `/ctf_next`. The outcome should be identical to invoking `/set_next <map>` and `/ctf_next`.
- Test "Skip to map" by pressing the button. The outcome should be identical to invoking `/ctf_next <map>`.

### DIFF
--- a/mods/ctf/ctf_map/maps_catalog.lua
+++ b/mods/ctf/ctf_map/maps_catalog.lua
@@ -2,10 +2,9 @@
 
 local indices = {}
 
-local function show_catalog(name, idx)
-	indices[name] = idx
-
+local function show_catalog(name)
 	-- Select map to be displayed
+	local idx = indices[name]
 	local map = ctf_map.available_maps[idx]
 
 	local fs = "size[10,9]"
@@ -23,7 +22,7 @@ local function show_catalog(name, idx)
 	fs = fs .. "label[1.75,0;" ..
 			minetest.colorize("#ffff00", minetest.formspec_escape(map.name)) .. "]"
 	fs = fs .. "label[1.75,0.5;" .. minetest.colorize("#cccccc",
-					"by " .. minetest.formspec_escape(map.author)) .. "]"
+			"by " .. minetest.formspec_escape(map.author)) .. "]"
 	fs = fs .. "container_end[]"
 
 	-- List of maps
@@ -77,14 +76,61 @@ local function show_catalog(name, idx)
 	end
 	fs = fs .. "container_end[]"
 
+	-- Add some special buttons, if player has ctf_admin priv
+	if minetest.check_player_privs(name, { ctf_admin = true }) then
+		fs = fs .. "button[4,8;2,1;btn_jump;Skip to map]"
+		fs = fs .. "button[6,8;2,1;btn_set;Set as next map]"
+	end
+
 	minetest.show_formspec(name, "ctf_map:maps_catalog", fs)
 end
 
-local function send_irc_catalog(name, idx)
+minetest.register_on_player_receive_fields(function(player, formname, fields)
+	if not player or formname ~= "ctf_map:maps_catalog" then
+		return
+	end
+
+	local name = player:get_player_name()
+
+	if fields.btn_prev then
+		indices[name] = indices[name] - 1
+		show_catalog(name)
+	elseif fields.btn_next then
+		indices[name] = indices[name] + 1
+		show_catalog(name)
+	end
+
+	if fields.maps_list then
+		local evt = minetest.explode_textlist_event(fields.maps_list)
+		if evt.type ~= "INV" then
+			indices[name] = evt.index
+			show_catalog(name)
+		end
+	end
+
+	if minetest.check_player_privs(name, { ctf_admin = true }) then
+		if fields.btn_set then
+			ctf_map.next_idx = indices[name]
+			minetest.chat_send_player(name, "Selected " ..
+					ctf_map.available_maps[ctf_map.next_idx].name .. " as next map.")
+		end
+
+		if fields.btn_jump then
+			ctf_map.next_idx = indices[name]
+			minetest.chat_send_player(name, "Skipping to next map")
+			ctf_match.next()
+		end
+	end
+end)
+
+local function send_irc_catalog(name)
 	-- Select map to be displayed
-	local map = ctf_map.available_maps[idx]
+	local map = ctf_map.available_maps[indices[name]]
+
+	-- IRC color codes
 	local red = string.char(3) .. "4"
 	local normal = string.char(3)
+
 	minetest.chat_send_player(name, red .. "Map: " .. normal .. map.name)
 	minetest.chat_send_player(name, red .. "Author: " .. normal .. map.author)
 	if map.hint then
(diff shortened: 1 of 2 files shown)


xenonca:
Tested and works well so far for me

ClobberXD:
Thanks for testing!

ClobberXD:
Rebased. Please review.

Thomas--S:
I will try to review this soon. I'm very interested in this.

ClobberXD:
That'd be great, thanks.

ClobberXD:
Rebased. Haven't yet tested after the rebase.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -107,19 +93,58 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
 	local name = player:get_player_name()
 
 	if fields.btn_prev then
+		indices[name] = indices[name] - 1
 		show_catalog(name, indices[name] - 1)
```
Thomas--S:
```suggestion
		show_catalog(name)
```

The second a parameter is being removed in this PR.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -107,19 +93,58 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
...
 	elseif fields.btn_next then
+		indices[name] = indices[name] + 1
 		show_catalog(name, indices[name] + 1)
```
Thomas--S:
```suggestion
		show_catalog(name)
```

The second a parameter is being removed in this PR.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -133,25 +158,24 @@ minetest.register_chatcommand("maps", {
...
 		end
 
+		indices[name] = indices[name] or (ctf_map.map and ctf_map.map.idx) or 1
+
 		if player then
-			show_catalog(name, idx or 1)
+			show_catalog(name)
 		else
 			minetest.chat_send_player(name, " *** CTF Map Catalog for IRC *** ")
 			if not param then
```
Thomas--S:
```suggestion
			if not param then
				indices[name] = (ctf_map.map and ctf_map.map.idx) or 1
```

This is needed to really show the current map in IRC if no param is specified.

ClobberXD: I haven't removed it completely, I've just moved it further up the file.

ClobberXD: I just tested, and it seems to work perfectly alright, both in-game and via IRC.

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: Done

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: Done

mods/ctf/ctf_map/maps_catalog.lua
Thomas--S: I think you are refering to (new) line 168.

However, this is not the same: In line 168, the last viewed map has first priority.
In IRC, we need to make sure that if no param is supplied, not the last viewed map is used, but the currently played map to ensure one can find out the current map via IRC. (/status is cut off often.) For this, you need to implement my suggestion above.

If you want to test this, use the following procedure for IRC:

1) `/maps` -> shows the currently played map :heavy_check_mark: 
2) `/maps <some other name>` -> shows the map <some other name> :heavy_check_mark: 
2) `/maps` -> shows <some other name>, although it should show the currently played map :x:

ClobberXD: Ah now I see what you mean. I thought showing the last viewed map was the correct behaviour ;)

ClobberXD: Do you want that to be the behaviour for the in-game maps catalog as well?

ClobberXD: Don't mind that question, it doesn't make sense to behave differently based on how the command is invoked.

Thomas--S: Actually, I'd prefer different behaviour in those different situations. If you are in-game, you usually know what the current map is or you can see it from the (un-truncated) /status output.

So I think the old behaviour should also be used in the new version, i.e. (if no param is specified):
 - In-Game: Show last viewed map (player probably wants to continue reading)
 - IRC: Show current map (last viewed map is in chat log anyways, player wants to know the current map)

ClobberXD: Done

ClobberXD:
Current order of precedence:

- Param (if specified)
- Current map (if `ctf_map.map` exists)
- Map at idx 1

Re-tested everything this PR offers, all works well.

Thomas--S:
Please see my latest comment above: https://github.com/MT-CTF/capturetheflag/pull/456#discussion_r344586612

The order of precedence should depend on the situation:

**In game:**
1) Param (if specified)
2) Last viewed map (player wants continue to read)
3) Current map (player hasn't open the catalog yet)
4) Map at idx 1 (fallback)

**IRC:**

1) Param (if specified)
2) Current map (player wants to know the current map, last viewed map is in chat log anyway)
3) Map at idx 1 (fallback)

mods/ctf/ctf_map/maps_catalog.lua
ClobberXD: > Actually, I'd prefer different behaviour in those different situations. If you are in-game, you usually know what the current map is or you can see it from the (un-truncated) /status output.

That's needlessly complicated, IMO. But ok...

Thomas--S:
Thank you, code looks good now, as far as I can see. I will try to test and approve later this day.

mods/ctf/ctf_map/maps_catalog.lua
```diff
@@ -133,25 +158,23 @@ minetest.register_chatcommand("maps", {
...
 
-		-- If arg. supplied, set idx to index of the matching map name
-		-- or path. Else, set to indices[name] or index of current map
 		if param then
-			idx = ctf_map.get_idx_and_map(param)
-		else
-			idx = (player and indices[name]) or ctf_map.map and ctf_map.map.idx or 1
+			indices[name] = ctf_map.get_idx_and_map(param)
+		end
+		if not player or not indices[name] then
```
Thomas--S:
```suggestion
		if not player and not param or not indices[name] then
```

Otherwise it will always (even when a parameter is specified) show the currently played map in IRC.

ClobberXD: Done

ClobberXD:
Applied latest suggestion. Tested (both in-game and on IRC), works as expected.

Thomas--S:
Thanks! I will try to review and approve it this evening or tomorrow.

PR #455, opened 2019-09-13
gauges: Remove breath indication; optimize HP setting code

ClobberXD:
This optimisation was suggested by @ranfdev in https://github.com/MT-CTF/capturetheflag/issues/320#issuecomment-529213581

****

This PR removes the breath indication from gauges (I didn't even know this existed until I started working on this), and updates the gauge entity in an `on_player_hpchange` callback instead of within `on_step`.

While this works as expected, I did notice *once*, that the HP indication kept alternating between the player's actual HP and full (20) while player was being healed by regen. So I've added a couple of debug print statements in a separate commit. Once this PR is confirmed to be working properly, I'll remove the second commit before merge.

PR #457, opened 2019-09-13
optimized gauges performance

ranfdev:
This is similar to #455, but should provide some more performance improvements, because it doesn't check continuosly for far away gauges, but checks for detached gauges only on player join.


ranfdev:
#320

ClobberXD:
Tested alongside `master` and #455. Benchmark results: https://gist.github.com/ClobberXD/0999b88a90a2eccddd17264cd626576c

ClobberXD:
https://github.com/MT-CTF/capturetheflag/pull/455#issuecomment-558887030

PR #458, opened 2019-09-14
Implement automatic door closure using node timers

ClobberXD:
This PR implements a very simple automatic door closure system using node timers, as an alternative to the relatively complex and lag-inducing idea proposed in #181. The auto-closure timeout, which defaults to 3 seconds, can be adjusted by defining the `doors.auto_closure_timeout` setting. This setting is overkill and probably not needed though, let me know if this should be removed.

- The timer is triggered every time a door is opened. If a timer is already running, it will be reset to the full timeout.
- The timer is aborted if the door is manually closed.
- Once the timer finishes, the node's `on_timer` callback is invoked, which simply calls the door toggle function again.

Tested; works perfectly on both left-hinged and right-hinged doors of all types.

****

Note: This PR also fixes a potential crash when `_doors.door_toggle(pos, node, clicker)` is called where `clicker=nil`; hence the "NO SQUASH".

### DIFF
--- a/mods/mtg/doors/init.lua
+++ b/mods/mtg/doors/init.lua
@@ -6,6 +6,9 @@ local _doors = {}
 _doors.registered_doors = {}
 _doors.registered_trapdoors = {}
 
+local auto_closure_timeout =
+		tonumber(minetest.settings:get("doors.auto_closure_timeout")) or 3
+
 local function replace_old_owner_information(pos)
 	local meta = minetest.get_meta(pos)
 	local owner = meta:get_string("doors_owner")
@@ -155,11 +158,20 @@ function _doors.door_toggle(pos, node, clicker)
 		return false
 	end
 
-	-- until Lua-5.2 we have no bitwise operators :(
+	-- States according to left-hinged or right-hinged doors
+	--
+	-- Left-hinged : 0 = closed, 1 = open
+	-- Right-hinged: 2 = closed, 3 = open
+
+	local nodetimer = minetest.get_node_timer(pos)
+
+	-- If state is not divisible by two, door is open. Else, door is closed.
 	if state % 2 == 1 then
 		state = state - 1
+		nodetimer:stop()
 	else
 		state = state + 1
+		nodetimer:start(auto_closure_timeout)
 	end
 
 	local dir = node.param2
@@ -420,6 +432,10 @@ function doors.register(name, def)
 
 	def.on_destruct = function(pos)
 		minetest.remove_node({x = pos.x, y = pos.y + 1, z = pos.z})
+	end
+
+	def.on_timer = function(pos)
+		_doors.door_toggle(pos, minetest.get_node(pos))
 	end
 
 	def.drawtype = "mesh"
--- a/mods/mtg/doors/init.lua
+++ b/mods/mtg/doors/init.lua
@@ -150,11 +150,8 @@ function _doors.door_toggle(pos, node, clicker)
 
 	replace_old_owner_information(pos)
 
-	local tname = ctf.player(clicker:get_player_name()).team
-	local owner_team = meta:get_string("owner_team")
-	local is_right_team = tname == owner_team
 	if clicker and not minetest.check_player_privs(clicker, "protection_bypass") and
-			not is_right_team then
+			ctf.player(clicker:get_player_name()).team ~= meta:get_string("owner_team") then
 		return false
 	end
 


Thomas--S:
I'm not sure if it is a good idea to merge this.

Auto-closing doors will lead to quite a few problems:
- When less experienced base builders build large rows of doors, it is usually most convenient to keep some of the doors open.
- Doors can be used in an inverted state, where the closed state is actually used as open and vice-versa.
- It might be a good idea to keep the door open in order to allow a team mate who captured the flag to enter the base quickly for ending the game.
- If your team's flag is already captured, it is important to keep the own door open in order to chase the capturer as fast as possible.
- Doors could be used in a two-way closing position, i.e.
```
|     |
|  D  o––––
|
+––––––––––
```
```
|     |
|     o––––
|     D
+––––––––––
```

Furthermore, I think that doors are already very strong. Enemy team players often prefer to break the walls – even three layered ones – instead of entering through the door. We shouldn't make doors even more powerful.

ClobberXD:
> When less experienced base builders build large rows of doors, it is usually most convenient to keep some of the doors open.

The solution is to remove the doors, not keep them open. ;)

> Doors can be used in an inverted state, where the closed state is actually used as open and vice-versa.

Well they... shouldn't be used that way. This isn't a valid argument against this PR, IMO.

> If your team's flag is already captured, it is important to keep the own door open in order to chase the capturer as fast as possible.
> Doors could be used in a two-way closing position

Very valid points. I can think of re-implementing this feature as a separate type of door - that way, players can use normal doors where required (e.g. two-way door systems) and not be bothered about them closing automatically. Auto-closing doors won't suit certain styles of bases, and players could freely opt for fully-manual doors. Thoughts?

Thomas--S:
> Very valid points. I can think of re-implementing this feature as a separate type of door - that way, players can use normal doors where required (e.g. two-way door systems) and not be bothered about them closing automatically. Auto-closing doors won't suit certain styles of bases, and players could freely opt for fully-manual doors. Thoughts?

This is indeed an interesting idea.

However, I think that a separate type of door will most likely cause a lot of confusion.
Instead, I'd suggest to add a auto-closer node (think of it as some kind of mechanism/machinery) that can be placed below (or above) a normal door to make it an auto-closing one.

LoneWolfHT:
Maybe pressure plates that take a while to release?

ClobberXD:
> However, I think that a separate type of door will most likely cause a lot of confusion.

It wouldn't, if it'd look different from manual doors.

> Instead, I'd suggest to add a auto-closer node (think of it as some kind of mechanism/machinery) that can be placed below (or above) a normal door to make it an auto-closing one.

The objective of this PR is to keep it simple. A separate node kinda complicates things by making doors having to check for an external factor to know whether to auto-close or not. It can be done, but it isn't the most elegant solution, IMO. However, I'm open to other ways of implementing this mechanism, provided they're simple and straight-forward.

> Maybe pressure plates that take a while to release?

This was suggested in #181, but... it's complicated. Again, this is an external factor, and there are a lot of permutations out of which only one would be conducive to opening/closing the door (e.g. the team of door placer and plate placer have to be the same, **and** the plate needs to be right next to the door). There's also the issue of multiple plates adjacent to a single door and vice-versa - Which plate opens which door? All of the aforementioned points mean that multiple node lookups would be involved every server step, apart from also checking whether a player is standing on any of the plates throughout the map.

TL;DR: Lot's of polling and checks involved. This idea could make the server buckle if already under load, which is why I didn't support #181 either.

Cahrs:
What about instead of having an open/closed state for the doors, have two different states that arent differentiated as open or closed? so if a player right-clicks a door, it will swap state, then after x seconds has elapsed will swap back states again.

ClobberXD:
> What about instead of having an open/closed state for the doors, have two different states that arent differentiated as open or closed? so if a player right-clicks a door, it will swap state, then after x seconds has elapsed will swap back states again, regardless of which state it was in.

I don't understand the point of this. Is this to fix the two-way door issue that Thomas-S pointed out? If the code doesn't differentiate states, this timer would also run when the toggle function is invoked to _close_ the door, opening it again after x seconds.

Cahrs:
> I don't understand the point of this. Is this to fix the two-way door issue that Thomas-S pointed out? If the code doesn't differentiate states, this timer would also run when the toggle function is invoked to close the door, opening it again after x seconds. 

Yes, but im not sure why that would have to be the case. i was thinking something like when the player rightclicks the door: increment value by two to set the corresponding open state (following how the code currently works), start a node timer, then after x seconds set the state to the value subtracted by two. Please correct me if im wrong

ClobberXD:
> when the player rightclicks the door: increment value by one to set the corresponding open state

How do you know if the door isn't already open? If the state of the door is 1 (i.e. left-open), incrementing it would set it to 2, which corresponds to right-closed

> ... (following how the code currently works), start a node timer, then after x seconds set the state to the value subtracted by one

If I understand you correctly, that sounds like a more complicated implementation, but one that's also potentially unsafe, as it doesn't explicitly check for the current state of the door.

ClobberXD:
There are a lot of valid points against automatic doors, and having two types of doors doesn't make much sense either. So, closing...

PR #461, opened 2019-09-26
Implement near-realtime tracking of team-mates

ClobberXD:
Video demonstration: https://youtu.be/PzZ56Cvu-mQ

Team-mates are tracked in near-realtime, thanks to periodically-updated HUD waypoints:

- A HUD waypoint is added for each player in the team, minus 1 (the player themselves).
  - A team of 10 would mean that each player would have 9 waypoints on their HUD, to track 9 of their team-mates.
- These waypoints' `world_pos` values are updated once every second, to the current position of their corresponding target players.

Here's the biggest disadvantage: Server load. While this feature could potentially be one of the best use-cases for SSCSM, this is implemented as a server-side mod at the moment - the server would have to process and send 180 (`10 * 9 * 2`) extra HUD updates every second (assuming there are 10 players in both teams). It's to be noted that the server already processes 180 (`20 * 6 * 3`) HUD elements for the kill history, but that's not updated every second.

This PR is just a proof-of-concept - I know that there's a high chance of this PR going down the drain, due to the potential performance constraints. While I haven't noticed any performance hits upon testing this with 2 players, things would be *very* different in a server environment with numerous players on either teams. One way to diminish the performance hit is to reduce the update interval to 2 or 3 seconds. The result would be less immersive, but more server-friendly.

### DIFF
--- a/mods/ctf/ctf_pos/init.lua
+++ b/mods/ctf/ctf_pos/init.lua
@@ -1,0 +1,126 @@
+-- mods/ctf/ctf_pos/init.lua
+
+local teams = {}
+--[[
+	Team-to-player mapping
+	----------------------
+
+	teams = {
+		team1 = {
+			name1, name2, name3
+		},
+		team2 = {
+			name4, name5, name6
+		}
+	}
+]]
+
+local players = {}
+--[[
+	Player-to-pos/player-to-team mapping
+	------------------------------------
+
+	players = {
+		name1 = {
+			pos  = {x =, y =, z =},
+			team = team_name
+		},
+		name2 = {
+			pos  = {x =, y =, z =},
+			team = team_name
+		}
+	}
+]]
+
+local UPDATE_INTERVAL = 1
+local hud = hudkit()
+
+ctf.register_on_new_team(function(team)
+	teams[team.data.name] = {
+		players = {}
+	}
+end)
+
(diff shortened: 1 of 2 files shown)


xenonca:
Minor Detail:
I'd change the text colour to white or something, else it will get difficult and confusing to spot the own base with the same colour (if lots of people are online).

MinetestSam:
Personally, I wouldn't want to know the location of each and every player since many of them can be new players who aren't pros. And even among the pros I wouldn't want to the location of every pro. Maybe a command like `/where xyz or /w xyz` would work better as I would only be interested in knowing the location of that one pro.

LoneWolfHT:
`/where` would be nice but what if an enemy player is using alts to find the locations of players on the opposite team?

LoneWolfHT:
Right now if I want to find the location of a player I ask them to place a marker at their pos

ClobberXD:
This PR is a proof-of-concept for a feature we'd take for granted on almost every other game. While it does allow for tracking team-mates, it does this in a very inefficient and ugly way. Hopefully we'd be able to achieve this in a better way in the future. Closing now...

> Maybe a command like /where xyz or /w xyz would work better as I would only be interested in knowing the location of that one pro.

I like this idea :+1:

PR #462, opened 2019-09-30
Improve ctf_traps and reinforced cobble: cheaper recipe's, higher-dmg, and texture-modifiers

Cahrs:
Changes:
Trapped blocks use the corresponding default textures with colorization via texture modifiers --the blocks aren't extremely obvious for texture-pack users anymore. Colorization is more faint, before it was so obvious you could easily tell the difference in very dim areas, making the blocks practically useless. 

Dmg-cobble uses the corresponding texture with colorization via texture modifiers.

Reinforced cobblestone now uses the default cobble texture with a reinforced frame overlay that was taken from the current reinforced cobble texture (made by ANAND?).

For crafting trap-blocks now have a 3:1 ratio of base-block:coal lump, making them a bit cheaper now.
Spikes now deal 7hp dmg instead of 5, which makes spikes more worth it to craft and consistent with dmg cobble.
Crafting recipe for spikes and dmg cobble yield 2 items and the recipe for both is a tad more expensive.

During the day:
screenshot_20190930_105801
During the night:
screenshot_20190930_105815
With a texturepack:
screenshot_20190930_112703


Cahrs:
Its hard to tell in the screenshot but if you look closely you can tell the difference between the trapped and normal types in the dark

mods/ctf/ctf_crafting/init.lua
```diff
@@ -147,7 +147,7 @@ crafting.register_recipe({
 
 crafting.register_recipe({
 	type   = "inv",
-	output = "ctf_traps:spike 1",
+	output = "ctf_traps:spike 2",
```
ClobberXD: AFAIK, @Thomas--S was against making spikes cheaper to craft.

ClobberXD:
I like the idea of using texture modifiers for this purpose. I'm okay with all the changes, except for the line comment that requires clarification from Thomas-S.

mods/ctf/ctf_crafting/init.lua
Thomas--S: I suggested to *make spikes cheaper* to craft: :smile: #358

Thomas--S:
I haven't test this in-game yet (and I probably won't do so soon), but I approve the concept of this PR.

If ClobberXD/ANAND tested this in-game and approves it too, I think this one can be merged.

mods/ctf/ctf_crafting/init.lua
ClobberXD: Ah ok, lol. Never mind :)

xenonca:
After applying my requested changes: 
screenshot_20191001_193711

Else +1

Cahrs:
I think the damage cobble should be less obvious then the ghost stone personally, id have the alpha at 15. It should also take a bit longer to dig since all the steel being used in it (my guess being the idea that there are spikes in the cobblestone that an enemy player wouldnt know about)

Cahrs:
Thinking on that again, i think it's simpler to just colorize the  dmg-cobble node *and* keep the alpha consistent with the other nodes, changes can be made after its merged if they are still to obvious.

Cahrs:
Updates: Damage cobble is now tinted red with modifiers as requested, and reinforced cobble now uses the default-cobblestone texture with an overlay (overlay taken from original texture by (ANAND?)

ClobberXD:
This PR is becoming a mess to review. It'd be nice if this could be split into separate PRs for texture modifiers, cheaper recipes, etc.

Cahrs:
> Personally, I like the red veins of damage cobble better than an overall overlay/tint.
So do i, but imo its the better choice as it prevents texture-pack users from having a serious advantage in that department

I can split it into smaller PR's later today

ClobberXD:
Oops, sorry

Cahrs:
Should i put them all up as PR's or one at a time? im not sure how to 'compare changes' between branches yet

ClobberXD:
> Should i put them all up as PR's or one at a time?

That's totally up to you.

> im not sure how to 'compare changes' between branches yet

If you mean you're not able to create a PR from the "Compare changes" page, check out this article: https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

rubenwardy:
I agree with increasing the damage of spikes, they should be pretty lethal. Reducing the cost too much could cause them to be overused, but that can always be tweaked

Cahrs:
Closing, i have no plans to finish this or work on ctf, best that people dont waste time reviewing/commenting.

PR #463, opened 2019-10-01
Add a 1000 score threshold to vote in vote kicks

MinetestSam:
Players with score above 1000 can vote in vote kicks but the players whose scores are below 1000 will not be able to vote. This is done to reduce players loging in through multiple accounts to prevent vote kick.

### DIFF
--- a/mods/other/vote/init.lua
+++ b/mods/other/vote/init.lua
@@ -164,8 +164,9 @@ vote.hud = hudkit()
 function vote.update_hud(player)
 	local name = player:get_player_name()
 	local voteset = vote.get_next_vote(name)
+	local player_stats, _ = ctf_stats.player(name)
 	if not voteset or not minetest.check_player_privs(name,
-			{interact=true, vote=true}) then
+			{interact=true, vote=true}) or voteset.can_vote and player_stats.score < 1000 then
 		vote.hud:remove(player, "vote:desc")
 		vote.hud:remove(player, "vote:bg")
 		vote.hud:remove(player, "vote:help")
--- a/mods/other/vote/init.lua
+++ b/mods/other/vote/init.lua
@@ -164,9 +164,8 @@ vote.hud = hudkit()
 function vote.update_hud(player)
 	local name = player:get_player_name()
 	local voteset = vote.get_next_vote(name)
-	local player_stats, _ = ctf_stats.player(name)
 	if not voteset or not minetest.check_player_privs(name,
-			{interact=true, vote=true}) or voteset.can_vote and player_stats.score < 1000 then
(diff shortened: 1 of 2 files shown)


mods/other/vote/init.lua
```diff
@@ -164,8 +164,9 @@ vote.hud = hudkit()
 function vote.update_hud(player)
 	local name = player:get_player_name()
 	local voteset = vote.get_next_vote(name)
+	local player_stats, _ = ctf_stats.player(name)
 	if not voteset or not minetest.check_player_privs(name,
-			{interact=true, vote=true}) then
+			{interact=true, vote=true}) or voteset.can_vote and player_stats.score < 1000 then
```
shivajiva101: should be 
` or voteset.can_vote and not voteset:can_vote(name)`
to retain objectivity. This change allows you to remove L167

mods/other/vote/vote_kick.lua
```diff
@@ -59,6 +59,15 @@ minetest.register_chatcommand("vote_kick", {
...
 			perc_needed = 0.8,
 
+			can_vote = function(self, pname)
+				local player_stats, _ = ctf_stats.player(pname)
+				local res = true
+					if player_stats.score < 1000 then
+						res = false
+					end
+				return res
+			end,
```
ClobberXD: This can be simplified to the following:

```lua
can_vote = function(self, pname)
	return ctf_stats.player(pname).score > 1000
end,
```

As `>` is a boolean operator, `score > 1000` is evaluated to either `true` or `false`, and that value is returned.

ClobberXD:
I support this concept. Will test and approve once the line comments are attended to.

MinetestSam:
Don't know why I wrote 'Simply the code' instead of 'Simplify the code' in the commit message. That was a derp from me.

mods/other/vote/init.lua
```diff
@@ -165,7 +165,7 @@ function vote.update_hud(player)
 	local name = player:get_player_name()
 	local voteset = vote.get_next_vote(name)
 	if not voteset or not minetest.check_player_privs(name,
-			{interact=true, vote=true}) then
+			{interact=true, vote=true}) or voteset.can_vote and not voteset:can_vote(name) then
```
ClobberXD: Inconsistent line-breaks. Please add line-breaks between two sub-conditions, instead of breaking right through them.

e.g.

```lua
if not voteset or
	not minetest.check_player_privs(name, {interact=true, vote=true}) or
	(voteset.can_vote and not voteset:can_vote(name)) then
```

ClobberXD: The lines are still too long. So I've sorta reverted this change, sorry for bothering you.

PR #467, opened 2019-10-07
Add colored random messages

MinetestSam:
Adds cyan colored random messages. This is to make sure that the random messages don't easily blend with the normal chat.
test
screenshot_20191007_163041


ClobberXD:
A screenshot with the coloured random messages would be nice.

ClobberXD:
The current colour blends in with the sky, and isn't very comfortable to read. A different hue would be nice.

PR #468, opened 2019-10-07
Better collision boxes

MinetestSam:
The aim of the mod is to change the collision box on death so that looting can be much easier and the collision box doesn't come in way while fighting another player


ClobberXD:
Code looks good to me, but if I may suggest an improvement:

- Set `collision_box[name]` to the `collisionbox` table in an `on_joinplayer` callback _once_. No need to set the same value every time a player dies.
- When a player dies, _only_ set their collision box to `{0,0,0,0,0,0}`.
- When a player respawns, _only_ set their collision box to `collision_box[name]`.
- When a player leaves, `collision_box[name] = nil`.

Thanks for this PR. Will test soon.

shivajiva101:
Personally I'd store it on first death and keep it till the player leaves the game, not all players die during a session...

ClobberXD:
> Personally I'd store it on first death and keep it till the player leaves the game, not all players die during a session...the question is does the conditional it would require take more resources than holding all the players collisionbox tables in memory.

That'd be ideal, but I share your concern about the performance differential too. It's for this reason I did not suggest this implementation.

OTOH, I think your approach handles edge cases better. e.g. mods modifying the collision boxes of certain players after they join. I'd prefer this approach too.

****

Btw: Tested, PR works as advertised.

PR #435, opened 2019-07-12
Improve directory structure of maps

ClobberXD:
Closes #352. Time to get rid of numeric prefixes for the map files.

- Each map will have its own sub-dir, containing
  - `map.conf` (used to be `<map_name>.conf`)
  - `map.mts` (used to be `<map_name>.mts`)
  - `screenshot.png` (used to be `<map_name>.png`)
  - `skybox_<n>.png` (used to be `<map_name>_skybox_<n>.png`)
- The `ctf_map` post-processing actions for maps has been moved into a dedicated shell script `setup_maps.sh`. This script appropriately renames all the textures to while copying them over to the mod's `textures/` sub-dir. e.g.
    - `cool_map/screenshot.png` ---> `cool_map.png`
    - `awesome_map/skybox_2.png` ---> `awesome_map_skybox_2.png`

****

#### Old directory structure

```
mods/ctf/ctf_map/maps
|
|-- 01_cool_map.conf
|-- 01_cool_map.mts
|-- 01_cool_map.png
|-- 02_awesome_map.conf
|-- 02_awesome_map.mts
|-- 02_awesome_map.png
|-- 02_awesome_map_skybox_1.png
|-- 02_awesome_map_skybox_2.png
|-- 02_awesome_map_skybox_3.png
|-- 02_awesome_map_skybox_4.png
|-- 02_awesome_map_skybox_5.png
\-- 02_awesome_map_skybox_6.png
```

#### New directory structure

```
mods/ctf/ctf_map/maps
|
|-- cool_map/
|   |
|   |-- map.conf
|   |-- map.mts
|   \-- map.png
|
\-- awesome_map/
    |
    |-- map.conf
    |-- map.mts
    |-- screenshot.png
    |-- skybox_1.png
    |-- skybox_2.png
    |-- skybox_3.png
    |-- skybox_4.png
    |-- skybox_5.png
    \-- skybox_6.png
```

PR #470, opened 2019-10-10
ctf_map: Auto-detect map's screenshot

ClobberXD:
If the file name of the screenshot is the exact same name as the map's other files, why not deduce the name automatically (like how we already do to get the schem file)? As we don't allow for different file-names, we might as well auto-deduce the file names of the screenshots instead of having the map maker explicitly define it in the map meta.

This PR doesn't change the behaviour of anything else. 

To test:

- Run `update.sh`, to copy screenshots to the mod's `textures/` dir.
- Launch CTF
- Open the maps catalog. Observe that the screenshots are displayed properly.

### DIFF
--- a/mods/ctf/ctf_map/schem_map.lua
+++ b/mods/ctf/ctf_map/schem_map.lua
@@ -97,7 +97,7 @@ local function load_map_meta(idx, path)
 		author        = meta:get("author"),
 		hint          = meta:get("hint"),
 		rotation      = meta:get("rotation"),
-		screenshot    = meta:get("screenshot"),
+		screenshot    = path .. ".png",
 		license       = meta:get("license"),
 		others        = meta:get("others"),
 		schematic     = path .. ".mts",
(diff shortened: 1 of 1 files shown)


ClobberXD:
> Looks good to me apart from that one single engine bug

Wrong PR?

MinetestSam:
> Wrong PR?

Yeah sadly I am quite dumb

Thomas--S:
I think the idea behind the setting was to allow maps to have no screenshot at all, but it would be okay for me to require a screenshot for every map.

ClobberXD:
@Thomas--S: Yes, screenshots were indeed originally intended to be optional. Then we quickly made it mandatory for all maps. But coming to think of it, I can tweak the code to actually check for the PNG file, similar to the skybox textures auto-detection implementation in #471.

ClobberXD:
Implemented in #435 instead.

PR #471, opened 2019-10-12
Add support for per-map skyboxes

ClobberXD:
- Skyboxes are optional.
- Skybox textures must follow this naming convention:

  `<mapname>_skybox_<1-6>.png`

- Thanks to the naming convention, `ctf_map` conveniently auto-detects the skybox textures, and handles player `set_sky` calls accordingly, without having to add any extra map meta fields.

### Screenshots

- Test skybox made using different textures from MTG's `wool` mod:
  
  screenshot_20191012_065417

- Proper skybox for `Intervention in the Iceage` (Changes ready to be published, once this PR is merged):
  
  screenshot_20191012_110115

  screenshot_20191012_110126

****

This is a [NO SQUASH] PR as it includes another commit that optimises map parsing for more flexibility with regards to sub-directory support.

Tested; works (as if the screenshots weren't enough proof :P)


Thomas--S:
:+1: 

It looks really impressive, but the night seems a little odd to me.

I fear we have to wait for better set_sky options in minetest.

ClobberXD:
Using a skybox is optional, but if a map maker does opt for a skybox, they should be fully aware of the consequences it has (incl. the map potentially having fixed time).

A more customisable `set_sky` would take ages to get merged (even if it's already open). Once it does, we can allow skybox-equipped maps to have a day-night cycle, but I don't think we need to wait until then.

Thomas--S:
I'm not sure, but this one might help: minetest/minetest#8964

It seems like this probably won't take ages to get merged anymore.

ClobberXD:
That's the PR I was referring to.

> It seems like this probably won't take ages to get merged anymore.

I wish you're right, but I bet you 500 free kills it would :P

But OTOH, I don't think our usecase would even benefit from the proposed improvements. How does the `set_sky` PR improve bright-daytime-resembling skyboxes looking weird at night?

Thomas--S:
Wouldn't it allow us to use the sun part of the skybox as a moving sun via set_sun?

While this would still not be perfect, it would improve the night appearance of skyboxes a lot.

It seems very weird to me when I see a bright sun during night time. :smile:

ClobberXD:
> Wouldn't it allow us to use the sun part of the skybox as a moving sun via set_sun?

That'd be _way_ more awkward than it already is. We'd have 2 suns, one stuck in the same place, and one moving across the sky. CTF doesn't yet simulate the experience of orbiting a binary star system, so that'd be _really_ weird. :P

And then the moon would rise. Umm... a moon? When there's a super-bright sun right next to it? :D

But that makes me think about the possibilities of also allowing perma-night battles, alongside perma-day battles. This would make a lot of sense, and could improve immersiveness on certain maps by a lot. Let me know what you think about being able to optionally specify a fixed time of day per-map. That'll have to come as a separate PR though - it's a major feature all by itself.

Thomas--S:
> That'd be way more awkward than it already is. We'd have 2 suns, one stuck in the same place, and one moving across the sky. CTF doesn't yet simulate the experience of orbiting a binary star system, so that'd be really weird. :P
>
> And then the moon would rise. Umm... a moon? When there's a super-bright sun right next to it? :D

How about replacing the (fixed) sun of the skybox with blank sky then via image manipulation? This was what I originally meant, I thought it was obvious not to have two suns :laughing: 


I appreciate your idea of introducing the possibility for fixed-time maps, where the time can be specified in map.conf.

I think skyboxes in the current state with the limited API only make sense in conjunction with this suggested feature, so I strongly suggest to introduce the fixed-time feature before we merge this PR.

ClobberXD:
> How about replacing the (fixed) sun of the skybox with blank sky then via image manipulation?

Why even wait for an improved `set_sky` then, lol? Isn't that how it already is? :)

> I think skyboxes in the current state with the limited API only make sense in conjunction with this suggested feature, so I strongly suggest to introduce the fixed-time feature before we merge this PR.

Righty-ho. Will do.

Thomas--S:
>>    How about replacing the (fixed) sun of the skybox with blank sky then via image manipulation?
>
> Why even wait for an improved set_sky then, lol? Isn't that how it already is? :)

My suggestion is to replace the fixed sun of the skybox with blank sky and to use the sun part of the skybox as a moving sun via set_sun. Afaik, this isn't possible yet. :)



---

>>    I think skyboxes in the current state with the limited API only make sense in conjunction with this suggested feature, so I strongly suggest to introduce the fixed-time feature before we merge this PR.
>
>Righty-ho. Will do.

Thanks. :+1:

ClobberXD:
> My suggestion is to replace the fixed sun of the skybox with blank sky and to use the sun part of the skybox as a moving sun via set_sun

Aaahhh, I misunderstood, as usual.

ClobberXD:
Rebased. Combined `time.lua` (from #478) and `skybox.lua` into `time_sky.lua` (well, "time and sky", so...).

Thomas--S:
I approve the concept of this PR now, but I'd like to test it before it's merged.

A small suggestion after skimming through the code:

How about making a helper function to check for the existence of a specified file, so that this rather ugly `table.indexOf() ~= -1 and table.indexOf() ~= -1 and table.indexOf() ~= -1 and ...` can be avoided?

Such a function could then also be used to implement the screenshot auto-detection as discussed in #470

ClobberXD:
> How about making a helper function to check for the existence of a specified file, so that this rather ugly table.indexOf() ~= -1 and table.indexOf() ~= -1 and table.indexOf() ~= -1 and ... can be avoided?

Good idea. Btw, `table.index_of` is still a pretty good way to check if a given name is the value of any key in a table. We'd have to use a `for` loop otherwise (ofc, `table.index_of` internally uses a `for` loop too).

Thomas--S:
Yeah, I'm not opposed to having one `table.indexOf` call in a helper function, but what I meant with "ugly" was this repetitive style of the code in conjunction with `table.indexOf`.

ClobberXD:
Requested changes have been attended to. I've added a helper function `ctf_map.file_exists`, that supports looking up a single filename, or also a list of filenames as a table. PR tested again, in conjunction with MT-CTF/maps#60. It works :tm:

Thomas--S:
I think this PR looks pretty good now.

I'll try to review it in about a week (I won't have time until then).

ClobberXD:
Awesome, thanks.

mods/ctf/ctf_map/init.lua
```diff
@@ -6,6 +6,22 @@ function ctf_map.get_team_relative_z(player)
...
 
+-- Convenience function to check whether a file (or multiple files) exists in mapdir
+function ctf_map.file_exists(subdir, target)
+	local list = minetest.get_dir_list(ctf_map.mapdir .. subdir, false)
+	if type(target) == "string" then
+		return table.indexof(list, target) ~= 1
+	elseif type(target) == "table" then
+		for _, filename in pairs(target) do
+			if table.indexof(list, filename) == -1 then
+				print("\t" .. filename .. " not found!")
```
Panquesito7:
```suggestion
				minetest.log("warning", "\t" .. filename .. " not found!")
```

ClobberXD: Again, debug message, it'll be dropped before merge.

ClobberXD:
Note: Debug messages need to be removed before merge

ClobberXD:
Debug message removed, will merge today.

mods/ctf/ctf_map/schem_map.lua
```diff
@@ -80,9 +80,9 @@ minetest.register_chatcommand("set_next", {
 	end,
 })
 
-local function load_map_meta(idx, path)
-	minetest.log("info", "load_map_meta: Loading map meta from \"" .. path .. "\"")
-	local conf_path = mapdir .. path .. ".conf"
+local function load_map_meta(idx, path, filename)
+	minetest.log("info", "load_map_meta: Loading map meta from \"" .. path .. filename .. "\"")
```
Panquesito7: Isn't this a debug message, too?

ClobberXD: No. Well, in a sense yes, but it can remain as it's not specific to the testing of this PR.

PR #473, opened 2019-10-12
Add indestructible torch [ctf_map:torch]

ClobberXD:
More MTG copy-pasta! This time, from `default/torches.lua`. :D

This PR adds 3 variants of indestructible torches - normal, ceiling-mounted, and wall-mounted. These behave like regular torches in every way, except for their indestructibility.

Tested; works. To test:

- `/giveme ctf_map:torch <how many ever you want>`.
- Run around, placing torches on floors, on walls, and on ceilings.
- Do let me know if you happened to notice something unintended.

### DIFF
--- a/mods/ctf/ctf_map/nodes.lua
+++ b/mods/ctf/ctf_map/nodes.lua
@@ -76,7 +76,7 @@ do
 	})
 end
 
--- Indestructible nodes from default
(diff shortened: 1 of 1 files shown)


ClobberXD:
A wild rubenwardy appeared! :)

PR #475, opened 2019-10-15
ctf_map: Convert to modpack; split map maker into separate mod

ClobberXD:
This PR performs some refactoring in `ctf_map`, mainly converting it into a modpack, and splitting `map_maker.lua` into a separate `map_maker` mod that depends on the now-renamed `ctf_map_core`. Note that map makers would still have to install the whole modpack, because of the aforementioned dependency.

### Rationale

I'm already working on 3-4 map maker features, and realised that the code has already gotten too messy to add any features while retaining readability. So the primary reason for this refactor is to allow adding more features to map maker, but without having to cram 2000 lines in a single file, which would end up looking like Martian vomit. i.e. this refactor would allow future PRs to add more files to `map_maker`, for much better readability and much better code organisation.

Tested, works.

### Testing

- Run CTF. Observe that everything related to maps work correctly.
- Run MTG with the `ctf_map` modpack enabled. Observe that the `*** CTF_MAP IS IN MAP MAKER MODE ***` is still displayed properly, and that you're able to access the map maker GUI.


ClobberXD:
Marking this PR as high priority, as it blocks any further development on that front.

Panquesito7:
The `Rebase needed` label should be added. 😉

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "mods/crafting"]
 	path = mods/crafting
 	url = https://github.com/rubenwardy/crafting
-[submodule "mods/ctf/ctf_map/maps"]
-	path = mods/ctf/ctf_map/maps
-	url = https://github.com/MT-CTF/maps.git
+[submodule "mods/ctf/ctf_map/ctf_map/maps"]
+	path = mods/ctf/ctf_map/ctf_map/maps
+	url = https://github.com/MT-CTF/maps
```
Panquesito7: Why remove `.git`?

mods/ctf/ctf_map/ctf_map/mod.conf
```diff
@@ -0,0 +1,3 @@
+name = ctf_map
+depends = default
+optional_depends = ctf, ctf_match, ctf_stats, ctf_treasure, stairs, wool, irc
```
Panquesito7: Add a description below.

mods/ctf/ctf_map/map_maker/README.md
```diff
@@ -1,16 +1,10 @@
...
-
-- Indestructible nodes adapted from various mods in `minetest_game`.
+# CTF Map - Map maker
 
 ## Creating a new map
 
 ### 1. Dependencies
 
-* Minetest 0.4.16 or later.
+* Minetest 5.0.0 or later.
```
Panquesito7: And/or later.

mods/ctf/ctf_map/map_maker/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = map_maker
+depends = ctf_map, worldedit
```
Panquesito7: Add a description below.

mods/ctf/ctf_map/modpack.conf
```diff
@@ -0,0 +1 @@
+name = ctf_map
```
Panquesito7: Add a description below.

.gitmodules
ClobberXD: I realised that the entries automatically created by `git submodule add <url> [path]`, do not have the `.git` extension at the end of the `url`. So I removed it for consistency.

mods/ctf/ctf_map/ctf_map/mod.conf
ClobberXD: Not necessary, but I'll try

mods/ctf/ctf_map/map_maker/mod.conf
ClobberXD: Same as above

mods/ctf/ctf_map/map_maker/README.md
ClobberXD: And/or doesn't make much sense. Why does a user need to have 5.0.0 **and** later to make a CTF map?

Although if we're talking about compat. "and" would make perfect sense - i.e. `ctf_map` is compatible with Minetest versions 5.0.0 and later. These are essentially just different ways of expressing the same thing.

mods/ctf/ctf_map/modpack.conf
ClobberXD: -_-

mods/ctf/ctf_map/map_maker/README.md
Panquesito7: IMHO, it's better with `and`.

mods/ctf/ctf_map/modpack.conf
Panquesito7: 🤣

ClobberXD:
Fixed merge conflict. While this PR is ready for review, it shouldn't be merged until #471 and #435 are merged.

ClobberXD:
Rebased. Adding to 2.0.0 milestone, as we have a lot of time...

ClobberXD:
Rebased.

#435 has been merged, this PR is ready for review.

ClobberXD:
Simply even updating the maps submodule creates a merge conflict... If someone wants to review, please let me know, and only then will I rebase.

mods/ctf/ctf_map/map_maker/README.md
```diff
@@ -1,16 +1,10 @@
...
 * Mods
   * ctf_map (by copying the folder from this game to `minetest/mods`)
```
Thomas--S: According to your first post, not the mod `ctf_map` is needed, but the modpack. The wording should be adjusted accordingly.

ClobberXD: Done

ClobberXD:
Rebased.

ClobberXD:
`ctf_map` the mod, has been renamed to `ctf_map_core` to prevent potential confusion from arising. Documentation has been updated, but I've intentionally left variable names and modname prefix in items' names untouched. This is to prevent confusion, and to retain compat - renaming all items from `ctf_map:item` to `ctf_map_core:item` will create a mass breakage of ind. nodes in existing maps. Btw, I've tested after the rename, and everything still works.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "mods/crafting"]
 	path = mods/crafting
 	url = https://github.com/rubenwardy/crafting
-[submodule "mods/ctf/ctf_map/maps"]
-	path = mods/ctf/ctf_map/maps
-	url = https://github.com/MT-CTF/maps.git
+[submodule "mods/ctf/ctf_map/ctf_map/maps"]
```
Thomas--S: Is there a `_core` missing?

mods/ctf/ctf_map/map_maker/README.md
```diff
@@ -94,16 +86,4 @@ An example `treasures` value that registers steel pick, shotgun, and grenade:
...
-- `ctf_map` provides indestructible nodes for most nodes from default, and all nodes from
-stairs.
-
-- All indestructible nodes have the same item name with the mod prefix being `ctf_map:`
-instead of their original prefixes (e.g. `default:stone` -> `ctf_map:stone` and
-`stairs:stair_stone` -> `ctf_map:stair_stone`) with the exception of wool, whose
-indestructible nodes have slightly different names from the original node names -
-`ctf_map:wool_<color>`. This is because the original nomenclature becomes meaningless
-if the modname prefix is changed.
+(See [here](../../../../other/treasurer/README.md) to understand the magic numbers)
```
Thomas--S: I think there's one `../` too much

.gitmodules
ClobberXD: Is there? :D

mods/ctf/ctf_map/map_maker/README.md
ClobberXD: Indeed, removed.

Thomas--S:
`update.sh` and `setup_maps.sh` need to be updated, too.

ClobberXD:
Ah yes, of course. Will do in a day.

ClobberXD:
Updated both of the shell scripts.

mods/ctf/ctf_map/ctf_map_core/mod.conf
```diff
@@ -0,0 +1,3 @@
+name = ctf_map
+depends = default
+optional_depends = ctf, ctf_match, ctf_stats, ctf_treasure, stairs, wool, irc
```
Panquesito7: Description below?

ClobberXD: Suggestions for the description are welcome :)

mods/ctf/ctf_map/ctf_map_core/mod.conf
```diff
@@ -0,0 +1,3 @@
+name = ctf_map
```
Panquesito7:
```suggestion
name = ctf_map_core
```

ClobberXD: Fixed, thanks.

mods/ctf/ctf_map/ctf_map_core/mod.conf
Panquesito7: I have no idea.
Will think on it... 🙂

Thomas--S:
It is not possible to update the server to this new structure via update.sh (because of submodule stuff). It'd be very good if you could fix this.

ClobberXD:
A manual `git pull` before running `update.sh` should suffice. But I agree that a better solution is required in the long-term...

Thomas--S:
> A manual git pull before running update.sh should suffice. But I agree that a better solution is required in the long-term...

I ran update.sh twice, which should have the same effect. The problem is, that the new maps directory doesn't exist and the contained submodule isn't initialized.

Having to do this by hand would be very bad design, so I think update.sh should be changed accordingly, so everything will work automatically, at least if you run update.sh twice or pull before running it once.

Thomas--S:
I re-reviewed this PR. There are several points which need to be addressed:

1) This PR proposes a structure, where a modpack (ctf_map) is inside a modpack (ctf). Although this seems to work, I strongly doubt whether this is officially supported.
2) The submodule configuration is messed up. See https://github.com/Thomas--S/capturetheflag/commit/57c931dbe4ec59b662f4a9f427eece729f5b79eb for a (tested) solution.
3) It isn't possible to start a game using the code from this PR because there are errors related to the renaming of `ctf_map` to `ctf_map_core`. See https://github.com/Thomas--S/capturetheflag/commit/2ed51cc031c6e54911aa28e0406c50c54144d3eb for a semi-tested and probably only partial solution.
4) You have to make sure that the old maps and textures directories are removed when from the ctf_map directory in order to make the game work.

ClobberXD:
> This PR proposes a structure, where a modpack (ctf_map) is inside a modpack (ctf). Although this seems to work, I strongly doubt whether this is officially supported.

Valid concern, and I see why you think nested modpacks aren't officially supported. Will clarify with the devs regarding this matter.

> `git submodule update --init --recursive`

I replaced the `cd dir; git pull` with `git submodule update --init --recursive` in the shooter submodule PR. I haven't used it anywhere else for the fear of merge conflicts, but I bet the PR already has a bunch of merge conflicts, so I'll fix the mess in this PR instead.

> It isn't possible to start a game using the code from this PR because there are errors related to the renaming of ctf_map to ctf_map_core.

This is strange. I specifically tested that CTF works after I've renamed the mod, and MT didn't crash. Anyway, I think prefixing item names `:` is the right thing to do.

ClobberXD:
rubenwardy [says](http://irc.minetest.net/minetest-hub/2019-11-23#i_5614341) that nested modpacks are officially supported, although the "Configure" dialog in the main menu doesn't yet allow individual modpacks to be toggled.

Thomas--S:
> rubenwardy says that nested modpacks are officially supported, although the "Configure" dialog in the main menu doesn't yet allow individual modpacks to be toggled.

That's very good indeed. I will then review this PR as soon as necessary and possible.

However, I'd strongly prefer it if rubenwardy's statement was added to Minetest's lua_api.txt. Probably we should open a PR for this?

ClobberXD:
> However, I'd strongly prefer it if rubenwardy's statement was added to Minetest's lua_api.txt

- Does that affect your approval of this PR in any way? ;)
- I'd wait for a bit, maybe even until after GUI support for nested modpacks has come by.

Thomas--S:
> Does that affect your approval of this PR in any way? ;)
   I'd wait for a bit, maybe even until after GUI support for nested modpacks has come by.

I'd be more comfortable with approving then when it is documented that nested modpacks are supported, but I guess that rubenwardy's statement is nearly as good as an official documentation; so missing documentation won't stop me from approving this.

ClobberXD:
Rebased

Thomas--S:
Rebase needed.

ClobberXD:
Waaah, I just rebased this morning! :_(

Thomas--S:
The actual git submodule for maps is still missing. Please merge ClobberXD/capturetheflag#2 to fix this.

mods/ctf/ctf_map/map_maker/mod.conf
Thomas--S:
```suggestion
depends = ctf_map_core, worldedit
```

PR #478, opened 2019-10-16
Add support for optionally setting time per-map

ClobberXD:
This PR adds support for optionally setting start time and time speed per-map, using the `start_time` and `time_speed` map meta fields respectively.

This feature is required by #471. Tested, works.

### Testing

- Add the following to the conf. file of a surface map, and observe that the sun goes to the horizon, and stays there forever:

```
start_time = 0.21
time_speed = 0
```

- Add the following to the conf. file of another surface map, and observe that time traverses 10x faster:

```
time_speed = 10
```

- Also, finish/skip the modified maps, and observe that time works as expected on maps that don't specify these fields.


ClobberXD:
I've indeed tested this PR quite extensively. Thanks for the approval.

PR #481, opened 2019-10-26
Add damage cobble to treasure chests

MinetestSam:
Damage cobble is the only trap which doesn't allow griefing and so should be added to treasure chests


JostP:
Now when here is a PR that modify treasure chest content, maybe you could also remove stone swords, because nobody takes them from chests and they are just there with no reason? (and maybe stone shovel too).

Although stone swords are removed on maps where you get them as initial stuff, they still appear in chests on cave maps and there where you get steel sword as initial stuff.

ClobberXD:
> Now when here is a PR that modify treasure chest content, maybe you could also remove stone swords, because nobody takes them from chests and they are just there with no reason?

PRs should not be combined, just because they make changes to the same region of code.

PR #482, opened 2019-10-28
[Trivial] Fixing documentation for reinforced cobble

MinetestSam:


mods/ctf/ctf_crafting/init.lua
```diff
@@ -64,7 +64,7 @@ crafting.register_recipe({
 	level  = 1,
 })
 
--- Wooden plank x4
+-- Reinforced Cobblestone
```
ClobberXD: What on earth? o_O

PR #487, opened 2019-11-02
[Trivial] Fix chest spawning in sky bug

MinetestSam:


ClobberXD:
`findGroundLevel` does the following:

- It scans a slice of nodes from `y_max` to `y_min`.
  - If node is air, water, or lava, `top` is set to the current `y` value.
- It scans a slice of nodes from `top` to `y_min`.
  - If node is not air, water, or lava, we have the top-most/surface level ground node.

When this algo runs at the barrier:

- The very first node at `y_max` isn't any of the void nodes - `top` is set to `y_max`, and the loop breaks immediately.
- Then while checking for the first ground node, the node at `top` is considered to be at ground level.

Taking the algo into consideration, adding `ctf_map:ind_glass_red` to the list of void nodes isn't really a solution, but a hack IMO. A better solution would be to not run at all, if the slice intersects the barrier.

ClobberXD:
Hmm, it turns out that rubenwardy indeed made chests spawn up there on purpose. In a way, at least. :P

The following diff removes a couple of lines of barrier-specific code added by rubenwardy after he initially added the `tsm_chests` mod:

```diff
diff --git a/mods/other/tsm_chests/init.lua b/mods/other/tsm_chests/init.lua
index 2ce92d3..5d0ce88 100755
--- a/mods/other/tsm_chests/init.lua
+++ b/mods/other/tsm_chests/init.lua
@@ -198,11 +198,6 @@ function place_chests(minp, maxp, seed, number_chests)
 		if ground ~= nil then
 			local chest_pos = {x = pos.x, y = ground + 1, z = pos.z}
 
-			-- Don't spawn at barrier
-			if chest_pos.z == 0 then
-				chest_pos.z = -1
-			end
-
 			-- chest node name (before it becomes a chest)
 			local nn = minetest.get_node(chest_pos).name
 			if nn == "air" or nn == "default:water_source" then
```

As can be seen right below the removed chunk, the chest is placed only if the selected pos contains air or water. If the node at pos is a barrier node, the chest isn't placed at all. I've also tested this, and I can confirm that this diff indeed fixes the chests spawning in the sky problem.

@MinetestSam, feel free to use the diff in any way you deem fit.

ClobberXD:
Replaced by #492

PR #489, opened 2019-11-08
Invoke ctf_match.next immediately on startup

ClobberXD:
It's wrapped in a `minetest.after(0)` so that it's called only after the server env. is fully loaded. `minetest.register_on_mods_loaded` doesn't work here, which I presume is called before the env. is fully ready.

Earlier, there was a waiting time of 5 seconds, when the player would be floating in the middle of nowhere, and the teams wouldn't be created. There'd also be this annoying error - "No teams to allocate <name> to!".

Tested, works.

### DIFF
--- a/mods/ctf/ctf_match/init.lua
+++ b/mods/ctf/ctf_match/init.lua
@@ -37,6 +37,4 @@ if minetest.global_exists("irc") then
 	end)
 end
 
-minetest.after(5, function()
-	ctf_match.next()
-end)
+minetest.after(0, ctf_match.next)


Thomas--S:
This needs very careful testing to check if a race condition is introduced by this.

ClobberXD:
When using `minetest.register_on_mods_loaded`, the map schem wasn't loading, and I was hanging next to the flag in thin air. But when using `minetest.after`, everything works fine. But of course, more testing is always better.

Thomas--S:
> When using minetest.register_on_mods_loaded, the map schem wasn't loading, and I was hanging next to the flag in thin air.

I think we shouldn't merge this PR until we know ...
 1) ... why the behaviour described in the quote happens (i.e. is this a technically defined order or is this a race condition)
 2) ... why rubenwardy (?) originally chose 5 seconds.

ClobberXD:
> ... why the behaviour described in the quote happens (i.e. is this a technically defined order or is this a race condition)

Perhaps that's just one of the numerous quirks of Minetest? Maybe `on_mods_loaded` runs right after the last mod is loaded, and `minetest.after(0, ...)` one step later?

ClobberXD:
:notes: Bumpity-bump;
My name is not Trump. :musical_note:

ClobberXD:
@Thomas--S Hi

Thomas--S:
I'm very hesitant regarding this PR because I fear this may cause unintended side-effects (race conditions).

For example, I think this line might be problematic in combination with this PR: https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_stats/init.lua#L279

ClobberXD:
> I'm very hesitant regarding this PR because I fear this may cause unintended side-effects (race conditions)

It doesn't cause any side-effects, going by my bunch of test runs. Due to the way `minetest.after` works, a `minetest.after(0, ...)` will execute only on the first server tick, which takes place after the server env. has initialised. OTOH, `on_mods_loaded` is on the C++ side, and gets called relatively earlier, right after the mods have been loaded, and before the server env. is ready (AFAIK).

> I think this line might be problematic in combination with this PR: https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_stats/init.lua#L279

That line *registers* a callback. The `minetest.after(0, ...)` there, is to ensure that `ctf_map` has loaded by then, because directly setting `ctf_map` as a dependency (even as an optional one) will create a cyclic dependency between `ctf_map` and `ctf_stats`. In hindsight, I should've used an `on_mods_loaded` callback instead, lol.

Thomas--S:
> It doesn't cause any side-effects, going by my bunch of test runs.

In such a scenario, one mustn't rely on testing results because they always happen under a certain environment. They can go well "by luck", but a different environment or further code changes may change the behaviour.

To examine if this change is unproblematic, one would have to examine all the implicit and explicit dependencies of the code and how and when the related code is executed.

I'd consider this PR rather low priority, given that the improvement is rather small and the review efforts must be huge.

ClobberXD:
@rubenwardy Your inputs would be appreciated.

PR #495, opened 2019-11-09
ctf_stats: Add map name to match summary

ClobberXD:
screenshot_20191109_173245

Opinions and thoughts regarding the placement of the map name is appreciated. Obviously, what looks good on my screen could look completely borked on another screen. That's the inherent limitation of formspecs; hopefully `real_coordinates` would help improve consistency of formspecs across different screens.

Tested; works. Pretty trivial to review and test.

### DIFF
--- a/mods/ctf/ctf_stats/init.lua
+++ b/mods/ctf/ctf_stats/init.lua
@@ -217,6 +217,14 @@ ctf_match.register_on_new_match(function()
 	_needs_save = true
 end)
 
+-- ctf_map can't be added as a dependency, as that'd result
+-- in cyclic dependencies between ctf_map and ctf_stats
+minetest.after(0, function()
+	ctf_map.register_on_map_loaded(function(map)
+		ctf_stats.current.map = map.name
+	end)
+end)
+
 ctf_flag.register_on_pick_up(function(name, flag)
 	local main, match = ctf_stats.player(name)
 	if main and match then
--- a/mods/ctf/ctf_map/schem_map.lua
+++ b/mods/ctf/ctf_map/schem_map.lua
@@ -262,6 +262,11 @@ local function place_map(map)
 	end, nil)
 end
 
+ctf_map.registered_on_map_loaded = {}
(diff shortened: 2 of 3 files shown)


ClobberXD:
Added `Map: ` before the map name

mods/ctf/ctf_stats/init.lua
```diff
@@ -215,6 +215,12 @@ ctf_match.register_on_new_match(function()
 	winner_player = "-"
 	ctf_stats.start = os.time()
 	_needs_save = true
+
+	-- The delay is required to prevent a race condition, because ctf_map
+	-- uses the same on_new_match callback for map initialization
+	minetest.after(1, function()
```
Thomas--S: To be honest I strongly oppose such a time-based approach.

Strictly speaking, this still adds kind of a race condition because it cannot be guaranteed that the other function takes less than a second.

The clean way to to this is to allow registering callbacks in the ctf_map mod that are run when the new map name is set there.

ClobberXD: 1s is far too much to create a race condition, but I can see where you're going with this.

> The clean way to to this is to allow registering callbacks in the ctf_map mod that are run when the new map name is set there.

Fair point. I'll do just that.

ClobberXD:
I've added a new callback `ctf_map.register_on_map_loaded`, that'll be executed when a map has fully finished loading. I've removed the `minetest.after(1, ...)`, and replaced it with the callback. Note that directly adding `ctf_map` as a dependency of `ctf_stats` will create a cyclic dependency, so I resorted to using `minetest.after(0, ctf_map.register_on_map_loaded)` instead. PR tested, works perfectly.

Note that there's a debug message for easier testing, and should be removed before merge.

mods/ctf/ctf_stats/init.lua
ClobberXD: Done

ClobberXD:
Debug message removed. Feel free to merge this if you like.

PR #496, opened 2019-11-10
ctf_match: Add automatic skip votes

ClobberXD:
### Issue at hand

At times, matches tend to get excessively long, at times even touching 4 hours. On the other hand, players have the ability to manually skip matches and they (ab)use this ability to skip matches they don't like, or because other players ask them to. I don't blame this squarely on the players though, because it's also our fault for not setting clear rules and guidelines. The solution - auto-trigger skip votes.

### Automatic skip-votes

The first skip vote of a match is triggered after a specified delay (defaults to 90m or 1h30m) calculated from the end of buildtime, and subsequent votes are triggered after a specified interval (defaults to 15m).

Automatic skip votes would run if and only if they were triggered in the same match. i.e. Skip votes would automatically be cancelled once the ongoing match ends, either by winning or by skip.

This PR also announces the vote in chat using `minetest.chat_send_all` instead of on IRC. In full colour, of course:

screenshot_20191110_085105

### Testing

While testing, make sure to set the delay and interval to very low values (e.g. 20s and 10s respectively), to make your life easier.

- Once the match starts, skip buildtime, and the first auto-vote will be triggered after 20s.
- Vote `/no`, and cancel the vote, so that the subsequent auto-votes can be tested.
- Every 10s after the first vote gets triggered, the subsequent votes are triggered.
- Test that the auto-vote stops at the end of the match, either by capturing, or by voting `/yes` in one of the follow-up votes.
- Use chat-messages/`print` statements if required, to ensure that everything indeed works properly.

Tested, also with the help of chat messages. Everything works as intended.

### Caveat

`/ctf_next` (a command that I overuse/abuse in singleplayer while testing) neither triggers the `on_winner` or the `on_skip_match` callbacks, so the auto-vote isn't disabled. Potential solutions to overcome this caveat are:

- Track same match using "match IDs" instead of a single boolean.
  - This is a good idea, **if and only if** we actually make use of the match IDs for other purposes. Unique match IDs per-session, just to regulate auto-vote is overkill.
- Make `same_match` global, and manually set it to `false` in `/ctf_next`
  - Hacky; pls no...
- Use `on_new_match` callback, instead of `on_winner` and `on_skip_map`.
  - The most straight-forward solution, and this is obviously what I tried at first. But setting `same_match = false` at the start of a new match, means that the first match after launching CTF will start with `same_match == false`, and will end up aborting after setting the var to `true`. The result is that the auto-vote won't run throughout the first match at all.
- Consider `/ctf_next` as a match skip, and trigger `on_skip_map` callbacks when invoked.
  - The next most straight-forward solution, and this is also what I personally prefer. But the implications of taking `/ctf_next` to be a match skip has to be carefully considered.

****

Discussion regarding the feature, the caveat, and its solution, are most welcome.

### DIFF
--- a/mods/ctf/ctf_match/vote.lua
+++ b/mods/ctf/ctf_match/vote.lua
@@ -6,11 +6,10 @@ function ctf_match.register_on_skip_map(func)
 	table.insert(ctf_match.registered_on_skip_map, func)
 end
 
-function ctf_match.vote_next(name, params)
-	if minetest.global_exists("irc") then
-		local tname = ctf.player(name).team or "none"
-		irc:say("Vote started by " .. name .. " (team " .. tname .. ")")
-	end
+function ctf_match.vote_next(name)
+	local tcolor = ctf_colors.get_color(ctf.player(name)).css or "#FFFFFFFF"
+	minetest.chat_send_all(minetest.colorize("#FFAA11", "Vote started by ") ..
+		minetest.colorize(tcolor, name))
 
 	return vote.new_vote(name, {
 		description = "Skip to next match",
@@ -50,13 +49,31 @@ minetest.register_chatcommand("vote", {
 	func = ctf_match.vote_next
 })
 
-minetest.register_on_chat_message(function(name, msg)
-	if msg == "/vote_next" and minetest.check_player_privs(name,
-			{interact=true, vote_starter=true}) then
-		local _, vmsg = ctf_match.vote_next(name)
-		if vmsg then
-			minetest.chat_send_player(name, vmsg)
-		end
-		return true
+local same_match = true
+local function auto_vote(interval)
+	-- If not same match, set it to true and return
+	if not same_match then
+		same_match = true
+		return
 	end
+
+	-- Start vote
+	ctf_match.vote_next("[CTF automatic skip-vote]")
+
+	-- Recursively call the same function after `interval` seconds
+	minetest.after(interval, auto_vote, interval)
+end
+
+local function disable_auto_vote()
+	-- At the end of a match, set same_match to false to disable auto_vote
+	same_match = false
+end
+ctf_match.register_on_winner(disable_auto_vote)
+ctf_match.register_on_skip_map(disable_auto_vote)
+
+-- Automatically start a skip vote after 90m, and subsequent votes every 15m
+ctf_match.register_on_build_time_end(function()
+	local delay    = tonumber(minetest.settings:get("ctf_match.auto_skip_delay"))    or 30 * 120
+	local interval = tonumber(minetest.settings:get("ctf_match.auto_skip_interval")) or 30 *  15
+	minetest.after(delay, auto_vote, interval)
 end)


KaylebJay:
Just my opinion here ... I agree this PR has plenty of common sense and it would really help to make those long matches end sooner. But things need to be taken into account here ... First, if any match is lasting that long, it means pro's are involved and plenty of pros have vote_starter. So long matches are skipped when people want them to ... already. Second, what happens if some person makes a bunch of alts join, just to vote /yes to skip the match, when everyone else wants to play? Although this would be an easy fix - one vote per IP - it also complicates where there are families playing on the same IP ... and TBH, why are we even trying to change this anyways? This will only make the players that love long matches mad, IMO. I'd seriously consider upping the time span ... 2 hours or 3 hours makes more sense, as the map does need to skip at some point, but 90 minutes is hardly enough. Again just my thoughts :D

KaylebJay:
"On the other hand, players have the ability to manually skip matches and they (ab)use this ability to skip matches they don't like, or because other players ask them to."

If people have vote_starter this is going to be abused, no matter what... what's your plan here?

ClobberXD:
> "On the other hand, players have the ability to manually skip matches and they (ab)use this ability to skip matches they don't like, or because other players ask them to."
> 
> If people have vote_starter this is going to be abused, no matter what... what's your plan here?

Their priv would be revoked, ofc. Only staff, and maybe the most trustworthy players would have the priv.

KaylebJay:
I can imagine some players being mad about that... but I agree with this completely. 
If the above is implemented then I see much sense in the whole idea. It will really come down to 'how much time' however, I think. Thanks for info!

ClobberXD:
@KaylebJay The objective is to not keep wallowing in the same match for too long, because players would soon get bored, and if they're unable to capture the flag to end the match, they'll leave. In fact, that's how most 2+ hour matches end - players start getting bored of the apparent stalemate and leave, tipping the team balance to one side, effectively expediting the flag capture. Also, we have a ton of maps now, and it's a good thing to keep the length of matches in check for the sake of variety in gameplay.

LoneWolfHT:
Before this is merged we need to fix voting first. 50% of the votes is a bit too little for something as big as skipping a match. Without putting much thought into it I suggest doing 75%.
Also foreign players currently can't really vote either. We need to work on translations

ClobberXD:
I agree that the percentage needed has to be increased. 50% is indeed too shallow. But I'm thinking of around 60-65%; 75% is probably a bit too steep.

KaylebJay:
Oh yes. I agree... and 60-65% sounds quite right.

KaylebJay:
@ClobberXD Wasn't there a thing like that about a year ago? There was like a 30m or 1hr limit on matches, whichever team had the more points won. I think?

ClobberXD:
I haven't encountered that system ever. But there's a chance it could've existed at around 2015-2017.

LoneWolfHT:
75%:
* [4 votes] -> Requires 3 yes votes
* [6 votes] -> Requires 4 yes votes (rounded down from 4.5)
* [10 votes] -> Requires 7 yes votes (rounded down from 7.5)
* [20 votes] -> Requires 15 yes votes

LoneWolfHT:
We usually get around 6 votes in 20-player matches AFAIK
Might be nice if we could log votes, including yes/no votes, players online, how many voted etc. Would help with deciding the best value

LoneWolfHT:
Updated list

MinetestSam:
I am against the 75% limit and for me the fine line is around 60%.  I am fine with the matches being auto skipped but the time limit does need to be increased and the players need to be given warnings too.

Note: I expect a lot of vote_starter privs being removed if this goes through which would be make a lot of people mad.

ClobberXD:
65% seems to be the ideal ratio to me. I'll make the time 2h by default.

ClobberXD:
I'd like to see this in 2.0.0

LoneWolfHT:
Replaced by https://github.com/MT-CTF/capturetheflag/pull/675 which fixes the skipping issue and doesn't start votes if a player has the flag

PR #498, opened 2019-11-10
ctf_map: Re-implement random map selection in a better way

### LINKED ISSUE #497: Make map selection random again
Two things:

- Number of maps have increased tenfold.
- Maps have been de-coupled from this repository, so we can't hardcode stuff depending on the number of maps.

My proposal: Implement both cyclic and random map selection, but employ the appropriate technique based on the number of maps. I'd say random selection if `#maps >= 10`.

ClobberXD:
This PR re-implements random map selection, but in a better way. Randomised selection order is employed only if the number of maps >= 10, otherwise the randomness won't be random enough.

### How the randomisation works (only applicable if #maps >= 10)

- Once `load_maps` executes, the map **indices** (not the maps themselves) are shuffled (using Fisher-Yates shuffling algorithm) and stored in a separate table.
- The map selection (now moved to a separate local function) iterates through the table of shuffled indices each time it's called.
- Once the end of the shuffled indices table has been reached, the indices are shuffled again, and are iterated from the beginning once again.

Closes #497. Tested; works perfectly.

Note: I haven't tested this PR after rebasing on top of #435.

### DIFF
--- a/mods/ctf/ctf_map/ctf_map_core/schem_map.lua
+++ b/mods/ctf/ctf_map/ctf_map_core/schem_map.lua
@@ -214,7 +214,7 @@ math.randomseed(os.time())
 
 -- Fisher-Yates shuffling algorithm, used for shuffling map selection order
 -- Adapted from snippet provided in https://stackoverflow.com/a/35574006
-local function shuffle_maps()
+local function shuffle_maps(idx_to_avoid)
 	-- Reset shuffled_idx
 	shuffled_idx = 1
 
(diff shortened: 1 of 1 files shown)


ClobberXD:
Contains debugging messages that need to be removed before merge.

ClobberXD:
Rebased. Fixed trivial merge conflict.

Thomas--S:
Do I understand this correctly that the order only gets shuffled when the maps are loaded, i.e. usually only on server restart?

BTW, rebase is needed.

ClobberXD:
> Do I understand this correctly that the order only gets shuffled when the maps are loaded, i.e. usually only on server restart?

Map-selection order is shuffled:
- when `load_maps()` is called, i.e. on startup and on `/maps_reload`.
- when the current cycle is over.

Let's assume there are 10 maps. The shuffled map-selection order is first created during `load_maps()`. Once all 10 maps have been selected (acc. to the shuffled order), the order is shuffled again. Since the map-selection order won't be stuck to the same shuffled order from the first (and only) shuffle at startup, it would be truly randomised.

Thomas--S:
See ClobberXD#3

Looks good otherwise.

Thomas--S:
See ClobberXD#4

PR #499, opened 2019-11-10
Remove redundant server join message; improve team join message

ClobberXD:
- Server join messages have been merged into team join messages. i.e. the first team join message after a player has connected to the server will indicate the same.
- Team join message now colorizes player name and team name, and is now prefixed by three stars.

screenshot_20191125_005513

Tested; works.

### DIFF
--- a/mods/ctf/ctf/teams.lua
+++ b/mods/ctf/ctf/teams.lua
@@ -145,6 +145,7 @@ function ctf.register_on_join_team(func)
 end
 
 ctf.player_last_team = {}
+local first_alloc = {}
 
 -- Player joins team
 -- Called by /join, /team join or auto allocate.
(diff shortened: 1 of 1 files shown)


ClobberXD:
:notes: Bumpity-bump
in my throat is a lump :musical_note:

Thomas--S:
According to `lua_api.txt` it is not allowed to call `minetest.register_on_joinplayer()` in `minetest.after()`:

```
Global callback registration functions
--------------------------------------

Call these functions only at load time!

[...]

* `minetest.register_on_joinplayer(function(ObjectRef))`
    * Called when a player joins the game
```

Even if this works at the moment, I'm strongly opposed to merging this because we have no guarantee that it will continue to work in future releases.

ClobberXD:
The reason I used `minetest.after(0, ...)` is because `ctf.setting` isn't accessible before init. However, I agree that placing `on_joinplayer` within the `minetest.after` is a bad idea. I did that to only prevent checking for the same setting twice.

This has been fixed now by reverting all changes to the `on_joinplayer` callback, and separately disabling server join messages in a `minetest.after(0, ...)` after checking for the setting.

Thomas--S:
I'd prefer it if the three stars are only shown when the player initially joins, not on team change or match change. I think this will avoid some confusion. Otherwise this looks good to me.

ClobberXD:
> I'd prefer it if the three stars are only shown when the player initially joins, not on team change or match change. I think this will avoid some confusion.

Are you sure? This *can* be implemented, but we'd lose out on consistency, and the team-join message would sometimes have `*** ` and sometimes wouldn't. It'd probably be just as confusing, if you ask me...

Thomas--S:
> Are you sure? This can be implemented, but we'd lose out on consistency, and the team-join message would sometimes have *** and sometimes wouldn't. It'd probably be just as confusing, if you ask me...

This is a valid concern.

However, I think we should provide some way to distinguish joining the game from changing teams.

An alternative solution might be to always print the stars but use two different messages:

- *** root joined the game and got allocated to team red.
- *** root joined team blue.

ClobberXD:
The initial join message is now different from the usual team-join message. Feedback on the new join message(s) appreciated.

Thomas--S:
I always get the first alloc message now. I think you probably forgot to set `first_alloc[name]` to false after the initial message was print.

Furthermore, I'd slightly prefer it if the team name was also mentioned in the first alloc message.

Apart from that, the PR looks very good, as far as I can see.

ClobberXD:
Ah yes, I hadn't tested the latest changes.

ClobberXD:
Fixed; done; over.

First post updated; PR's ready for review.

PR #500, opened 2019-11-10
ctf_match: Increase skip vote's required majority ratio to 0.6

ClobberXD:
See also - #496

### DIFF
--- a/mods/ctf/ctf_match/vote.lua
+++ b/mods/ctf/ctf_match/vote.lua
@@ -16,7 +16,7 @@ function ctf_match.vote_next(name, params)
 		description = "Skip to next match",
 		help = "/yes,  /no  or  /abstain",
 		duration = 60,
-		perc_needed = 0.5,
+		perc_needed = 0.6,
 		unanimous = 5,
 
 		on_result = function(self, result, results)


LoneWolfHT:
Do you round up when the votes needed are fractional?

Thomas--S:
65% seems a bit too much for me, it is nearly a two-third majority.

Something like 55% to 60% seems far more appropriate to me.

LoneWolfHT:
A 2/3 majority with non-english players understanding how to vote works for me

ClobberXD:
With nubs voting `/yes` without even taking their eyes off the enemy, I'd say 0.65 is a very fair ratio. In the long run though, I'd prefer a weightage-based system, where the more "pro" a player is, the more weightage their votes would carry.

Thomas--S:
If every player voted conciously, I think a 50% threshold would be appropriate.

To take the unexperienced players into account, I think 55-60% percent can be appropriate, but a two third majority seems a bit to much, especially given the fact that votes are only started under circumstances where a skip is useful.

Calinou:
A 60% majority sounds good to me. In my experience, many players will never vote no matter what you do, so this should be taken into account.

LoneWolfHT:
I think instead of not voting most players just vote 'yes'

ClobberXD:
> A 60% majority sounds good to me. In my experience, many players will never vote no matter what you do, so this should be taken into account.

From what I've seen, it seems to me that the exact opposite is true.

MinetestSam:
I agree with Thomas-S and Calinou, 65% is a bit too much and 60% should work fine

ClobberXD:
From https://github.com/MT-CTF/capturetheflag/pull/496#issuecomment-552165944:

> ### 65%:
> 
>     * [4 votes] -> Requires 2.6 yes votes
>     * [6 votes] -> Requires 3.9 yes votes
>     * [10 votes] -> Requires 6.5 yes votes
>     * [16 votes] -> Requires 10.4 yes votes
>     * [26 votes] -> Requires 16.9 yes votes

4/6 votes, 7/10 votes, 11/16 votes; these sound very sensible to me. If the vote is about skipping the match, a substantial portion of the online player-base should agree, that's what I feel. Also, this PR goes hand-in-hand with #496, which means that votes aren't started with the clear intention to skip. This means that the skip vote should all-the-more require the aforementioned approval from a substantial percentage of online players.

Some people have also highlighted the fact that players can use alts to cast multiple votes, and I agree that it's a grave concern. But this issue has been existing right from the time voting was introduced. Citing that to claim that this bump in percentage is useless doesn't make much sense to me.

Thomas--S:
> Some people have also highlighted the fact that players can use alts to cast multiple votes, and I agree that it's a grave concern. But this issue has been existing right from the time voting was introduced. Citing that to claim that this bump in percentage is useless doesn't make much sense to me.

It makes sense insofar as this makes blocking votes increasingly easy with a higher threshold.

When the threshold is rather low, it is rather difficult to block otherwise clear votes with this method.
With a high threshold, however, it is very easy to block even rather clear votes. This should be considered when thinking about a threshold.



Apart from that, I think it might be a good compromise to use 60% (personally I'd prefer 55%), and then see how this change effects the voting system. If we then see the need to increase the threshold even more, we can think about raising it again at a later point in time.

ClobberXD:
> Apart from that, I think it might be a good compromise to use 60% (personally I'd prefer 55%), and then see how this change effects the voting system. If we then see the need to increase the threshold even more, we can think about raising it again at a later point in time.

Sounds good; I'm fine with trial and error too. So 60% it is, for now.

*hides the `Controversial` label under his bed* :D

LoneWolfHT:
When the vote starts save a list of all online players and only allow players in that list to vote?

MinetestSam:
But then that'll also prevent new players who have just joined from voting which would go against the idea of everyone getting to vote.

LoneWolfHT:
> But then that'll also prevent new players who have just joined from voting which would go against the idea of everyone getting to vote.

No new players that share an IP with someone online then?

ClobberXD:
Reduced required majority ratio to 0.6, and rebased.

ClobberXD:
> I didn't test this ...

You did not WHAT?! You're FIRED! :angry:

PR #503, opened 2019-11-12
[Trivial] Increase damage of spikes to justify its cost.

MinetestSam:
I have not yet decided the specific damage value and I'm open to suggestions


JostP:
Maybe use damage 7 and not 10 so you will die in 3 hits. 10 is too much IMO, you know that spikes are used for griefing most times.

PR #505, opened 2019-11-13
Change vote_kick percentage

TSafa-23:
Change vote_kick percentage to pass the vote from 0.8 to 0.75 (for rounds where there are few players).

### DIFF
--- a/mods/other/vote/vote_kick.lua
+++ b/mods/other/vote/vote_kick.lua
@@ -54,7 +54,7 @@ minetest.register_chatcommand("vote_kick", {
 			help = "/yes,  /no  or  /abstain",
 			name = param,
 			duration = 60,
-			perc_needed = 0.8,
+			perc_needed = 0.75,
 
 			can_vote = function(self, pname)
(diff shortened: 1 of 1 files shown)


MinetestSam:
vote_kick was recently changed so that only players above 1000 score could vote, so I don't think this plays an important role and even if there are few players then probably less than half would be eligible to vote.

ClobberXD:
4/5 is the perfect ratio for kick votes, in my opinion - there needs to be a significant portion of players agreeing to the kick, because it's a major decision. I'm opposed to this, sorry :-1:

Thomas--S:
I had supported this PR before voting was restricted to players with score > 1000.

In the current situation, I think we can leave the 80% threshold. If we see concrete examples where this doesn't work, we can still consider fine-tuning.

What I would appreciate is the following: The player to be kicked shouldn't be able to vote.

ClobberXD:
> What I would appreciate is the following: The player to be kicked shouldn't be able to vote.

PRs welcome ;)

PR #508, opened 2019-11-15
Fix collision boxes of dead players not being restored on new match

### LINKED ISSUE #506: Collision boxes of dead players are not restored at the start of a new match
Collision box is not returned to default if the player never respawns when the flag is captured. This results in the player having no collision box on the next map. The player can get back to the normal collision box only by killing themselves.

**To Test**
1. Make a private server
2. Log in through two accounts.
3. Kill the other player and make sure to NOT respawn with the player that has been killed
4. Grab the flag and end the game.
5. In next match, you'll see that the player that was killed still has no collision box at all

ClobberXD:
The cause of this bug is that `on_respawnplayer` isn't being invoked when the player is "respawned" from code by setting their HP. A workaround to this is to manually restore the collisionboxes of all dead players at the start of a new match, which is exactly what this PR does.

Tested, works. Semi-trivial in terms of size of changes and review/testing required. Closes #506.

### DIFF
--- a/mods/pvp/collisionbox/init.lua
+++ b/mods/pvp/collisionbox/init.lua
@@ -2,14 +2,27 @@ local collision_box = {}
 
 minetest.register_on_dieplayer(function(player)
 	local name = player:get_player_name()
-	if not collision_box[name] then
-		collision_box[name] = player:get_properties().collisionbox
-	end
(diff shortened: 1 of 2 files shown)


ClobberXD:
Thanks for testing

PR #515, opened 2019-11-23
Maps Catalog: Fix crash when index is invalid

Thomas--S:


ClobberXD:
I'd prefer the check to be done in `/maps`, as the chat-command is currently the only way the `idx` could be `nil`, as far as I know. Even then, I haven't been able to reproduce this (repro steps would be great, if you managed to reproduce the crash, btw).

But if that's the case, the following might be a better place to check for `nil` idx: https://github.com/MT-CTF/capturetheflag/blob/f0c4e19b51f0e57ba1214cae5f25c2be719f03cc/mods/ctf/ctf_map/maps_catalog.lua#L162-L165

But if my guess is right and the a `nil` idx is only possible by passing an invalid param to the chat-command, then my suggestion is to change this part of the code to the following (untested):

``lua
if param then
	local idx = ctf_map.get_idx_and_map(param)
	if idx then
		indices[name] = idx
	else
		return false, "Invalid keyword. Map does not exist."
	end
end
```

Thomas--S:
Passing an invalid keyword doesn't trigger this problem. I wasn't able to reproduce it, but I guess it might be caused by sending invalid/modified formspec events.

After all, I think it is generally a good idea to make sure that `map` is really a map and not nil because this is where the problem happens.

`ctf_map.available_maps[idx]` returns either nil or a map, and the following code relies on a map being returned. So, we have to make sure that the execution of the function is aborted in case of nil.

ClobberXD:
> Passing an invalid keyword doesn't trigger this problem. I wasn't able to reproduce it, but I guess it might be caused by sending invalid/modified formspec events.

I wasn't able to reproduce it by passing an invalid keyword either. Corrupted formspec events is a possibility. But if that's the case, it might be better to set idx to that of the current map, instead of returning with an error message.

Thomas--S:
I chose this simple solution with the error message because of two reasons:
1) It will never happen to a normal player anyway; only players who send modified formspec events. I think it is okay in this case to just return with an error message and provide as little vulnerability possibilities as possible.
2) It might be possible that there is no current map under certain circumstances (No maps are available, no map is active etc.). Writing a sophisiticated logic to take all these cases into account is just too much for this edge case.

ClobberXD:
> It will never happen to a normal player anyway; only players who send modified formspec events.

Packet corruption might cause such occurrences too. At times of extreme server load or client-side lag, players report for example, their medkit HUD overlays not disappearing, although the probability of this happening is 0. I myself have experienced this a couple of times, and I suspect the same might be the cause for the crash.

In this particular case, I suspect a player was on idx 2, and pressed the `<---` button twice, before the formspec had the chance to update and not render the `<---` button. The second event sets their idx to 0, and `ctf_map.available_maps[0]` doesn't exist. I'm unable to reproduce this in singleplayer because it's not laggy enough; increasing `dedicated_server_step` could help?

But the problem here is the lack of sanitation of inputs in the `on_player_receive_events` callback. This is because I originally assumed that the idx can't go out of range because the buttons aren't rendered when they aren't valid, but TOCTOU (time of check ~= time of use). With that in mind, sanitising the idx within the callback is probably the better solution, IMO, because that's the only other part of code that modifies the idx apart from the chat-command.

Thomas--S:
Adding a sanity check there is a very good idea. However, I think this is out of scope for this PR: As I wrote above, this PR simply is a solution for the following problem:

`ctf_map.available_maps[idx]` can return either nil or a map, but the following code relies on the result being a map. So, in order to prevent a crash, we have to abort the execution of this function in case of nil.

Such a safeguard check may only be omitted if we used something like Design by Contract and define as precondition that `idx` must be an existing index value. However, at the current state we don't do this, so we have to implement the safeguard measure.

Sanitizing the formspec events is also useful in order to use a fallback value for idx there, for example, to avoid the error message. I appreciate an implemention of this for a follow-up PR.

ClobberXD:
I'm ok with sanitising user input in a later PR. This PR is good to go.

PR #516, opened 2019-11-24
Remove stone sword from default treasure registration

JostP:
Stone sword appear in chests where you don't get it as initial stuff (caverns, IceAge...). Noone takes them from chest and they are there unnecessarily. Instead of getting stone sword we will get better items.

Maybe all stone tools should be removed?


mods/ctf/ctf_treasure/init.lua
```diff
@@ -9,13 +9,12 @@ function ctf_treasure.get_default_treasures()
...
-		{ "default:shovel_stone", 0.6, 5, { 1, 10 } },
-		{ "default:shovel_steel", 0.3, 5, { 1, 10 } },
-		{ "default:axe_steel",    0.4, 5, { 1, 10 } },
-		{ "default:axe_stone",    0.5, 5, { 1, 10 } },
+		{"default:pick_steel",   0.5, 5, { 1, 10 } },
+		{"default:sword_steel",  0.4, 5, { 1,  4 } },
+		{"default:shovel_stone", 0.6, 5, { 1, 10 } },
+		{"default:shovel_steel", 0.3, 5, { 1, 10 } },
+		{"default:axe_steel",    0.4, 5, { 1, 10 } },
+		{"default:axe_stone",    0.5, 5, { 1, 10 } },
```
ClobberXD: Missing space between `{"` in all changed lines

JostP: Fixed

PR #517, opened 2019-11-26
Add ctf_stats.is_pro; consolidate "pro-ness" calculations

ClobberXD:
This PR consolidates existing "pro-ness" calculations (basically 1000+ score and 1.5+ K/D) strewn across the code-base into one API function `ctf_stats.is_pro`, and this function call now replaces all individual checks. This would prevent code duplication, and would allow the criteria to be modified more easily without having to sync it across all the existing checks. And more importantly, this is the right way to do it. ;)

Tested; works.

### DIFF
--- a/mods/ctf/ctf_bounties/init.lua
+++ b/mods/ctf/ctf_bounties/init.lua
@@ -27,7 +27,7 @@ local function bounty_player(target)
 	-- bounty_score = -----------, or 500 (whichever is lesser)
 	--                   5000
 
-	local pstat, _ = ctf_stats.player(target)
+	local pstat = ctf_stats.player(target)
 	if pstat.deaths == 0 then
 		pstat.deaths = 1
 	end
@@ -44,7 +44,7 @@ local function bounty_player(target)
 		for _, player in pairs(minetest.get_connected_players()) do
 			local name = player:get_player_name()
 			if bountied_player ~= name then
-				local prev_color = ctf_colors.get_color(prev, ctf.player(prev)).css
+				local prev_color = ctf_colors.get_color(ctf.player(prev)).css
 				minetest.chat_send_player(player:get_player_name(),
 					minetest.colorize("#fff326", "Player ") ..
 					minetest.colorize(prev_color, prev) ..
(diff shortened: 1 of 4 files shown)


mods/ctf/ctf_map/chest.lua
```diff
@@ -68,8 +59,7 @@ for _, chest_color in pairs(colors) do
 				"list[current_player;main;0,9.08;8,3;8]",
 			}, "")
 
-			local pstat = ctf_stats.player(name)
-			if not pstat or not pstat.score or pstat.score < 10 then
```
ClobberXD: The check `not pstat or not pstat.score` is redundant, because `ctf_stats.player` will automatically initialise pstat and all its values to 0 if they don't exist.

ClobberXD:
Rebased

Thomas--S:
I experienced a crash related to bounties while reviewing this, but I didn't examine the reason for it yet. Needs further testing.

ClobberXD:
The error message would be helpful, in case the crash is related to this PR

Thomas--S:
```
2019-12-01 18:46:20: ERROR[Main]: ServerError: AsyncErr: environment_Step: Runtime error from mod 'ctf_bounties' in callback environment_Step(): ...inetest/games/capturetheflag/mods/ctf/ctf_stats/init.lua:200: c
tf_stats.is_pro | Invalid player!
2019-12-01 18:46:20: ERROR[Main]: stack traceback:
2019-12-01 18:46:20: ERROR[Main]:       [C]: in function 'assert'
2019-12-01 18:46:20: ERROR[Main]:       ...inetest/games/capturetheflag/mods/ctf/ctf_stats/init.lua:200: in function 'is_pro'
2019-12-01 18:46:20: ERROR[Main]:       ...test/games/capturetheflag/mods/ctf/ct
f_bounties/init.lua:65: in function 'func'
2019-12-01 18:46:20: ERROR[Main]:       /usr/local/share/minetest/builtin/common/after.lua:20: in function </usr/local/share/minetest/builtin/common/after.lua:5>
2019-12-01 18:46:20: ERROR[Main]:       /usr/local/share/minetest/builtin/game/register.lua:429: in function </usr/local/share/minetest/builtin/game/register.lua:413>
```

ClobberXD:
Interesting... The assertion if `ctf_stats.players[name]` is nil, which leads me to believe that the player you're testing this with has no score. Nevertheless, this is related to this PR, and I'll try to fix it in the evening.

Thomas--S:
> that the player you're testing this with has no score

This is perfectly possible, I used a new account as well as two already existing (pro) accounts for testing this.

ClobberXD:
@Thomas--S I've made `ctf_stats.is_pro` to now use `ctf_stats.player(name)` instead of directly accessing `ctf_stats.players[name]`. (Unfortunately) I don't have the time to test this change, but I'm sure it'd work just fine. But do re-test the same scenario, and let me know if the problem has been fixed.

Thomas--S:
Ok. Thanks. I will try to re-test and review this evening or during the next few days.

ClobberXD:
Ok cool. I've removed the "testing needed" label, as it's reserved for PRs that require *extensive* testing (like the shooter submodule PR or even classes). Normal testing already falls under "Ready for review" label. :)

Thomas--S:
I originally added this label because the error needed further investigation, but as you found the error quickly, the label became obsolete here.

PR #521, opened 2019-12-04
Change the menu icon for CTF

MinetestSam:


ClobberXD:
The texture seems to be very blurry, and doesn't look as crisp as the old one.

MinetestSam:
> The texture seems to be very blurry, and doesn't look as crisp as the old one.
Yes, sadly no image editor allows you to rotate an image without it getting blurry.

ClobberXD:
You're right, but this shouldn't be too difficult to do from scratch, IMO. The art style we follow makes it easier to re-create a texture from scratch, than to rotate it and fix the resulting blurriness.

MinetestSam:
Rather than adding two flags should we just have 'Capture The Flag` in text?

ClobberXD:
Nooooo, why would ya want dat? :/

MinetestSam:
Realised how bad I was at drawing pixel art.... so closing the PR.

PR #522, opened 2019-12-07
setup_maps.sh: Fix file not found

Thomas--S:
~Also add a basic EditorConfig file~

### DIFF
--- a/setup_maps.sh
+++ b/setup_maps.sh
@@ -4,14 +4,18 @@ cd mods/ctf/ctf_map/ctf_map_core/maps/
 for f in *; do
 	if [ -d ${f} ]; then
 		# Screenshot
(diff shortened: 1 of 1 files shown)


ClobberXD:
I avoided the "if file then copy" checks per-file, to avoid cluttering the script, at the cost of a couple of harmless pesky warnings. But checking only for the first skybox and including all of them is a neat idea. :+1: 

About `.editorconfig`, I didn't know this even existed. But I read up on this, and I think this would be a good addition to the code-base. But as this file isn't related to this PR, I'd prefer adding this file in a separate PR.

ClobberXD:
Oh, `.editorconfig` has been removed, I see. I didn't get any notification regarding this, sorry.

PR #523, opened 2019-12-07
Add a basic EditorConfig file

Thomas--S:


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = tab
+insert_final_newline = true
+max_line_length = 120
```
ClobberXD: Whoa! We do 80, or 90 max, but 120 is way too much! :O

Issue #525, opened 2019-12-15
Take captures into consideration for "pro" players.


MinetestSam:
Currently, pro players are those who have a score more than 10,000 and a K/D of more than 1.5. The captures also need to be added as a criteria in pertaining a pro player.

PR #527, opened 2019-12-15
Consider captures as a requirement for pro players

MinetestSam:
Closes: #525

### DIFF
--- a/mods/ctf/ctf_map/ctf_map_core/chest.lua
+++ b/mods/ctf/ctf_map/ctf_map_core/chest.lua
@@ -83,7 +83,7 @@ for _, chest_color in pairs(colors) do
 			else
 				formspec = formspec .. "label[4.75,3;" ..
 					minetest.formspec_escape("You need at least 10000" ..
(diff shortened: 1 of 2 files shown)


ClobberXD:
Fixed merge conflict. Unless someone's playing during the dead hours, I think 20 might be a bit too much. I say we start with 10, and increase it if necessary.

MinetestSam:
> Unless someone's playing during the dead hours, I think 20 might be a bit too much. I say we start with 10, and increase it if necessary.

Yeah I also think this is a hit and trial process and we will only be able to determine the correct value after several tests

ClobberXD:
10 seems reasonable at first glance, but I'd prefer to test this before actually approving. Also, marking this PR as low priority as this won't be a part of 2.0.0 as we've earlier discussed.

fjmng:
The formspec to unlock the pro section of the chest and the random messages should be updated

ClobberXD:
@fjmng Please continue this discussion in #541, as this PR has been closed.

PR #530, opened 2019-12-18
tsm_chests: Prevent chests being placed on top of other chests

ClobberXD:
Because it looks ugly. Contains another commit which makes blacklisting nodes a lot simpler.

Tested; works.

To test, skip to a map that has a chest zone crammed with chests (e.g. Tunnels), and ensure that there are no chests on top of each other. Maybe also repeat this step a couple of times for good measure.

### DIFF
--- a/mods/other/tsm_chests/init.lua
+++ b/mods/other/tsm_chests/init.lua
@@ -17,6 +17,13 @@
 
 tsm_chests = {}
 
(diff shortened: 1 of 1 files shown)


fjmng:
tested too - works

ClobberXD:
That's great. Thanks for testing!

PR #531, opened 2019-12-18
ctf_bounties: Prevent the previous target from being selected again

ClobberXD:
If this isn't trivial, I don't know what is...


ClobberXD:
Tested by dumping the players table into the terminal, and reducing bounty selection interval to 5s:

- My name was in the table the first time
- I got selected as the bounty target
- My name wasn't in the table the next couple of iterations (I was the only pro, so no other players were selected)
- I manually set another player as bounty target
- Sure enough, my name appears in the table in the next iteration.

PR #535, opened 2019-12-18
Afkkick message

TSafa-23:
Pretty self explanatory. It sends a message to everyone that the specified player has been kicked for inactivity. NOTE: Please do not merge or close until I get home, around 16:30 CDST so I can read possible comments or concerns.


MinetestSam:
What would be the possible benefit of this?

TSafa-23:
So players could see when a player leaves when they are afk for too long. (also that afkkicked players usually rejoin right afterwards so players can assume that the player will rejoin. In a matter of fact, I should just change the message to *** <player> left the game (afk))

ClobberXD:
> *** left the game (afk)

Kicking would already show the OG player leave message, and I don't recommend using the exact same message just to display the reason.

PR #536, opened 2019-12-19
(Re-)add nodes from MTG

ClobberXD:
screenshot_20191219_163044

(`flowers` were added after taking the screenshot)

Mods added: `fire`, `flowers`, `vessels`, `walls`, `xpanes`.

All included mods have been trimmed to exclude unnecessary bloat like growth, spread, node inv., ABMs, LBMs, etc.

Tested, everything works.


MinetestSam:
Everything looks fine just had two doubts in my mind,
1) Should the fire do 4 damage per second? If it is a part of decor then I think it should be rather low to about 0.5 or 1.
2) Do we need to have the names of the creators of textures or models in different languages?

ClobberXD:
1. Fire isn't a part of decor, and shouldn't be considered as such. It has been added as another node in the map maker's arsenal, which they can use in any way they deem fit. And for the same reason, fire has been left as is.

2. I don't really know if these are even worth translating; we already have an issue for adding client-side translations. I don't see anything wrong with adding them, and so I did not remove them.

Btw, what exactly do you mean by "names of the creators of textures or models in different languages?"? None of the texture/model author names have been translated...

MinetestSam:
> Btw, what exactly do you mean by "names of the creators of textures or models in different languages?"? None of the texture/model author names have been translated...

Yeah I realised that later when I read through it again..

mods/mtg/fire/locale/fire.it.tr
```diff
@@ -0,0 +1,3 @@
+# textdomain: fire
+Permanent Flame=Fiamma permanente
+Flint and Steel=Acciarino
```
LoneWolfHT: Do translation files need a newline at the end of them?

ClobberXD: A newline at EOF is always for the best, unless it breaks the parsing system in some way. Since most other translation files do have a newline at the end, it's most likely safe to do so. I'll attend to this within a couple of days. Thanks.

PR #539, opened 2019-12-21
Improve match duration metric; move to ctf_match

ClobberXD:
- Match's start time is initialised only after the buildtime ends. This isn't the case in `master`, and the buildtime is also taken into consideration.
- Since this is a very fundamental metric relating to matches, match duration code has been moved from `ctf_stats` to `ctf_match`.
- A helper function `ctf_match.get_match_duration`, has been added to remove code duplication. Returns `nil` if match start time hasn't been initialised yet.
- If the match summary formspec is displayed when the match start time hasn't been initialised, the duration field in the header would contain `-` instead of `HH:MM:SS`.

#### Testing

- View match summary during build-time. Observe that the header-field duration reads `-`.
- View match summary after build-time ends (`/ctf_start`). Observe that the duration has started only when the build-time ended.
- Complete match. Observe the correct duration being displayed in the end-of-match summary.
- Skip a match during buildtime. Observe the end-of-match summary duration to be `-`.

### DIFF
--- a/mods/ctf/ctf_match/init.lua
+++ b/mods/ctf/ctf_match/init.lua
@@ -15,7 +15,12 @@ ctf.register_on_init(function()
 	minetest.settings:set_bool("enable_pvp", true)
 end)
 
+ctf_match.register_on_build_time_start(function()
+	ctf_match.match_start_time = nil
+end)
+
 ctf_match.register_on_build_time_end(function()
+	ctf_match.match_start_time = os.time()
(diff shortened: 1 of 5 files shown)


fjmng:
tested - works

LoneWolfHT:
Tested and found only one problem (Listed on Discord), which I think is unrelated to this PR

ClobberXD:
Fixed trivial merge conflict. Merging tonight, as this PR has been tested by three people (incl. me).

PR #540, opened 2019-12-22
Make bounties secretive

MinetestSam:


Thomas--S:
To be honest, I'm not really sure about this concept.

I think that the current bounty system supports team spirit (by telling the bounty that they are bounty) and also adds variety to the game because players try to get the price.

With the new system, there won't be any of these benefits mentioned above because gameplay would just continue as if there was no bounty. The price can only be won by chance and basically just acts like a random kill score.

So, in conclusion a :-1: from me for the concept.

MinetestSam:
Well I think this is rather a step in the right direction due to the following points,
1) When players do have bounty on their head they are usually told that by their teammates but it is not the case every time. Some teammates don't tell them. Furthermore, what is the point of the bounty system if the enemy is even telling you that you have a bounty on your head? The CTF community is pretty friendly and usually tell their friends who even if are on the enemy team that they have bounty.
2) Enemy corners the person who has bounty: The players in the enemy team, only focus on the player who has bounty because the risk-reward is high and this leads to player who has bounty being focused on and ruining their gameplay.
3)The main idea behind bounty is that the player who is bounty shouldn't know that they are bounty and thus with everyone telling them it doesn't work.
4) If bounties are made secretive then players would play normally and wouldn't be able to focus on one person just for the bounty prize. They'll have to pick and choose their fights and even go with their teammates into fights.

Thomas--S:
To 1.) I think we should encourage this friendlyness, even across teams, because a friendly atmosphere will lead to less rule violations.

To 2.) I think this adds variation to the gameplay, secret bounties wouldn't change anything regarding gameplay.

To 3.) This is a valid point, but I think especially point 1 outweighs this.

To 4.) I think the bounty system could be removed alltogether then because it wouldn't be more than a random addition to the normal score.

As a solution to our contrary opinions, I'd suggest a poll.

ClobberXD:
> To 1.) I think we should encourage this friendlyness, even across teams, because a friendly atmosphere will lead to less rule violations.

You do realise that CTF is centred around hacking and slashing other players and/or blowing their heads right off with a shotgun, right? :P

If a fair chunk of players (because nubs don't count) are already able to maintain friendly relations with other players while massacring them once every two minutes, then this won't affect their friendships in any way, I believe.
 
> To 2.) I think this adds variation to the gameplay, secret bounties wouldn't change anything regarding gameplay.

Secret bounties would mean that no one would know who's the bounty, and it'd be a little more thrilling, due to the added gameplay element of surprise.

In this regard, I think the fact that a player has been chosen as a bounty should be announced to all the players, while still withholding the target name and the prize on their head.
  
> To 4.) I think the bounty system could be removed alltogether then because it wouldn't be more than a random addition to the normal score.

Well, that's what a bounty literally is - it's an extra sum of money (read score) awarded to someone who kills a notorious criminal (i.e. player - all players are mass-murderers in CTF :D)
 
> As a solution to our contrary opinions, I'd suggest a poll.

-_-

mods/ctf/ctf_bounties/init.lua
```diff
@@ -2,25 +2,20 @@ local bountied_player = nil
 local bounty_score = 0
 
 local function announce(name)
-	local tcolor = ctf_colors.get_color(ctf.player(bountied_player))
 	minetest.chat_send_player(name,
-			minetest.colorize("#fff326", "The next person to kill ") ..
-			minetest.colorize(tcolor.css, bountied_player) ..
-			minetest.colorize("#fff326", " will receive " .. bounty_score .. " points!"))
+			minetest.colorize("#fff326", "Someone has") ..
+			minetest.colorize("#fff326", " been chosen as bounty!"))
```
ClobberXD: I had previously added `minetest.colorize` for each line because the line in between (comprising of the target's name) required a different colour. Since this is now all using one single colour, there's no need for two `minetest.colorize` calls.

mods/ctf/ctf_bounties/init.lua
```diff
@@ -94,7 +84,13 @@ ctf.register_on_killedplayer(function(victim, killer)
...
 
-	local msg = killer .. " has killed " .. victim .. " and received the prize!"
+	local killer_color = ctf_colors.get_color(ctf.player(killer)).css
+	local victim_color = ctf_colors.get_color(ctf.player(victim)).css
+	local msg = minetest.colorize(killer_color, killer) ..
+				minetest.colorize("#fff326"," has killed ") ..
+				minetest.colorize(victim_color, victim) ..
+				minetest.colorize("#fff326", " and received ") ..
+				minetest.colorize("#fff326", bounty_score .. " points!")
 	minetest.log("action", msg)
```
ClobberXD: `msg` contains colour-codes, which the would end up polluting the logs with stuff similar to `(C@fff326)`. The following would strip out the colour codes and log only the plain message.

```suggestion
	minetest.log("action", minetest.strip_colors(msg))
```

mods/ctf/ctf_bounties/init.lua
```diff
@@ -94,7 +84,13 @@ ctf.register_on_killedplayer(function(victim, killer)
 	end
 	bountied_player = nil
 
-	local msg = killer .. " has killed " .. victim .. " and received the prize!"
+	local killer_color = ctf_colors.get_color(ctf.player(killer)).css
+	local victim_color = ctf_colors.get_color(ctf.player(victim)).css
+	local msg = minetest.colorize(killer_color, killer) ..
+				minetest.colorize("#fff326"," has killed ") ..
```
ClobberXD:
```suggestion
				minetest.colorize("#fff326", " has killed ") ..
```

mods/ctf/ctf_bounties/init.lua
```diff
@@ -94,7 +84,13 @@ ctf.register_on_killedplayer(function(victim, killer)
...
+				minetest.colorize(victim_color, victim) ..
+				minetest.colorize("#fff326", " and received ") ..
+				minetest.colorize("#fff326", bounty_score .. " points!")
```
ClobberXD: Again, no need for two consecutive `minetest.colorize` calls for the same colour.

Thomas--S:
> Well, that's what a bounty literally is - it's an extra sum of money (read score) awarded to someone who kills a notorious criminal (i.e. player - all players are mass-murderers in CTF :D)

But I guess that one should know in before who is the bounty target. It'd make little sense to just kill a random person and then see if they had a bounty on their had.

ClobberXD:
> But I guess that one should know in before who is the bounty target. It'd make little sense to just kill a random person and then see if they had a bounty on their had.

True, but that'd unfortunately result in a team-vs-team turning into a manhunt.

Thomas--S:
> True, but that'd unfortunately result in a team-vs-team turning into a manhunt.

To be honest, I haven't really experienced this to an unacceptable amount on the server.

MinetestSam:
IMO, the messages are also quite vague so I need some suggestions regarding it

MinetestSam:
> I think that the current bounty system supports team spirit (by telling the bounty that they are bounty) and also adds variety to the game because players try to get the price.

I had been on the server two times and in both cases I saw that not many players were interested in telling the player who was bounty that they had bounty on their head. In the first case, a well known player in the community `Tozikor` had bounty on his head and many pros were online but only one player told him that he was bounty and that was his teammate. In the second case, a player who also plays regularly named `Virus_` had bounty and also at that time quite a lot of pros were on but in this case no one from his team told him that he was bounty and he was told by an enemy that he had bounty.

MinetestSam:
> With the new system, there won't be any of these benefits mentioned above because gameplay would just continue as if there was no bounty. The price can only be won by chance and basically just acts like a random kill score.

The current system is also the same IMO. It is chance that you got to fight the player who had bounty or the bounty popped up when both the players were fighting. It is very hard to even read nametags so it is very random that you ended up fighting the player who had bounty.

Thomas--S:
> and he was told by an enemy that he had bounty

And this is one of the things that I like about the bounty system. Players see themselves as friends, even across team borders. They help each other on a personal level, but not on a gameplay level, where they are opponents; similiar to a soccer match under friends. It helps to maintain the friendly atmosphere of the server.

---

> The current system is also the same IMO. It is chance that you got to fight the player who had bounty or the bounty popped up when both the players were fighting. It is very hard to even read nametags so it is very random that you ended up fighting the player who had bounty.

> Enemy corners the person who has bounty: The players in the enemy team, only focus on the player who has bounty because the risk-reward is high and this leads to player who has bounty being focused on and ruining their gameplay.

Both quotes are from you, and it seems to me like those are nearly a contradiction.

However, I think both quotes are a little bit exaggerated and the reality is in between those two scenarios most of the time.

Because it is rather difficult to focus on a certain player there won't be "manhunts". But there will be a greater incentive to fight against a certain player because of the bounty prize. This will add a bit of variation to the gameplay without ruining it.

---

So, in conclusion, I can say that I am still opposed to secret bounties.
If you are keen on them, I'd suggest to make a poll so that we can see the players' opinion. If a majority of the players really supports this change, I can accept it (although I personally still don't approve it.)

MinetestSam:
I may have lost the battle but I'LL RETURN...... closing the PR

PR #542, opened 2019-12-23
Add /s chat-command alias for /summary

ClobberXD:
Super-trivial. Tested; works.

### DIFF
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -51,20 +51,28 @@ local function return_as_chat_result(to, target)
 	return result
 end
 
+local function summary_func(name)
+	local fs = ctf_stats.get_formspec_match_summary(ctf_stats.current,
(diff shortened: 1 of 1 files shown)


ClobberXD:
Some have raised questions as to whether `/summary` is needed any more. I agree that `/summary` will never be used, because there's literally no difference between it and `/s`. Shall we remove it, or keep it for the sake of backwards-compat?

Thomas--S:
I'd prefer to keep it. It will be easier to understand when players skim through the available chat commands with `/help`.

ClobberXD:
Well, we've got random messages to help us with that. We could replace `/summary` with `/s` in that message. Also, `/help all` will list all commands along with their help text.

Currently, `/r` is linked to `/rankings` in only the random message, and the help texts of neither of these commands say anything about the other.

ClobberXD:
Going by the current consensus, `/summary` is going to remain alongside `/s`. I'll consider this PR ready for merge.

PR #543, opened 2019-12-26
Make ind_glass_red pointable

TSafa-23:
This is so players can't take from chests on the other side of the barrier before the round begins.


Thomas--S:
As far as I know, it is an intended feature that chests on the other side near the barrier are accessible.

ClobberXD:
I too feel that stealing items from chests on the other side is okay, and I'd prefer to keep the barrier nodes unpointable. Sorry, but :-1:

PR #546, opened 2019-12-28
Fix colourization of messages in help page

MinetestSam:
Tested and works

### DIFF
--- a/mods/ctf/ctf_inventory/init.lua
+++ b/mods/ctf/ctf_inventory/init.lua
@@ -1,7 +1,7 @@
-local colour = "#66A0FF"
+local color = "#66A0FF"
 local items = {
 	"",
-	colour .. "Game Play",
+	color .. "Game Play",
 	"",
 	"*  Capture an enemy's flag by walking up to it, punching it, then",
 	"   running and punching your team's flag.",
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_inventory/init.lua
```diff
@@ -12,21 +13,21 @@ local items = {
 	"   killing 2 people for every death.",
 	"",
 
-	minetest.colorize("#66a0ff", "Team Co-op"),
+	"#66a0ffTeam Co-op",
```
Thomas--S: I think that string concatenation like shown in the example below might improve readibility for a neglectible performance impact.

```suggestion
	"#66a0ff" .. "Team Co-op",
```

MinetestSam: I agree. Will add that.

ClobberXD: If you're keeping the colour separate, you might as well store it in a local var to keep things nice and clean.

ClobberXD: i.e.

```lua
local color = "#66A0FF"

-- *burp*

color .. "Team Co-op",

-- *Hachooo*

color .. "Player Rankings",

```

mods/ctf/ctf_inventory/init.lua
```diff
@@ -1,6 +1,7 @@
+local colour = "#66A0FF"
 local items = {
 	"",
-	"#66a0ff" .. "Game Play",
+	colour .. "Game Play",
```
ClobberXD: Trivial nit:
CTF follows MT's code-style closely, and like MT, we use American English in all places to prevent ambiguity. So please rename this var to `color`. :P

MinetestSam: *Sigh*

PR #548, opened 2019-12-31
Change the build time based upon the number of players online

MinetestSam:
The aim of this PR is to change the build time based on the number of players online. Currently it sets the build time to 2 minutes when 10 or less than 10 players are playing.

### DIFF
--- a/mods/ctf/ctf_match/buildtime.lua
+++ b/mods/ctf/ctf_match/buildtime.lua
@@ -24,11 +24,22 @@ function ctf_match.is_in_build_time()
 	return ctf_match.build_timer > 0
 end
 
+local btime
+local mply = 0.5
+local n = 10 --Number of players required to set the build time to half of what it is
 ctf_match.register_on_new_match(function()
 	ctf_match.build_timer = ctf.setting("match.build_time")
+	if #minetest.get_connected_players() <= n then
+		ctf_match.build_timer = ctf_match.build_timer * mply
+	end
+	btime = ctf_match.build_timer
 end)
 ctf.register_on_new_game(function()
 	ctf_match.build_timer = ctf.setting("match.build_time")
+	if #minetest.get_connected_players() <= n then
+		ctf_match.build_timer = ctf_match.build_timer * mply
+	end
+	btime = ctf_match.build_timer
 	if ctf_match.build_timer > 0 then
 		for i = 1, #ctf_match.registered_on_build_time_start do
 			ctf_match.registered_on_build_time_start[i]()
@@ -71,7 +82,7 @@ end)
 
 ctf_match.register_on_build_time_start(function()
 	minetest.chat_send_all("Prepare your base! Match starts in " ..
-		ctf.setting("match.build_time") .. " seconds.")
+		btime .. " seconds.")
 end)
 
 ctf_match.register_on_build_time_end(function()


Thomas--S:
I'm really sorry for downvoting a PR made by you again and I really appreciate your intention to increase your contributions to this game; but I think that the build time should be independet of the player count.

If only a few players are online, they have to do all the preparation tasks, i.e. building a base and emptying chests, alone, so they need the full build time in order to have a match with well prepared teams.

MinetestSam:
> I'm really sorry for downvoting a PR made by you again.

No worries, I don't make the PRs hoping they would be accepted all the time.
 
> If only a few players are online, they have to do all the preparation tasks, i.e. building a base and emptying chests, alone, so they need the full build time in order to have a match with well prepared teams.

 It depends on what you define as few players. If the goal of CTF is to offer fast gameplay with shorter rounds then I think the build time is a major factor here. 4 minutes of the match alone goes in build time and many players are ready before that. Also, when there are low number of players online not all the players help in making the base or loot all the chests even with a 4 minute build time. With lower number of players, a 4 minute build time is really annoying and they want to be involved in action right away and not wait for 2 minutes for the barrier to be removed.

Thomas--S:
I can understand the concerns regarding a too long build time.

However, I don't understand inhowfar the player count should influence the build time.

To be honest, when I originally saw the title of this PR I thought of exactly the opposite of what is done here: I thought that the build time would be shorter when there are more players online as the "work" can be distributed better.

Additionally, I'm not sure if rubenwardy would appreciate a shorter build time, one year ago he suggested to increase the build time to 6 minutes.

MinetestSam:
> However, I don't understand inhowfar the player count should influence the build time.

Let us take the example that only four players are online. Now they don't want to wait for 4 minutes just for the battle to start and slowly they will lose interest due to long build time. If the build time is short then there will be faster rounds and retention rate would be more

> To be honest, when I originally saw the title of this PR I thought of exactly the opposite of what is done here: I thought that the build time would be shorter when there are more players online as the "work" can be distributed better.

I understand that, but the work isn't distributed necessarily. When more players are online, players need to prepare their base for everything like tunnelers or those who may make a sky bridge which won't happen when lower number of players are online. Also there may be cases when higher number of players are on, that they didn't find any iron sword or spares so they need to and mine for iron or other resources which takes time and is a part of preparation. Again when lower number of players are online, generally there is a good distribution of swords and there are spares too.

> Additionally, I'm not sure if rubenwardy would appreciate a shorter build time, one year ago he suggested to increase the build time to 6 minutes.

His idea was rejected by the players by a majority in the poll.

Thomas--S:
> Let us take the example that only four players are online. Now they don't want to wait for 4 minutes just for the battle to start and slowly they will lose interest due to long build time. If the build time is short then there will be faster rounds and retention rate would be more

I think it is wrong to think of the build time as "waiting" time. It should be used to prepare for the match.
When I experienced matches with only very few people online (2-6 players), I always needed the full 4 minutes to gather some gear and build an acceptable base.

However, I can understand your concerns and that especially non-experienced players won't use the full build time.
Instead of generally shorten the build time for short matches, I'd suggest the following solution:

If there are less than at least one pro per team online, the reduced build time of this PR should be used.
Else, the pro players should signalize with a `/ready` command that they don't need the rest of the build time. As soon as the default build time is elapsed or at least 50% of the pros of each team is ready, the match should start.

The `/ready` command can be advertised to pro players as the `/yes` and `/no` is advertised during a vote.

I think this approach combines the benefits of both points of view.

ClobberXD:
Lower players can either mean more buildtime (as Thomas-S said) or less buildtime, depending on other factors like the "pro-ness" of the players, their play-style, team balance. I really agree with the fact that a longer buildtime is going to be very boring, but I'm inclined to overlook this concern for the aforementioned reason.

Regarding `/ready`: I'm apprehensive about giving players more control than is required. We should keep this in mind: a player qualifying as a pro doesn't mean they're intelligent. As in, I wouldn't trust 8-year-olds and 10-year-olds to consider the preparedness of their base/team before entering `/ready`.

MinetestSam:
I made this PR because some times I was the only one on the server and had to wait every 4 minutes just for the match to start. This was very long for me and made the game boring.

Thomas--S:
> I made this PR because some times I was the only one on the server and had to wait every 4 minutes just for the match to start. This was very long for me and made the game boring.

I'd be fine with reducing the build time if the total player count is <= 1.

In a 1vs1 scenario however, I think that the full build time is required to be prepared appropriately.

MinetestSam:
> In a 1vs1 scenario however, I think that the full build time is required to be prepared appropriately.

I don't think people will put an effort in to make a base when they know the rounds are going to be shorter

Thomas--S:
A 1vs1 match can last rather long if there are good bases.

A good base is especially important in a 1vs1 scenario because you can't attack and defend at the same time, so the base should protect the flag while you are away.

MinetestSam:
> A 1vs1 match can last rather long if there are good bases.

IMO, yes although 1v1s can be long but they shouldn't be long.

> A good base is especially important in a 1vs1 scenario because you can't attack and defend at the same time, so the base should protect the flag while you are away

If the player is away. it takes probably only a second to mine the cobble structure or use rifles or shotguns to break the blocks. Although I agree that you need to make your base stronger but I don't think that should be the case in 1v1s as you'd want small and engaging rounds rather than just your base and camping there.

Thomas--S:
I agree that 1vs1 matches won't be very long compared to 10vs10 matches on the walls map with 2 hours and more.

However, I think that very short rounds are even more boring, they tend to have little variation and the build time (no matter if it's shortened or not) disrupts the game too often.
IMHO, we should see medium length (15min probably) matches as desirable for 1vs1.

And regarding the one second through the wall:
If a good base can be built during the build time, it's very well possible to make one that lasts for at least 10 seconds. This is a lot of valueable and usable time for the defender.

MinetestSam:
With the addition of classes I believe that the gameplay has changed totally and this is not required anymore.

PR #549, opened 2020-01-01
Add some modern weapons and vehicles

zaoqi:
This PR add some modern weapons and vehicles.
`dye`, `fire` and `tnt` are from minetest_game 5.0.0
`vehicles` is from https://github.com/D00Med/vehicles


LoneWolfHT:
:thumbsdown: for vehicles. They're buggy and don't really work on most of the terrain.
What modern weapons are added? I can't see much in the flood of other seemingly unneeded code additions

MinetestSam:
As LW said, vehicles are quite laggy and buggy and that combined with ctf lag is going to be a mess. TNT will also cause quite a bunch of lag and probably crash the server. As for dye, I don't see any use for them since CTF only allows crafting of certain stuff and dyes won't be used by map creators. For fire, see #536. Since the weapons are provided by the vehicles, I don't see the use for them.

PR #554, opened 2020-01-06
Disallow transfer of rankings if source and destination are same

### LINKED ISSUE #553: /r crashes the the server after /transfer_rankings to self
qwe_download
I did /transfer_rankings fjmng fjmng, and after doing /r I got this. You should make it so you cant transfer to yourself

_Originally posted by @fjmng in https://github.com/MT-CTF/capturetheflag/pull/194#issuecomment-570656415_

MinetestSam:
Closes #553

### DIFF
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -183,8 +183,8 @@ minetest.register_chatcommand("transfer_rankings", {
 		if not ctf_stats.players[dest] then
 			return false, "Player '" .. dest .. "' does not exist."
 		end
-		if ctf.player(src) == ctf.player(dest) then
-			return false, "You cannot transfer your own rankings!"
+		if src == dest then
+			return false, "Source name and destination name cannot be the same!"
 		end
 
 		ctf_stats.players[dest] = ctf_stats.players[src]
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -183,6 +183,9 @@ minetest.register_chatcommand("transfer_rankings", {
 		if not ctf_stats.players[dest] then
 			return false, "Player '" .. dest .. "' does not exist."
 		end
+		if ctf.player(src) == ctf.player(dest) then
+			return false, "You cannot transfer your own rankings!"
+		end
 
 		ctf_stats.players[dest] = ctf_stats.players[src]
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_stats/chat.lua
```diff
@@ -183,6 +183,9 @@ minetest.register_chatcommand("transfer_rankings", {
 		if not ctf_stats.players[dest] then
 			return false, "Player '" .. dest .. "' does not exist."
 		end
+		if ctf.player(src) == ctf.player(dest) then
```
ClobberXD: You're checking for the team-data instead of player stats. Instead, you can directly compare the names.

mods/ctf/ctf_stats/chat.lua
```diff
@@ -183,6 +183,9 @@ minetest.register_chatcommand("transfer_rankings", {
...
+			return false, "You cannot transfer your own rankings!"
```
ClobberXD: I'd prefer to avoid active voice here, because the caller needn't be the same as the player whose stats are being transferred. Instead, I'd prefer something similar to the following:

```suggestion
			return false, "Source name and destination name cannot be the same!"
```

Thomas--S:
I approve the concept of this fix, but strictly speaking, this isn't really a fix for #553 because it only tries to prevent the initial situation, but without fixing the actual problem that in [chat.lua](https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_stats/chat.lua#L26) some variables might not be initialized appropriately.

ClobberXD:
The actual problem is that the source stats was set to `nil` after the transfer - the stats were essentially reset.

BTW, the variable you linked to has been declared in L8, and initialised and modified in subsequent lines.

Thomas--S:
> BTW, the variable you linked to has been declared in L8, and initialised and modified in subsequent lines.

It will not be initialized if `type(target) == "string"` and no entry with `pstat.name == name` exists in `players`.

That's most likely the cause of the crash found in #553 (https://user-images.githubusercontent.com/59484290/71741204-2de0e500-2e56-11ea-939e-db05d6a115ed.png).

ClobberXD:
Hmm, so `place` not being initialised in that specific case is intentional, but I'm performing the wrong check in L26 like an idiot, lol. Should've checked if it was `nil` instead. Thanks for pointing this out. :)

Issue #534, opened 2019-12-18
Re-add removed nodes from MTG


ClobberXD:
This was removed before custom crafting was introduced, probably because the recipes required all sorts of random items. But craftable or not, these nodes allow map makers to add variety to their maps.

Thomas--S:
:+1:

ClobberXD:
#536

PR #558, opened 2020-02-22
Add basic TNT from mtg

An0n3m0us:
Ref: #15 

Added the TNT mod from Minetest Game and removed effects from it to reduce lag.
I've left it so that when the TNT explodes it throws entities.


ClobberXD:
#534 re-adds the `fire` mod, trims out a lot of the unnecessary stuff from it. It might be worth waiting for that PR to be merged before we decide on the direction this PR is going to take.

PR #581, opened 2020-04-19
Rename on_skip_map to on_skip_match

ClobberXD:
This simple PR renames the callback `ctf_match.register_on_skip_map` to `ctf_match.register_on_skip_match`, because it makes more sense. Ready for review; will test in a bit.


ClobberXD:
Tested, works. To test, skip a match. Make sure nothing crashes.

PR #592, opened 2020-04-26
Sniper class

ClobberXD:
- Spawns with
  - `sniper_rifles:rifle_762`
  - 2x `grenades:smoke`
- Allowed guns
  - Pistol
  - SMG
  - 7.62mm rifle
  - Magnum rifle
- Unlocked crafting recipes
  - 7.62mm rifle
  - Magnum rifle
- Shooter perks
  - +25% range
  - +25% faster shooting
- Normal HP and speed.
- Temporarily uses the now-unused rocketeer skin.

****

This PR contains another commit that makes `give_initial_stuff` split itemstacks if the stack count is greater than the item's `stack_max`. This is very useful when adding more than one of a non-stackable item in `initial_stuff` (like weapons and grenades).
Note: Multiple stuff providers providing the same item is still disallowed, as was originally intended.

PR is ready for review.

### Testing instructions

- Verify that snipers are given the specified initial items, and use the rocketeer skin (the skin will mostly be changed later).
- Also verify that the given smoke grenades aren't stacked in the same slot - they should occupy individual slots each.
- Verify that other classes (esp. shooter) can't use the sniper rifles.


LoneWolfHT:
If the grenades PR gets merged it might be neat if the sniper had smoke grenades instead of regular ones

rubenwardy:
> If the grenades PR gets merged it might be neat if the sniper had smoke grenades instead of regular ones

With this, it could be worth having these two classes:

* Assault - SMG + smoke grenades
* Sniper - Sniper rifle + grappling hook

ClobberXD:
Both suggestions sound good. :+1:

ClobberXD:
Added `additional_item_blacklist`. PR is ready for review. Tested, works well.

One thing I noticed is that `shooter:ammo` (being a class item now) can't be dropped or put in chests.

Related: #599 - it isn't a solution per-se, but should ease this problem out.

ClobberXD:
> One thing I noticed is that shooter:ammo (being a class item now) can't be dropped or put in chests.

Fixed. PR is ready for merge.

ClobberXD:
#357 has been merged. I'll need to replace `shooter:grenade` with the new smoke grenades. WIP for now.

mods/other/give_initial_stuff/init.lua
```diff
@@ -28,19 +28,21 @@ function give_initial_stuff.register_stuff_provider(func, priority)
...
+		-- Disallow multiple stuff providers from giving the same items,
+		-- while also allowing each provider to give more than one of the same item
+		local seen_stuff = {}
 		for j=1, #new_stuff do
 			local name = ItemStack(new_stuff[j]):get_name()
 			if not seen_stuff[name] then
 				seen_stuff[name] = true
 				stuff[#stuff + 1] = new_stuff[j]
 			end
+			seen_stuff = {}
```
rubenwardy, line 45: this is definitely a bug

mods/ctf/ctf_classes/classes.lua
```diff
@@ -78,6 +76,51 @@ ctf_classes.register("medic", {
...
+ctf_classes.register("sniper", {
+	description = "Sniper",
+	pros = { "+25% range", "+25% faster shooting" },
+	cons = {},
+	color = "#839",
+	properties = {
+		initial_stuff = {
+			"sniper_rifles:rifle_762_loaded",
+			"shooter_grenade:grenade",
+			"shooter_grenade:grenade",  -- Grenades don't stack, remember?
```
rubenwardy, line 88: This should be `shooter_grenade:grenade 2`, and give initial stuff should support de-stacking

ClobberXD: You're right

mods/other/give_initial_stuff/init.lua
ClobberXD: What's a bug?

mods/ctf/ctf_classes/classes.lua
LoneWolfHT: Once #615 is merged stacked grenades won't break btw

mods/other/give_initial_stuff/init.lua
rubenwardy: you're assigning to the list but never reading from it. See other comment, with the other comment you'd remove this change

mods/ctf/ctf_classes/classes.lua
ClobberXD: > Once #615 is merged stacked grenades won't break btw

Yeah, I noticed. :+1:

ClobberXD: > This should be shooter_grenade:grenade 2, and give initial stuff should support de-stacking over the stack max

@rubenwardy Can we do this in a later PR? We're running pretty late for 2.0 already. :)

ClobberXD:
I've removed the first commit in favour of making `give_initial_stuff` respect items' stack_max. The code is much simpler, and much more flexible now. Tested; works.

ClobberXD:
Another update: Snipers have been disallowed from capping flags - they're supposed to function as "ranged support".

ClobberXD:
:notes: @rubenwardy, can we merge it? :musical_note:
:musical_note: @rubenwardy, yes we CAN!!! :notes:

mods/other/give_initial_stuff/init.lua
```diff
@@ -1,5 +1,15 @@
 give_initial_stuff = {}
 
+-- Add item to inv. Split item if count > stack_max using recursion
+local function give_item(inv, item)
+	inv:add_item("main", item:take_item(item:get_stack_max()))
```
LoneWolfHT, line 5: Shouldn't `get_stack_max()` be `get_count()`?

LoneWolfHT:
Looks like there are merge conflicts

mods/other/give_initial_stuff/init.lua
ClobberXD: From the `take_item` API docs:

```
* `take_item(n)`: returns taken `ItemStack`
    * Take (and remove) up to `n` items from this stack
    * `n`: number, default: `1`
```

> **up to** `n` items from this stack

Even if a number larger than the stack count is passed to `take_item`, it can remove only as much as the stack count. So `item:take_item(item:get_stack_max())` is equivalent to the following:

```lua
item:take_item(math.min(itemdef.stack_max, item:get_count())
```

LoneWolfHT: Ah, guess I half missed the comment too lol

ClobberXD:
> It'd be nice if sniper rifles could be set to not destroy blocks too, you can destroy the enemy base from the opposite side of the map atm. Being able to get sniper rifles instantly with a class will probably make this worse

I'd prefer to keep that separate from this PR. It's a valid request, but it's unrelated to this PR. That being said, feel free to make a PR for that if you'd like to - it's as simple as removing the `groups` in the rifles' item defs.

ClobberXD:
I've updated the first commit to make `ctf_classes` use a global method in `give_initial_stuff` when changing classes. This fixes the bug where the player receives 2 smoke grenades in a single stack when changing classes, unlike when joining/respawning

LoneWolfHT:
I'll look into making it check for the items it gives first...

ClobberXD:
https://github.com/MT-CTF/capturetheflag/blob/baf5f1787815e1a7bbcd62b0ae921acb6098e7bf/mods/ctf/ctf_classes/items.lua#L35-L60

Class-items belonging to the old class are first removed, and the new class items are added. The problem here is that smoke grenades have been omitted from the blacklist on purpose as they can/should be used by all players. So they won't be removed when changing class.

> I'll look into making it check for the items it gives first...

Are you intending to check if the items exist already? I'm not sure if that's the best solution. Take the smoke nades for example, a player might've gathered some from chests or other players - they won't be given 2 nades while changing class because of this.

LoneWolfHT:
> https://github.com/MT-CTF/capturetheflag/blob/baf5f1787815e1a7bbcd62b0ae921acb6098e7bf/mods/ctf/ctf_classes/items.lua#L35-L60
> 
> Class-items belonging to the old class are first removed, and the new class items are added. The problem here is that smoke grenades have been omitted from the blacklist on purpose as they can/should be used by all players. So they won't be removed when changing class.
> 
> > I'll look into making it check for the items it gives first...
> 
> Are you intending to check if the items exist already? I'm not sure if that's the best solution. Take the smoke nades for example, a player might've gathered some from chests or other players - they won't be given 2 nades while changing class because of this.

They technically wouldn't need them though

ClobberXD:
Technically, you're absolutely correct. But treasures are just that, treasures. Treasures are "extras" and shouldn't meddle with what the player is entitled to receive by default. :)

LoneWolfHT:
Alrighty.

Tested ingame: The name text feels a little too dark 
image

Other than that I can't find anything wrong

ClobberXD:
Any suggestions for a better colour? How about yellow or orange?

LoneWolfHT:
> Any suggestions for a better colour? How about yellow or orange?

I like the current color, maybe just brighten to the color at the top of the button

ClobberXD:
Done. Let me know if this one's better.

LoneWolfHT:
That works I guess

ClobberXD:
@LoneWolfHT Colour has been fixed, did you test the rest of the PR while testing earlier?

LoneWolfHT:
Yep, didn't find anything wrong other than that admin priv allowing me to change class at the enemy flag

ClobberXD:
Ok, thanks. I'm assuming that's an approval.

PR #594, opened 2020-04-26
Increase the speed of medic by 10%

MinetestSam:


ClobberXD:
:+1: for the concept, but 20% might be a little too high, IMO. I suggest we start with 10%, and increase it if need be.

PR #601, opened 2020-04-30
Move flag-bearer slowdown to a separate speed modifier

ClobberXD:
Improves separation; less prone to error.

Kinda identical to #390.

### DIFF
--- a/mods/ctf/ctf_match/matches.lua
+++ b/mods/ctf/ctf_match/matches.lua
@@ -80,10 +80,20 @@ function ctf_match.create_teams()
 	error("Error! Unimplemented")
 end
 
+ctf_flag.register_on_pick_up(function(name)
+	physics.set(name, "ctf_match:flag_mult", { speed = 0.9 })
+end)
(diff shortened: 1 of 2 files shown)


mods/ctf/ctf_match/matches.lua
```diff
@@ -80,11 +80,21 @@ function ctf_match.create_teams()
 	error("Error! Unimplemented")
 end
 
+ctf_flag.register_on_pick_up(function(name)
+	physics.set(name, "ctf_match:flag_mult", { speed = 0.90 })
```
LoneWolfHT, line 84: Why 0.90? Wouldn't '0.9' be exactly the same?

LoneWolfHT: Seems a little weird to have the unneeded 0 but I'll resolve this...

ClobberXD: Nvm, removed the 0 while making another change.

PR #602, opened 2020-04-30
Fix crash when setting HUD

LoneWolfHT:
We were passing a string to `number` when using `hud_add`. This causes a crash in Minetest 5.3-dev with latest master.

### DIFF
--- a/mods/ctf/ctf_colors/hud.lua
+++ b/mods/ctf/ctf_colors/hud.lua
@@ -74,13 +74,13 @@ function ctf_colors.update(player, name, tplayer)
 				position      = {x = 1, y = 0},
 				scale         = {x = 100, y = 100},
 				text          = "Team " .. tplayer.team,
-				number        = tcolor.hex,
+				number        = tonumber(tcolor.hex),
 				offset        = {x = -20, y = 20},
 				alignment     = {x = -1, y = 0}
 			})
 		else
 			ctf.hud:change(player, "ctf:hud_team", "text", "Team " .. tplayer.team)
-			ctf.hud:change(player, "ctf:hud_team", "number", tcolor.hex)
+			ctf.hud:change(player, "ctf:hud_team", "number", tonumber(tcolor.hex))
 		end
 	end
 end
--- a/mods/ctf/ctf_colors/hud.lua
+++ b/mods/ctf/ctf_colors/hud.lua
@@ -9,7 +9,7 @@ function ctf_colors.get_color(tplayer)
 		tcolor_hex = 0x000000
 	end
 
-	local tcolor_css = "#" .. tcolor_hex:sub(3, 8)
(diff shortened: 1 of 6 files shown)


mods/pvp/hud_score/init.lua
```diff
@@ -32,6 +32,8 @@ local function update(name)
 			text = "+ " .. text
 		end
 
+		text = tostring(text)
```
ClobberXD, line 35: I'd prefer this line to be combined with L30. i.e.

```lua
local text = tostring(def.value)
```

LoneWolfHT: Done

mods/other/hudkit/init.lua
```diff
@@ -11,11 +11,6 @@ function hudkit()
 				elements = self.players[name]
 			end
 
-			if def.number and type(def.number) ~= "number" then
-				ctf.warning("hudkit", "type of def.number is "..dump(type(def.number))..". Should be a number")
-				def.number = tonumber(def.number)
-			end
-
```
ClobberXD: Why has this whole block been removed? Isn't it better to let `def.number = tonumber(def.number)` exist?

ClobberXD: Or rather,
```lua
def.number = def.number and tonumber(def.number)
```

(calls `tonumber` only if `def.number` isn't `nil`)

LoneWolfHT: I figured it'd be better to let future PR authors fix the crash caused by bad values instead of requiring me to add a type check for all possible def values. The warning can be easy to miss sometimes

ClobberXD:
I'm not sure if adding `tonumber` to all `tcolor.hex` is the right solution here. Currently, `ctf_colors.get_color().hex` is only used as a number everywhere (`grep -nrF .hex mods/`). It's probably better to make `ctf_colors.get_color` return `hex` as a number instead - this PR would be much smaller then.

mods/other/hudkit/init.lua
ClobberXD: You have a point. Makes sense to remove this warning. :+1:

mods/ctf/ctf_colors/hud.lua
```diff
@@ -6,14 +6,14 @@ function ctf_colors.get_color(tplayer)
 	end
 	local tcolor_hex = ctf.flag_colors[tcolor_text]
 	if not tcolor_hex then
-		tcolor_hex = "0x000000"
+		tcolor_hex = 0x000000
 	end
 
 	local tcolor_css = "#" .. tcolor_hex:sub(3, 8)
```
ClobberXD, line 12: This line causes a crash, as `tcolor_hex` isn't a string now.

```suggestion
	local tcolor_css = "#" .. tostring(tcolor_hex):sub(3, 8)
```

PR #518, opened 2019-11-27
ctf_stats: Improve performance by storing a rank-to-name mapping

ClobberXD:
### Summary

This PR implements some rank-to-name mapping to fix some long-term performance issues in `ctf_stats`.

### Implementation

- The most important addition is a newly-added table `ctf_stats.ranks`, which serves as a rank-to-name mapping to simplify stat lookups.
- `ctf_stats.players` is left intact, with the exception of a new field in each player's stats - this field `rank` corresponds to the rank of the player, and also the index in `ctf_stats.ranks`.
- A local function `regenerate_ranks` is used to (re-)generate `ctf_stats.ranks`, and update `rank` field in all existing stats. This function is called on regular intervals to keep the ranks up-to-date.

****

Ready for review; lotsa testing needed.

Partially attends to #313 and #320; although somewhat unrelated, closes #183

PR #606, opened 2020-05-01
Give medic points for healing

MinetestSam:
Medic will be awarded 5 points each time they heal a team-mate using bandages

### DIFF
--- a/mods/ctf/ctf_classes/medic.lua
+++ b/mods/ctf/ctf_classes/medic.lua
@@ -80,7 +80,7 @@ minetest.override_item("ctf_bandages:bandage", {
 			local hp = object:get_hp()
 			local percentage = 0.75 --percentage of total HP to be healed
 			local limit = percentage * object:get_properties().hp_max
-			if hp > 0 and hp < limit then 
+			if hp > 0 and hp < limit then
 				local main, match = ctf_stats.player(name)
 				if main and match then
 					local reward = 5
--- a/mods/ctf/ctf_classes/medic.lua
+++ b/mods/ctf/ctf_classes/medic.lua
@@ -78,7 +78,9 @@ minetest.override_item("ctf_bandages:bandage", {
 		local name = user:get_player_name()
 		if ctf.player(pname).team == ctf.player(name).team then
 			local hp = object:get_hp()
-			if hp > 0 and hp < 15 then --15 here is the healing limit
+			local percentage = 0.75 --percentage of total HP to be healed
+			local limit = percentage * object:get_properties().hp_max
+			if hp > 0 and hp < limit then 
 				local main, match = ctf_stats.player(name)
 				if main and match then
 					local reward = 5
--- a/mods/ctf/ctf_classes/medic.lua
+++ b/mods/ctf/ctf_classes/medic.lua
@@ -63,3 +63,36 @@ minetest.register_globalstep(function(delta)
 
 	regen_update()
 end)
+
+local bandage_on_use = minetest.registered_items["ctf_bandages:bandage"].on_use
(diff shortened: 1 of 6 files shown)


MinetestSam:
Shouldn't awarding points be kept in ctf_stats itself?

ClobberXD:
There are a couple of mods (like [this one](https://github.com/MT-CTF/capturetheflag/blob/6238fb2540ef73da4ee157b69824056e81fb8f6c/mods/ctf/ctf_bounties/init.lua#L88)) that modify stats outside of `ctf_stats`. The problem with keeping all score-awarding code in one place is that the one place (aka `ctf_stats`) will have to depend on a whole lot of mods to work, greatly increasing the possibility of cyclic dependencies.

I'm planning to add a proper API to `ctf_stats` after #518, but it's still usable from other mods in it's current state. Ideally, the most basic kill score awarding would be the only score-awarding that would take place within the mod, and everything else would be outside of it.

mods/ctf/ctf_classes/medic.lua
```diff
@@ -63,3 +63,36 @@ minetest.register_globalstep(function(delta)
...
+		end
+		local object = pointed_thing.ref
+		if not object:is_player() then
+			return
+		end
+		local pname = object:get_player_name()
+		local name = user:get_player_name()
+		if ctf.player(pname).team == ctf.player(name).team then
+			local hp = object:get_hp()
+			if hp > 0 and hp < 15 then --15 here is the healing limit
```
ClobberXD, line 81: This line caught my attention. This is the reason why knight are only healable upto half their max_hp. I suggest changing 15 into a percentage of hp_max instead. e.g.

```lua
local limit = ctf_bandages.healing_perc * object:get_properties().hp_max
if hp > 0 and hp < limit then
```

where `ctf_bandages.healing_perc` (sorry, can't think of a better name atm) is a constant defined in `ctf_bandages` mod. To preserve current behaviour, this constant should be equal to 0.75.

mods/ctf/ctf_classes/medic.lua
```diff
@@ -78,7 +78,9 @@ minetest.override_item("ctf_bandages:bandage", {
 		local name = user:get_player_name()
 		if ctf.player(pname).team == ctf.player(name).team then
 			local hp = object:get_hp()
-			if hp > 0 and hp < 15 then --15 here is the healing limit
+			local percentage = 0.75 --percentage of total HP to be healed
```
ClobberXD, line 81: As aforementioned, I'd prefer this variable to be defined as a variable in the `ctf_bandages` namespace. The logic is that this is a configuration variable used only by the bandage item's `on_use`. But since the callback is being overridden here, the variable needs to be within the mod's global namespace to be accessible from other mods.

mods/ctf/ctf_classes/medic.lua
```diff
@@ -63,3 +63,37 @@ minetest.register_globalstep(function(delta)
...
+		end
+		local object = pointed_thing.ref
+		if not object:is_player() then
+			return
+		end
+		local pname = object:get_player_name()
+		local name = user:get_player_name()
+		if ctf.player(pname).team == ctf.player(name).team then
+			local hp = object:get_hp()
+			local limit = ctf_bandages.heal_percent * object:get_properties().hp_max
```
Panquesito7, line 81: `ctf_bandages` should be added in `.luacheckrc` for this file.

ClobberXD: Done

ClobberXD:
@MinetestSam Would you be able to rebase? Merge conflict seems to be pretty trivial. I'll try to review this PR today.

PR #609, opened 2020-05-02
Add percentage based healing to bandages

MinetestSam:
This needs to be merged before #606

### DIFF
--- a/mods/ctf/ctf_bandages/init.lua
+++ b/mods/ctf/ctf_bandages/init.lua
@@ -1,9 +1,7 @@
 --Inspired from Andrey's bandages mod
 
-local healing_limit = 15
-
 minetest.register_craftitem("ctf_bandages:bandage", {
-	description = "Bandage\n\nHeals teammates for 3-4 HP until HP is equal to "..healing_limit,
+	description = "Bandage\n\nHeals teammates for 3-4 HP until HP is equal to 75% of the total HP",
 	inventory_image = "ctf_bandages_bandage.png",
 	on_use = function(itemstack, player, pointed_thing)
 		if pointed_thing.type ~= "object" then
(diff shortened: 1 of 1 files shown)


ClobberXD:
Oh you've done this in a separate PR, lol. :)

But [as I mentioned](https://github.com/MT-CTF/capturetheflag/pull/606#discussion_r418923244) in #606:

> ... I'd prefer this variable to be defined as a variable in the `ctf_bandages` namespace. The logic is that this is a configuration variable used only by the bandage item's `on_use`. But since the callback is being overridden here, the variable needs to be within the mod's global namespace to be accessible from other mods.

mods/ctf/ctf_bandages/init.lua
```diff
@@ -1,9 +1,9 @@
 --Inspired from Andrey's bandages mod
 
-local healing_limit = 15
+heal_percent = 0.75 --Percentage of total HP to be healed
```
ClobberXD, line 3: Apologies if I'm being too pedantic. But it's a bad idea to expose a mod's internal variable as a global. I suggest adding this to the mod's namespace, as I earlier suggested.

```suggestion
ctf_bandages = {}
ctf_bandages.heal_percent = 0.75
```

mods/ctf/ctf_bandages/init.lua
```diff
@@ -1,9 +1,10 @@
 --Inspired from Andrey's bandages mod
 
-local healing_limit = 15
+ctf_bandages = {}
+ctf_bandages.heal_percent = 0.75 --Percentage of total HP to be healed
 
 minetest.register_craftitem("ctf_bandages:bandage", {
-	description = "Bandage\n\nHeals teammates for 3-4 HP until HP is equal to "..healing_limit,
+	description = "Bandage\n\nHeals teammates for 3-4 HP until HP is equal to 75% of the total HP",
```
Thomas--S, line 7: The "75%" should be generated from `ctf_bandages.heal_percent`, like this (untested):

```lua
"Bandage\n\nHeals teammates for 3-4 HP until HP is equal to " .. ctf_bandages.heal_percent * 100 .. "% of the total HP"
```

ClobberXD: Thought it wasn't necessary at the moment. But I agree, it's better to not hardcode stuff.

PR #615, opened 2020-05-06
Improve Grenades API

LoneWolfHT:
This will allow for things like the rocketeer class launching grenades

## How to Test
* Join game
* Run `//lua grenades.throw_grenade("grenades_frag", 20, minetest.get_player_by_name("<Yourname>")`

### DIFF
--- a/mods/pvp/grenades/init.lua
+++ b/mods/pvp/grenades/init.lua
@@ -49,7 +49,7 @@ function grenades.register_grenade(name, def)
 			if not vector.equals(self.last_vel, vel) and vector.distance(self.last_vel, vel) > 4 then
 				if def.on_collide and def.on_collide(obj, self.thrower_name) then
 					if self.thrower_name then
-						minetest.log("[Grenades] A grenade thrown by " .. self.thrower_name ..
+						minetest.log("action", "[Grenades] A grenade thrown by " .. self.thrower_name ..
 						" explodes at " .. minetest.pos_to_string(vector.round(pos)))
 						def.on_explode(pos, self.thrower_name)
 					end
@@ -121,7 +121,7 @@ function grenades.register_grenade(name, def)
 
 			if self.timer > def.clock or not self.thrower_name then
 				if self.thrower_name then
-					minetest.log("[Grenades] A grenade thrown by " .. self.thrower_name ..
+					minetest.log("action", "[Grenades] A grenade thrown by " .. self.thrower_name ..
 					" explodes at " .. minetest.pos_to_string(vector.round(pos)))
 					def.on_explode(pos, self.thrower_name)
 				end
--- a/mods/pvp/grenades/init.lua
+++ b/mods/pvp/grenades/init.lua
@@ -21,15 +21,17 @@ function grenades.register_grenade(name, def)
 	end
 
 	local grenade_entity = {
-		physical = true,
+		initial_properties = {
+			physical = true,
+			collide_with_objects = false,
+			visual = "sprite",
+			visual_size = {x = 0.5, y = 0.5, z = 0.5},
+			textures = {def.image},
+			collisionbox = {-0.2, -0.2, -0.2, 0.2, 0.15, 0.2},
+			pointable = false,
+			static_save = false,
+		},
 		sliding = 1,
-		collide_with_objects = false,
-		visual = "sprite",
-		visual_size = {x = 0.5, y = 0.5, z = 0.5},
-		textures = {def.image},
-		collisionbox = {-0.2, -0.2, -0.2, 0.2, 0.15, 0.2},
-		pointable = false,
-		static_save = false,
 		particle = 0,
 		timer = 0,
 		on_step = function(self, dtime)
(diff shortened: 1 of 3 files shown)


LoneWolfHT:
Actually don't merge until I give the OK. Test isn't working...

LoneWolfHT:
Bug has been fixed. This works now

mods/pvp/grenades/init.lua
```diff
@@ -2,15 +2,15 @@ grenades = {
 	grenade_deaccel = 9
 }
 
-local function throw_grenade(name, player)
+function grenades.throw_grenade(name, startvel, player)
```
ClobberXD, line 5: `startvel` is just a velocity multiplier, `startvel` is misleading. Renaming this to something more intuitive would be nice.

LoneWolfHT: Well, `dir` will always be around a value of 1. So that roughly is the starting velocity if I didn't miss something
I should probably change this to `vector.multiply(dir, startvel)` come to think of it

LoneWolfHT:
Some grenade particle improvements incoming, plus a fix

mods/pvp/grenades/init.lua
```diff
@@ -2,15 +2,15 @@ grenades = {
...
 	local obj = minetest.add_entity({x = pos.x + dir.x, y = pos.y + 1.5 + dir.y, z = pos.z + dir.z}, name)
-
-	local m = 20
-	obj:set_velocity({x = dir.x * m, y = dir.y * m, z = dir.z * m})
+	obj:set_velocity({x = dir.x * startvel, y = dir.y * startvel, z = dir.z * startvel})
 	obj:set_acceleration({x = 0, y = -9.8, z = 0})
 
+	obj:get_luaentity().thrower_name = player:get_player_name()
+
 	return(obj:get_luaentity())
```
ClobberXD, line 14: Superfluous parenthesis - we don't use the `return()` syntax.

```suggestion
	return obj:get_luaentity()
```

mods/pvp/grenades/init.lua
ClobberXD: Velocity is a vector; `startvel` is a scalar that's multiplied with an existing vector to calculate velocity. This doesn't make `startvel` a velocity by itself. :)

ClobberXD: > vector.multiply(dir, startvel)

I agree. This would be much more cleaner and readable than manual multiplication. :+1:

LoneWolfHT: Done

mods/pvp/grenades/init.lua
```diff
@@ -46,6 +47,16 @@ function grenades.register_grenade(name, def)
 			-- Check for a collision on the x/y/z axis
 
 			if not vector.equals(self.last_vel, vel) and vector.distance(self.last_vel, vel) > 4 then
+				if def.on_collide and def.on_collide(obj, self.thrower_name) then
+					if self.thrower_name then
+						minetest.log("[Grenades] A grenade thrown by " .. self.thrower_name ..
```
ClobberXD, line 52: Would this work without the log level param?

LoneWolfHT: It defaults to `Action` IIRC

ClobberXD: Just checked the [docs](https://github.com/minetest/minetest/blob/454dbf83a9bf292910c1495a2aa49fd8b960c28f/doc/lua_api.txt#L4237). It seems like the default log level is `none`, which means that this would be logged even if the server was set to log only errors and above. This doesn't seem to be required for something as trivial as this. Please explicitly specify the log level (`action` would be right for this).

LoneWolfHT: Done

ClobberXD:
Ready for review? Let me know when it is.

LoneWolfHT:
Yep, should be

mods/pvp/grenades/init.lua
```diff
@@ -22,7 +23,7 @@ function grenades.register_grenade(name, def)
 	local grenade_entity = {
 		physical = true,
 		sliding = 1,
-		collide_with_objects = true,
+		collide_with_objects = false,
 		visual = "sprite",
 		visual_size = {x = 0.5, y = 0.5, z = 0.5},
 		textures = {def.image},
```
ClobberXD, line 29: The standard object properties need to be placed in `initial_properties`. Directly placing them in the entity def is deprecated.

ClobberXD:
@LoneWolfHT Rebase needed. Merge conflict due to 3aa0008c.

LoneWolfHT:
Done I think

ClobberXD:
Uh, no: https://github.com/MT-CTF/capturetheflag/compare/356f5657d94d3b64a936a77e91009ebdccc7e0d9..60c2febcd96e56ec2116752627fddf0e63ef689c

Stuff from other commits seem to have crept in.

PR #625, opened 2020-05-11
Re-add rocketeer class

LoneWolfHT:
Depends upon #615


mods/ctf/ctf_classes/classes.lua
```diff
@@ -79,29 +79,26 @@ ctf_classes.register("medic", {
 	},
 })
 
---[[ctf_classes.register("rocketeer", {
+ctf_classes.register("rocketeer", {
```
ClobberXD, line 82: The name "Rocketeer" doesn't go well with the new equipment, IMO. A different name might be required, although I can't think of better ones, atm.

ClobberXD: "Bomber"? Or even "Bomberman", lol

mods/ctf/ctf_classes/classes.lua
```diff
@@ -79,29 +79,26 @@ ctf_classes.register("medic", {
...
-			"shooter_rocket:rocket_gun_loaded",
-			"shooter_rocket:rocket 4",
+			"grenade_launcher:launcher_loaded",
+			"grenades:frag 4",
 		},
 
 		additional_item_blacklist = {
-			"shooter_rocket:rocket_gun",
+			"grenade_launcher:launcher_loaded",
+			"grenade_launcher:launcher",
```
ClobberXD, line 95: If `item_blacklist` isn't explicitly defined, the class item blacklist will consist of all the items in `initial_stuff` and `additional_item_blacklist`.

Currently, the loaded launcher and frag grenade will be added to the item_blacklist, which will prevent players from dropping/sharing grenades. I suggest converting this field to `item_blacklist`, which means that *only* the items in this list will be blacklisted.

Or... there's the argument of players obtaining infinite grenades if they aren't blacklisted. In that case, remove the loaded launcher item (it's already a part of blacklist, through `initial_stuff`), and let everything else remain.

mods/pvp/grenade_launcher/init.lua
```diff
@@ -0,0 +1,140 @@
+local reloading = {}
+local RELOAD_INTERVAL = 2
+
+minetest.register_node("grenade_launcher:launcher", {
```
ClobberXD, line 4: Why is this a node? To support meshes?

mods/pvp/grenade_launcher/init.lua
```diff
@@ -0,0 +1,140 @@
...
+		damage_groups = {fleshy = 3},
+	},
+	node_placement_prediction = "",
+	drawtype = "mesh",
+	stack_max = 1,
+	on_place = function(itemstack)
+		return itemstack
+	end,
+	on_use = function(itemstack, user, pointed_thing)
+		grenades.throw_grenade("grenade_launcher:grenade", 30, user)
```
ClobberXD, line 62: Why does the grenade launcher use a separate grenade, and not the existing `grenades:frag` (given that the latter is already provided to the player on (re)spawn)?

ClobberXD:
I like the concept. :+1:

mods/ctf/ctf_classes/classes.lua
LoneWolfHT: Grenadier maybe

mods/pvp/grenade_launcher/init.lua
LoneWolfHT: To other players and your wielditem it will be a big fancy mesh

mods/pvp/grenade_launcher/init.lua
LoneWolfHT: The grenade launcher grenade explodes on contact and shoots a little farther

mods/ctf/ctf_classes/classes.lua
ClobberXD: That sounds good too :+1:

mods/pvp/grenade_launcher/init.lua
ClobberXD: Oh yes, exploding on contact is a must. Thanks for clarifying.

ClobberXD:
Does the player require the special grenades to be present in their inv? I only see the normal grenades in `initial_stuff`.

LoneWolfHT:
Nope, I only have them for the entity part

ClobberXD:
Oh ok, interesting.

mods/ctf/ctf_classes/classes.lua
LoneWolfHT: How's this?

LoneWolfHT: Moving to new PR #631

PR #631, opened 2020-05-14
Re-add rocketeer class

LoneWolfHT:
New PR because i broke the old one

Depends upon #615


mods/pvp/grenade_launcher/init.lua
```diff
@@ -0,0 +1,140 @@
+local reloading = {}
+local RELOAD_INTERVAL = 2
+
+minetest.register_node("grenade_launcher:launcher", {
+	description = "Grenade Launcher",
+	mesh = "grenade_launcher_launcher.obj",
+	tiles = {"grenade_launcher_launcher.png"},
+	inventory_image = "grenade_launcher_launcher_inv.png",
+	tool_capabilities = {
+		full_punch_interval = RELOAD_INTERVAL,
```
Panquesito7, line 10: Why need to use this function?
Wouldn't be better if you just type `2`?

Same in other parts of the code.

ClobberXD: I assume `RELOAD_INTERVAL` is just a named constant. The following are the advantages to using variables for constants:

- The name of the variable makes the purpose of the variable doubly clear.
- In case a constant requires changing, changing one occurrence (the variable def) is better than manually going through all occurrences of the variable and changing it.

samuelstegerKOLLEGAH:
I think u should readd it and remove Sniper class.

mods/pvp/grenade_launcher/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = grenade_launcher
+depends = grenades
```
Panquesito7, line 2: Maybe add a description?

Panquesito7 (approved):
Code looks good. 👍

Avyukt1:
I think we should merge this class with the `sniper` class.

PR #632, opened 2020-05-14
Attempt to fix grapples usable by firing before punching flag

### LINKED ISSUE #600: Players are able to grapple with the flag by shooting the grapple gun before punching flag
TOC != TOU

A better impl. would be checking if the player has the flag right before teleporting.

ClobberXD:
Check for player carrying flag in hook entity's `on_step` as well. Fixes #600, by eliminating the `TOC ~= TOU` issue that's responsible for the bug.

If a player takes the flag while their hook is in the air, the hook is returned to the player and the following chat message is displayed:

`You can't use grapples whilst carrying the flag`

Ready for review.

### Testing instructions

- Go close to the enemy flag.
- Shoot the grapple far away and punch the flag before the grapple lands.
- Observe that the grapple is returned, with the aforementioned message in chat.
- Also verify that the grapple works as expected when you don't have the flag.

### DIFF
--- a/mods/ctf/ctf_classes/ranged.lua
+++ b/mods/ctf/ctf_classes/ranged.lua
@@ -69,3 +69,21 @@ end
 check_grapple("shooter_hook:grapple_gun_loaded")
 check_grapple("shooter_hook:grapple_gun")
 check_grapple("shooter_hook:grapple_hook")
+
+-- Override grappling hook entity to check if player has flag before teleporting
+local old_grapple_step = minetest.registered_entities["shooter_hook:hook"].on_step
+minetest.registered_entities["shooter_hook:hook"].on_step = function(self, dtime, ...)
+	-- Remove entity if player has flag
+	-- This is to prevent players from firing the hook, and then punching the flag
+	if ctf_flag.has_flag(self.user) then
+		local player = minetest.get_player_by_name(self.user)
+		if player then
+			player:get_inventory():add_item("main", "shooter_hook:grapple_hook")
+		end
+		minetest.chat_send_player(self.user,
+			"You can't use grapples whilst carrying the flag")
+		self.object:remove()
+		return
+	end
+	return old_grapple_step(self, dtime, ...)
+end


ClobberXD:
Tested; works. PR is ready for review.

Thomas--S:
If you'd prefer to not duplicate the code, something like this could work (pseudo code):

```lua
old_on_step = registered_entities["hook"].on_step
registered_entities["hook"].on_step = function ()
    if (all kind of funny CTF checks) then
        return false
    end
    return old_on_step()
end
```

ClobberXD:
I for some reason was thinking that wouldn't work. What an idiot I am, of course that's possible and the better option. :)

mods/ctf/ctf_classes/ranged.lua
```diff
@@ -69,3 +69,21 @@ end
...
+		local player = minetest.get_player_by_name(self.user)
+		if player then
+			player:get_inventory():add_item("main", "shooter_hook:grapple_hook")
+		end
+		minetest.chat_send_player(self.user,
+			"You can't use grapples whilst carrying the flag")
+		self.object:remove()
+		return
+	end
+	return old_grapple_step
```
Thomas--S, line 88: Are you sure that this works?

I think there should be paranthesis, like this: `return old_grapple_step(self, dtime)`.

But it is possible that I'm wrong here, I didn't test it.

LoneWolfHT: It should be `old_grapple_step(self, dtime, ...)` and you'll need to add the `...` to line 75 too.
The '...' in this case is all the potential params after `dtime`

ClobberXD: That's rather stupid, who even wrote this dumb code?

ClobberXD: To whoever wrote this stupid code: I've fixed it for you.

ClobberXD:
Fixed `old_grapple_step` not being invoked properly, and fixed the function def so that it overrides in a backwards-compatible manner.

Works as expected. I've re-tested using grapples and grapple guns, both with and without the flag.

LoneWolfHT:
Do we want to convert this to use an on_collide while we're at it? (via grenades mod I guess)

ClobberXD:
> Do we want to convert this to use an on_collide while we're at it? (via grenades mod I guess)

I'd prefer to not do that right now, as we have v2.0.0 to think about. The objective of this PR is to fix the ability for players to abuse grapples, and that objective has been reached.

While I like the idea of using the grenades API for this, it's a complete overhaul of grapples, and as such, I'd like to see that happen after v2.0.

PR #633, opened 2020-05-15
ctf_stats: API improvements

ClobberXD:
- The save-*checking* logic has been separated from the actual save function (`ctf_stats.save`). This would allow mods to directly call `ctf_stats.save` without passing in extra params (like `force_save`).
- New API method `ctf_stats.request_save` - this would allow other mods to set `_needs_save` after modifying player stats. This is temporary, and would be automatically handled when a proper API is in place.

Tested; works. PR is a "no squash", as it contains two different commits.

Partially attends to #579.

### DIFF
--- a/mods/ctf/ctf_bounties/init.lua
+++ b/mods/ctf/ctf_bounties/init.lua
@@ -91,6 +91,14 @@ ctf.register_on_killedplayer(function(victim, killer)
 		match.score = match.score + bounty_score
 		main.bounty_kills = main.bounty_kills + 1
 		match.bounty_kills = match.bounty_kills + 1
+
+		ctf_stats.request_save()
+
+		hud_score.new(killer, {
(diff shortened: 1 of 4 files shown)


ClobberXD:
This can wait until after #518 is merged.

ClobberXD:
Upon second thought, this PR is actually important - currently, there's no way for medics healing-score and `ctf_bounties` to notify `ctf_stats` that something has changed. This results in healing score and bounty-related stats remaining unsaved until something else within `ctf_stats` itself marks the stats as dirty (like kills, captures, etc.). TL;DR: Higher chance of data loss in case of a crash.

`ctf_stats.request_save()` allows other mods to notify `ctf_stats` that a save is required, thereby saving in near-real-time when healing-score or bounties update player stats as well.

PR #638, opened 2020-05-19
ctf_stats: Improve /makepro chat-command

ClobberXD:
This PR tweaks the `/makepro` chat-command to provide more meaningful results. Meaningful, as in whole number kills, instead of something like 4.3 kills. :P

### Changes

- `/makepro` can now be used on other players. It will work, even if the said player doesn't exist. This is intentional, and allows for making a new player a pro, without requiring them to already have a score of 1 or more.
- The return string is more intuitive, and conveys target name, and whether the target is already a pro.
- The value assigned to `kills` (to set the K/D higher than 1.5) is rounded off to the next lowest integer (`math.ceil`).

### To test

- Make note of the stats of an existing player. Let's call them "player1".
- Type `/makepro player1`. This should make them a pro, and return "Made player1 a pro!".
- Type `/makepro player1` again. This time, you should get the "player1 is already a pro!" message.
- Type `/reset_rankings player1` and repeat steps 2 - 3. Verify that you get the same results.
- Repeat steps 2 - 4 on a non-existent player. Verify that you get the same results.
- Repeat steps 2 - 4 on yourself. Verify that you get the same results. Use a different account if you don't want to lose your current stats.

### DIFF
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -178,9 +178,9 @@ minetest.register_chatcommand("transfer_rankings", {
 	privs = {ctf_admin = true},
 	func = function(name, param)
 		if not param then
-			return false, "Invalid syntax. Provide source and destination player names."
+			return false, "Invalid usage, see /help transfer_rankings"
 		end
-		param = param:trim()
+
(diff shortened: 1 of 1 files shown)


ClobberXD:
Note: CI failure is just Travis acting up.

Issue #472, opened 2019-10-12
Break away from MTG completely


ClobberXD:
This idea was inspired by ITB. MTG's `default` is just too much bloat, and updating our nodes after a new MTG release is ever so painful. We've also got a ton of indestructible nodes, created thanks to shameless copy-pasta, resulting in enormous "code" duplication.

If we develop our own nodes instead, we could manage all these issues in a much more elegant manner. Map makers might still be able to make maps from MTG nodes if they want to - we could have some sort of a conversion system available, that'd convert MTG nodes to their CTF counterparts.

Like, share, and subscribe. And feel free to share your thoughts about this proposal in the comments section down below.

Thomas--S:
To be honest, I'm not sure if this is really worth the efforts. Eventually, it will most likely be a duplication of large parts of MTG's `default` anyways.

ClobberXD:
> it will most likely be a duplication of large parts of MTG's default anyways

To a basic extent, yes. But the resulting code would have a *much* better structure, as all nodedefs are in the mod, and the ind node code duplication can be done with for good. That's the main takeaway of this proposal.

Cahrs:
the indestructible nodes are made via copy-pasta? why not just loop through minetest.registered_nodes and register the equivalant with no changed values but groups?

Cahrs:
Id like to say im in favor of this proposal though, this would also maybe open the door for custom deco nodes per map?

ClobberXD:
> the indestructible nodes are made via copy-pasta? why not just loop through minetest.registered_nodes and register the equivalant with no changed values but groups?

There's no guarantee of the map maker having the exact same arsenal of nodes as CTF (the game) does. If they're using a newer version of MTG, for example, they could potentially add nodes in their map that CTF doesn't recognise. While we have to way to check for that, I thought we'd at least avoid creating the same situation for indestructible nodes, as the latter is completely under our control. However, looping through `minetest.registered_nodes` is exactly what we're doing to register indestructible stairs and wool, at the moment.

ClobberXD:
Some more perspective into why breaking away from MTG is a good idea. A non-exhaustive list reasons to break away from MTG:

- More consolidated API:

  - A single mod for nodes instead of fragmentation like wool, xpanes, stairs, etc.
[...]

PR #639, opened 2020-05-20
MTG: Trim out the remaining bloat from default

ClobberXD:
(Continuation of #364?)

This PR removes `legacy.lua` and `mapgen.lua` entirely, and some unused portions in `functions.lua`, in the `default` mod. `legacy.lua` contains legacy (duh) unused code, while `mapgen.lua` contains various ore registrations and mapgen aliases. Since CTF runs solely on singlenode, none of these ore registrations and mapgen aliases are actually used in our code-base.

The only bits of removed code that was actually in use were `default.gui_bg`, `default.gui_bg_img`, and `default.gui_slots`. These had no effect ever since formspec prepends replaced them, and they've been set to `""` ever since. So the fix was to simply remove references to these variables from the mods that use them - namely `tsm_chests`, `ctf_map` (team-chest), and `furnaces`. I've tested all three mods, and their GUIs render as expected.

Needs a little more testing to ensure nothing else is broken - I'm 99% sure about this, but a second pair of eyes would be very helpful in being 100% sure.


ClobberXD:
Trimming out the bloat also helps in understanding exactly what functionalities MTG provides to the game, and would help to better evaluate what needs to be developed from scratch, when MTG is shown the door (#472).

LoneWolfHT (approved):

PR #645, opened 2020-05-22
Prevent putting class items in team chest through swapping

### LINKED ISSUE #576: Can place class items into team chest by swapping

LoneWolfHT:
Fixes #576


ClobberXD:
Tested; works. I doubt this will be fixed in the engine soon enough for us to completely bypass this hack.

@LoneWolfHT It'd be nice if you could apply the same code to treasure chests (`mods/other/tsm_chests`) later, as they suffer from the same issue (#56).

PR #646, opened 2020-05-25
Add configurable respawn delay

Cahrs:
How it works:

When a player dies: store current time in respawns[name].time

When a player respawns: 
    freeze player by attaching them to an invisible immortal object.
    lock player in respawn formspec, telling the user they will respawn in [respawn_time - ((minetest.get_us_time() / 1000000) - respawns[name].time) seconds.
    after duration is over, detach player from and remove object, move them to spawn.

When a player leaves during respawn delay: player is kicked from the server for duration seconds.

### DIFF
--- a/mods/ctf/ctf/teams.lua
+++ b/mods/ctf/ctf/teams.lua
@@ -358,25 +358,64 @@ function ctf.get_spawn(team)
 	end
 end
 
+-- Handle respawning
+
+local respawns = {}
+local respawn_time = 5
+local respawn_formname = "teams:respawn"
+
+local respawn_form = {
+    "size[4,2]",
+	"real_coordinates[true]",
+	"label[1,1;Respawn in " .. respawn_time .. " seconds]",
+}
+respawn_form = table.concat(respawn_form, "")
+
+function handle_respawn(player, name, spawn)
+	minetest.close_formspec(name, respawn_formname)
+	player:move_to(spawn, false)
+
+	respawns[name] = nil
+end
+
 function ctf.move_to_spawn(name)
 	local player = minetest.get_player_by_name(name)
 	local tplayer = ctf.player(name)
 	if ctf.team(tplayer.team) then
 		local spawn = ctf.get_spawn(tplayer.team)
 		if spawn then
-			player:move_to(spawn, false)
+			minetest.after(respawn_time, handle_respawn, player, name, spawn)
 			return true
 		end
 	end
 	return false
 end
 
+minetest.register_on_player_receive_fields(
+	function(player, formname, fields)
+		local name = player:get_player_name()
+		local time = minetest.get_us_time() / 1000000
+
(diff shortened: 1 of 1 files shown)


Cahrs:
Note: ill probably be gone for a couple days so i wont be able to reply, hopefully it can be merged without me making any changes :]

mods/ctf/ctf/teams.lua
```diff
@@ -358,25 +358,64 @@ function ctf.get_spawn(team)
...
-			player:move_to(spawn, false)
+			minetest.after(respawn_time, handle_respawn, player, name, spawn)
 			return true
 		end
 	end
 	return false
 end
 
+minetest.register_on_player_receive_fields(
+	function(player, formname, fields)
```
ClobberXD, line 395: This should be inlined with the prev. line, because:

- consistency in code-style
- one tab less indentation required

mods/ctf/ctf/teams.lua
```diff
@@ -358,25 +358,64 @@ function ctf.get_spawn(team)
...
+		local name = player:get_player_name()
+		local time = minetest.get_us_time() / 1000000
+
+        if formname ~= respawn_formname then
+            return
+        end
```
ClobberXD, line 401: This should be moved right to the top of this callback, so that `name` and `time` are initialised only if the received fields are from the expected formspec.

LoneWolfHT:
> Also, the changes include a cocktail of spaces and tabs - please convert them all into whitespaces.

Ctrl+Shift+P and search for 'convert indentation to tabs' or something similar (@Cahrs)

ClobberXD:
Given the concept of the PR (which I like, btw), there's a lot of testing to do, and a whole lot of things to potentially be fixed - it's going to be pretty time-taking to review, and I'd prefer to wait until v2.0 is released, to begin reviewing this PR.

Cahrs:
All issues mentioned should be fixed now. :tada:
Ive also update the code to freeze the player while waiting for the respawn delay.

Cahrs:
Note: i have tested things to the best of my abilities to make sure they work, but due to how slow running CTF as a local server is for me (during callback runs), the delays in things executing and the server catchup caused by that makes it impossible for me to be completely sure everything will work as it should on a live server: this needs extensive testing.

mods/ctf/ctf/teams.lua
```diff
@@ -362,62 +362,120 @@ end
...
 end)
 
+minetest.register_on_prejoinplayer(function(name, ip)
+	if respawns[name] then
+		local time = respawn_time - ((minetest.get_us_time() / 1000000) - respawns[name].time)
+
+		if time > 0 then
+			return "Kicked: Reason: Leaving the game during respawn timer. You may join again in " .. math.ceil(time) .. " seconds."
+		end
+	end
```
ClobberXD: I suppose this approach works too :)

PR #653, opened 2020-05-31
Nerf bandages heal reward to 1 score/use

### LINKED ISSUE #648: Current points/heal is too much
I'm leaning towards 1-2 points/heal. Feel free to discuss/make a PR.

ClobberXD:
Closes #648

I feel 1 is enough for now. If we really feel that 1 is too low, we can always increase it later.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -12,7 +12,7 @@ git submodule update --init --recursive
 # Run build.sh
 ./build.sh ../games/capturetheflag
 
-# Queue update
+# Queue restart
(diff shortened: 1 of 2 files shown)


update.sh
```diff
@@ -12,7 +12,7 @@ git submodule update --init --recursive
 # Run build.sh
 ./build.sh ../games/capturetheflag
 
-# Queue update
+# Queue restart
```
ClobberXD, line 15: Irrelevant change, but I didn't feel like making a separate commit for this.

PR #655, opened 2020-06-02
Award heal score only to medics

ClobberXD:
I believe this was the original idea - only medics should be awarded score for healing, because that's their job.

Super-trivial PR. Tested; works.


ClobberXD:
`ctf_classes.get(<player or name>)` returns a class definition, which also includes the aptly-named `name` field.

Issue #642, opened 2020-05-20
Use managed HUD elements instead of spammy chat-messages for feedback


ClobberXD:
Use a HUD API resembling the one provided by `hud_scores` (maybe even re-use the same HUD API), to display feedback to player, instead of spammy chat-messages. Examples:

```
<name> is on your team!
<name> has more than <n> HP, you can't heal them!
Feel free to suggest more (no, this isn't a feedback message in-game)
```

The advantage to this is that the player's chat won't be spammed with these messages - they'll be displayed as a HUD element for a couple of seconds and disappear.

xenonca:
`Can't place another marker <x seconds left>`

xenonca:
`You need to wait to change class again`
`You can't explode smoke grenade so close to you flag!`
`Pillar digging started`
`Pillar digging stopped at undiggable noce Can activate again in x seconds`
`You can't activate yet`
`Can't dig node or build time active`
`<player> has x HP. You can't heal them`
`<player> is now a <class>`  

Inventory (Crafting):
`Missing required items`

PR #656, opened 2020-06-03
[no squash] HUD score improvements

ClobberXD:
This PR is half-maintenance, and half-bugfix:

- `hud_score` has been moved from the `pvp` modpack to `other` modpack. This isn't relevant to PvP in anyway, and can be used for displaying any value, really (see also: #642).
- Existing HUD score elements are reused when adding new ones with matching names. This fixes a bug where a HUD score element can be tricked into believing it's a frickin' neon sign.

Tested; works. Could do with some testing with multiple HUD score elements.

### To test

- Run two instances of Minetest in a server-client configuration.
- Launch CTF, make both players join the same team (`/team join <playername> <teamname>` is your friend).
- One player needs to be severely low on health, and the other needs to be a medic.
- Heal the damaged player using the medic, and observe that the heal score is static, and doesn't flicker all over the place (this is what happens in `master`).


LoneWolfHT:
I'll trust your testing and fix any problems that arise

PR #663, opened 2020-06-11
Make stone pickaxe capable of breaking cracky=1 nodes

### LINKED ISSUE #660: Reinforced cobble can be used to trap players
This is because players spawn with stone pickaxes by default, which can't break `cracky=1` nodes.

- Changing reinforced cobble to `cracky=2` results in it being as *soft* as normal stone and cobblestone.
- Making stone pick capable of breaking `cracky=1` nodes allows players to mine many more hardy nodes like ores, [steel|bronze] blocks, etc.
  - I've also tried `cracky=2` + `level=2`, but the `level` group only affects wear, and not dig times.

Any ideas?

Note: This also affects v1, and the fix needs to be backported to it.

ClobberXD:
Tested; works. fixes #660.

Note: Post-merge, this fix needs to be backported to `stable-1` as well.

To test, try breaking `cracky=1` nodes (like reinforced cobble, and mese/diamond ores) using a stone pickaxe - those hardy nodes should take 12s to completely break. Softer nodes like stone and cobblestone should take 4s as before, as this hasn't been changed.

### DIFF
--- a/mods/ctf/ctf_map/ctf_map_core/init.lua
+++ b/mods/ctf/ctf_map/ctf_map_core/init.lua
@@ -26,6 +26,16 @@ function ctf_map.can_cross(player)
 	return false
 end
 
+-- Modify MTG's stone pickaxe to be capable of breaking cracky=1 nodes
+-- Do it here, for lack of a better place
+do
+local tool_caps = minetest.registered_items["default:pick_stone"].tool_capabilities
+tool_caps.groupcaps.cracky.times[1] = 12.0
+minetest.override_item("default:pick_stone", {
+	tool_capabilities = tool_caps
+})
+end
+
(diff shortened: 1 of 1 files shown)


olliy1or:
cant break steel/bronze bloks with stone pickaxe, can break ores but takes like 4secs to breaks ores
and stone and cobble takes like 2seconds -olliy unaproved :c

ClobberXD:
@olliy1or The PR works, but I incorrectly assumed that the stone pickaxe can now break obsidian and steel/bronze blocks. This is incorrect - stone picks can now break reinforced cobble (this is the main reason for this change, see #660), mese/diamond ore, etc. but not the super-hardy nodes like obsidian and steel/bronze blocks.

KaylebJay:
Merge it

PR #666, opened 2020-06-17
Update minetest.conf

Zipitque:
A little more flag time bc i was fighting and the time ran out while in the middle of a fight. :-1:

### DIFF
--- a/minetest.conf
+++ b/minetest.conf
@@ -22,7 +22,7 @@ ctf.flag.capture_take = true
 ctf.flag.waypoints = true
 ctf.flag.nobuild_radius = 3
 ctf.flag.alerts = true
-ctf.flag.drop_time = 300
+ctf.flag.drop_time = 350
 
 ctf.allocate_mode = 4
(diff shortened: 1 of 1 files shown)


ClobberXD:
The devil himself has opened a PR. Welcome to GitHub, little red fella with horns.

****

:-1: I'm against this. 300s is five full minutes. If the flag is dropped before you can capture it, grab it again and capture it. The drop time used to be seven minutes earlier, but we felt that it was too much, and reduced it to five. If you really were fighting for 5 entire minutes without dying, firstly, kudos; secondly, the flag drop time exists for two reasons:

- To prevent match stalling by malicious players who take the flag and keep it forever.
- To add a challenge to the flag-bearer. If it takes you more than five minutes to capture, either you're just stalling or someone else has your flag. If it's the latter, kill the enemy ASAP and capture the flag, or if the flag drops before you can cap, just go back and get the flag again.

Zipitque:
Ah thanks for explaining. and yes i am the devil!

PR #675, opened 2020-09-03
Add automatic skip votes

LoneWolfHT:
If someone could just review the code this should be good to go. I tested this extensively ingame


Panquesito7 (approved):
Code looks good. 👍

PR #684, opened 2020-10-12
Increase sword damage dealt by knights

LoneWolfHT:
Also makes steel sword craftable again

# How to test
* Join to clients into a CTF game
* Punch a knight with a medic and vise versa. Knight should deal 1 extra damage (Easier to see damage difference when punching medic)
* Craft a steel sword as a non-knight class and use it
* Play around fighting the different classes as a knight
* Try to exploit the damage dealing code


mods/ctf/ctf_crafting/init.lua
```diff
@@ -7,6 +7,7 @@ local full_ores = {
 }
 
 local sword_materials = {
+	steel   = "default:steel_ingot",
 	mese    = "default:mese_crystal",
 	diamond = "default:diamond",
```
Panquesito7, line 12:
```suggestion
	diamond = "default:diamond"
```

Panquesito7 (approved):
Code looks good. 👍

mods/ctf/ctf_crafting/init.lua
Thomas--S: I think it's a good practice to use a comma at the end of a table.

This allows to add entries more easily.

LoneWolfHT: Whoops, guess I had forgotten to reply. Yeah, that was the intent with that

LoneWolfHT:
Looks like the Discord hasn't found any bugs

PR #685, opened 2020-10-13
[NO SQUASH] Buff medics and improve initial_stuff

LoneWolfHT:
Will need some good code reviewing for this.

# How To Test

* Make sure physics are still working fine. I allowed for changing more than just class speed
* Make sure you have no issues using cobble.
* Make sure all your items are fine when switching classes.
  * Try dropping some/all but one of a stack of blocks and switching classes
  * Switch maps a few times when testing to get a good variety of initial items
* Make sure tool crafting looks fine

I would have done a second PR for the initial stuff but the medic is good for testing it


LoneWolfHT:
Got 2 tests in. Merging...

PR #686, opened 2020-10-20
Give medic 99 bandages and 3 points per bandage

JostP:


xenonca:
Keep in mind 5 points was a lot and 3 points gives them 3 times what they get now. Maybe a 2 or 2.5 would do?

JostP:
Rebased

PR #689, opened 2020-10-24
Infinite bandages for medics

KaylebJay:
plus only giving them *one* bandage initially, since it will never run out.

I thought I remembered bandages in loot, was going to remove that, but it wasn't there - grepped the code and it wasn't anywhere related to loot, if I missed something there let me know.


KaylebJay:
Note also https://github.com/MT-CTF/maps/pull/136

PR #692, opened 2020-10-27
Make lava insta-kill players

KaylebJay:


mods/mtg/default/nodes.lua
```diff
@@ -1842,7 +1842,7 @@ minetest.register_node("default:lava_source", {
 	liquid_alternative_source = "default:lava_source",
 	liquid_viscosity = 7,
 	liquid_renewable = false,
-	damage_per_second = 4 * 2,
+	damage_per_second = 30,
```
KaylebJay, line 1845: 30, because knights have to be instakilled too.
Removed the multiplication, I have no idea why the person who wrote this did that...

LoneWolfHT: 4 hearts I guess

PR #693, opened 2020-10-27
Add mod `real_suffocation` with custom tweaks for CTF

KaylebJay:
Players now slowly suffocate inside nodes, since they should not be in blocks they cannot obtain, I have left this feature (damage instead of suffocation in blocks they cannot obtain) out for the moment. 

Also, players only have half as much breath as before.

There is no way to damage in water less than `1`, so putting this up here, feel free to suggest changes in regards to making water similar to how it was before.

### DIFF
--- a/mods/other/real_suffocation/README.md
+++ b/mods/other/real_suffocation/README.md
@@ -1,0 +1,24 @@
+# Real Suffocation [`real_suffocation`]
+Version: 1.0.0
+
+This mod adds suffocation. Suffocation is basically the same as drowning, but it
+is for being stuck inside solid blocks. If you're inside a solid block, you lose
+breath. If you lost your breath completely and you're still inside, you suffer
+10 HP (=5 “hearts”) of damage every 2 seconds.
+
+Specifically, suffocation is added to all blocks which:
+
+* Are solid
+* Are full cubes
+* Don't already have built-in damage or drowning damage
+* Are not excluded from suffocations by mods
+
+## Info for modders
+This mod will not add suffocation to all nodes with the group
+`disable_suffocation=1`.
+
+This mod adds the group `real_suffocation=1` to all nodes it has modified,
+this is mostly done for informational purposes.
+
+## License
+Everything is under WTFPL.
--- a/mods/other/real_suffocation/README.md
+++ b/mods/other/real_suffocation/README.md
@@ -21,4 +21,4 @@ This mod adds the group `real_suffocation=1` to all nodes it has modified,
 this is mostly done for informational purposes.
 
 ## License
-Everything is under WTFPL.
+Everything is under the MIT license. Forked from https://repo.or.cz/minetest_real_suffocation.git.
--- a/mods/other/real_suffocation/description.txt
+++ b/mods/other/real_suffocation/description.txt
@@ -1,1 +1,0 @@
-The player will lose breath inside solid blocks.
--- a/mods/other/real_suffocation/description.txt
+++ b/mods/other/real_suffocation/description.txt
@@ -1,0 +1,1 @@
+The player will lose breath inside solid blocks.
--- a/mods/other/real_suffocation/init.lua
+++ b/mods/other/real_suffocation/init.lua
@@ -1,0 +1,56 @@
+-- Load setting
+local suffocation_damage = 2
+local setting = minetest.settings:get("real_suffocation_damage")
+if tonumber(setting) ~= nil then
+	suffocation_damage = tonumber(setting)
+end
+
+-- Skip the rest if suffocation damage is 0, no point in overwriting stuff
+if suffocation_damage > 0 then
+
+-- Checks all nodes and adds suffocation (drowning damage) for suitable nodes
+local function add_suffocation()
+	-- For debugging output
+	local suffocate_nodes = {}
+	local no_suffocate_nodes = {}
+	-- Check ALL the nodes!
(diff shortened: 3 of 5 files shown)


mods/other/real_suffocation/description.txt
```diff
@@ -0,0 +1 @@
+The player will lose breath inside solid blocks.
```
Panquesito7, line 1: Consider using `mod.conf`, `description.txt` is deprecated.

mods/other/real_suffocation/mod.conf
```diff
@@ -0,0 +1 @@
+name = real_suffocation
```
Panquesito7, line 1:
```suggestion
name = real_suffocation
description = The player will lose breath inside solid blocks.
```

mods/other/real_suffocation/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Info for modders
+This mod will not add suffocation to all nodes with the group
+`disable_suffocation=1`.
+
+This mod adds the group `real_suffocation=1` to all nodes it has modified,
+this is mostly done for informational purposes.
+
+## License
+Everything is under WTFPL.
```
Panquesito7, line 24: Would be nice if you could choose another license such as MIT.

KaylebJay:
I copied all this from Wuzzy, but I agree with those changes, let me implement them.

mods/other/real_suffocation/README.md
KaylebJay: I don't think Wuzzy would mind.

Panquesito7: The license allows doing any changes with the code.
_Off-topic: [WTFPL is harmful to software developers](https://cubicspot.blogspot.com/2017/04/wtfpl-is-harmful-to-software-developers.html)._

mods/other/real_suffocation/init.lua
```diff
@@ -0,0 +1,56 @@
+-- Load setting
+local suffocation_damage = 2
+local setting = minetest.settings:get("real_suffocation_damage")
```
Panquesito7, line 3: Better to add the mod name at the beginning.
```suggestion
local setting = minetest.settings:get("real_suffocation.real_suffocation_damage")
```

mods/other/real_suffocation/init.lua
```diff
@@ -0,0 +1,56 @@
+-- Load setting
```
Panquesito7, line 1:
```suggestion
real_suffocation = {}

-- Load setting
```

mods/other/real_suffocation/init.lua
```diff
@@ -0,0 +1,56 @@
...
+local setting = minetest.settings:get("real_suffocation_damage")
+if tonumber(setting) ~= nil then
+	suffocation_damage = tonumber(setting)
+end
+
+-- Skip the rest if suffocation damage is 0, no point in overwriting stuff
+if suffocation_damage > 0 then
+
+-- Checks all nodes and adds suffocation (drowning damage) for suitable nodes
+local function add_suffocation()
```
Panquesito7, line 12:
```suggestion
function real_suffocation.add_suffocation()
```

mods/other/real_suffocation/init.lua
```diff
@@ -0,0 +1,56 @@
...
+			table.insert(no_suffocate_nodes, itemstring)
+		end
+	end
+	minetest.log("info", "[real_suffocation] Suffocation has been hacked into "..#suffocate_nodes.." nodes.")
+	minetest.log("verbose", "[real_suffocation] Nodes with suffocation: "..dump(suffocate_nodes))
+	minetest.log("verbose", "[real_suffocation] Suffocation has not been hacked into "..#no_suffocate_nodes.." nodes: "..dump(no_suffocate_nodes))
+end
+
+-- This is a minor hack to make sure our loop runs after all nodes have been registered
+minetest.after(0, add_suffocation)
```
Panquesito7, line 50:
```suggestion
minetest.after(0, real_suffocation.add_suffocation)
```

mods/other/real_suffocation/settingtypes.txt
```diff
@@ -0,0 +1,4 @@
+# How much damage is caused by “suffocation” by default, that is, running
+# out of breath while being inside solid blocks like dirt, gravel, sand,
+# etc. Like for drowning, damage is caused every 2 seconds.
+real_suffocation_damage (Suffocation damage) int 2 0
```
Panquesito7, line 4:
```suggestion
real_suffocation.real_suffocation_damage (Suffocation damage) int 2 0
```

mods/other/real_suffocation/init.lua
KaylebJay: I don't think this is really needed, @LoneWolfHT?

KaylebJay: e.g. turning the functions from local functions into table functions.

KaylebJay: I'm going to leave it as much Wuzzy's as I can, so although this is simple, we shouldn't need to override any of these functions (fix in source, instead) and thus using local functions is fine with me.
We use a lot of local functions in ctf, as example.

Panquesito7 (approved):
Tested; works. 👍

olliy1or:
drowning and suffocating both works just fine

PR #695, opened 2020-10-27
Prevent medics from healing players damaged by drowning/lava

LoneWolfHT:
# How to Test
* Join the game as a knight and medic
* Have the non-medic jump into lava and out again once they take some damage
* Try to heal them
* Do the same again but use water to deal the damage

### DIFF
--- a/mods/ctf/ctf_classes/medic.lua
+++ b/mods/ctf/ctf_classes/medic.lua
@@ -64,6 +64,21 @@ minetest.register_globalstep(function(delta)
 	regen_update()
 end)
 
+local dont_heal = {}
+minetest.register_on_player_hpchange(function(player, hp_change, reason)
+	local name = player:get_player_name()
+
+	if reason.type == "drown" or reason.type == "mode_damage" then
+		dont_heal[name] = true
+	elseif dont_heal[name] then
+		dont_heal[name] = nil
+	end
+end)
+
+minetest.register_on_leaveplayer(function(player)
+	dont_heal[player:get_player_name()] = nil
+end)
+
 local bandage_on_use = minetest.registered_items["ctf_bandages:bandage"].on_use
 minetest.override_item("ctf_bandages:bandage", {
 	on_use = function(stack, user, pointed_thing)
@@ -91,7 +106,11 @@ minetest.override_item("ctf_bandages:bandage", {
 
 				if main and match then
 					local reward = 3
+
 					if ctf_flag.has_flag(pname) then reward = 6 end
+
+					if dont_heal(pname) then reward = 0 end
+
 					main.score  = main.score  + reward
 					match.score = match.score + reward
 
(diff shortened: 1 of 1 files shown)


Panquesito7 (commented):
Tested (with MT 5.3.0); received this error while healing someone drowning in water:

Minetest error

mods/ctf/ctf_classes/medic.lua
```diff
@@ -91,7 +106,11 @@ minetest.override_item("ctf_bandages:bandage", {
 
 				if main and match then
 					local reward = 3
+
 					if ctf_flag.has_flag(pname) then reward = 6 end
+
+					if dont_heal(pname) then reward = 0 end
```
Panquesito7, line 112: This shall fix the error above.
```suggestion
					if dont_heal[pname] then reward = 0 end
```

KaylebJay: :thumbsup:

LoneWolfHT: dang

Panquesito7 (approved):
Tested; works. 👍

mods/ctf/ctf_classes/medic.lua
```diff
@@ -91,7 +106,11 @@ minetest.override_item("ctf_bandages:bandage", {
 
 				if main and match then
 					local reward = 3
+
 					if ctf_flag.has_flag(pname) then reward = 6 end
+
+					if dont_heal[pname] then reward = 0 end
```
Thomas--S, line 112: I guess that the two if-statements above should be swapped. Otherwise, you'd still be rewarded if you tried to heal a drowning flag bearer.

KaylebJay: Yes, this change must happen for it to work properly.

LoneWolfHT: It doesn't award a flag bearer any differently right now does it?

olliy1or:
in water you can heal them but sometimes you get 0 points but when they are in lava you seem to be able to spam heals and get points too, in both ways i can spam heal tho

LoneWolfHT:
Note that this new update will make bandages not work in water

mods/ctf/ctf_classes/medic.lua
```diff
@@ -79,8 +94,9 @@ minetest.override_item("ctf_bandages:bandage", {
 		local pname = object:get_player_name()
 		local name = user:get_player_name()
 		if ctf.player(pname).team == ctf.player(name).team then
-			if minetest.get_node(object:get_pos()).name:find("lava") then
-				return -- Can't heal players in lava
+			local nodename = minetest.get_node(object:get_pos()).name
+			if dont_heal[pname] or nodename:find("lava") or nodename:find("water") then
```
Panquesito7, line 98: What if it's flowing lava/water?
It could be abused that way.

LoneWolfHT: I use 'find' because that way it doesn't care if it's a source node or a flowing node

PR #700, opened 2020-10-31
Give capturers 10% of the match.score or 50 points

KaylebJay:
Have not tested - this seemed trivial. Tester's will need to review.


KaylebJay:
This is not functional currently, needs revision

PR #704, opened 2020-11-21
Fix anti coward

appgurueu:
Disable anti coward during building time


LoneWolfHT:
Fixed this via `ctf_match.is_in_build_time()`

PR #710, opened 2020-11-30
Add respawn_timer

appgurueu:
Uses a submodule. Minor inconvienences might need to be addressed (health bar & nametag).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,9 @@
 [submodule "mods/pvp/respawn_timer"]
 	path = mods/pvp/respawn_timer
 	url = https://github.com/appgurueu/respawn_timer.git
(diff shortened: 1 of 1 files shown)


Panquesito7 (commented):
This mod depends on the `modlib` mod, and `modlib` is not present here.

appgurueu:
> This mod depends on the `modlib` mod, and `modlib` is not present here.

Indeed

LoneWolfHT:
This would be better done as a quick, fast, and simple mod that is intended for CTF only

LoneWolfHT:
I've done what I suggested. I'll push it once my computer has internet.
Sorry for closing but there's a bit too much in this. Despite that I have used some of your ideas in my version. So this wasn't a complete waste :P

PR #712, opened 2020-12-05
Add CTF Respawn Delay

LoneWolfHT:
Closes #646

## How to test
* Kill and be killed with multiple clients
* Make sure everything that normally happens after death/respawn will still happen
* See how the 5 second delay feels. Does it need tweaking?


KaylebJay:
https://github.com/MT-CTF/capturetheflag/pull/646 can be closed if this does the same thing

LoneWolfHT:
Ah thanks, edited

PR #715, opened 2020-12-09
Colorize "flag regarding" messages

JostP:
Old:
Zaslonska slika 2020-12-09 11-27-58

New:
Zaslonska slika 2020-12-09 09-54-32

### DIFF
--- a/mods/ctf/ctf_flag/api.lua
+++ b/mods/ctf/ctf_flag/api.lua
@@ -83,11 +83,12 @@ function ctf_flag.player_drop_flag(name)
 
 			local flag_return_color = "#ffffff"
 			if flag_name == "red's flag" then
-			flag_return_color = "#ff4444"
+				flag_return_color = ctf_colors.colors["red"]
 			end
 			if flag_name == "blue's flag" then
-				flag_return_color = "#4466ff"
+				flag_return_color = ctf_colors.colors["blue"]
 			end
+			flag_return_color = "#" .. tostring(flag_return_color):sub(3, 8)
 
 			minetest.chat_send_all(minetest.colorize(flag_return_color, (flag_name.." has returned.")))
 
--- a/mods/ctf/ctf_flag/api.lua
+++ b/mods/ctf/ctf_flag/api.lua
@@ -80,7 +80,16 @@ function ctf_flag.player_drop_flag(name)
 			ctf.hud.updateAll()
 
 			ctf.action("flag", name .. " dropped " .. flag_name)
-			minetest.chat_send_all(flag_name.." has returned.")
+
+			local flag_return_color = "#ffffff"
+			if flag_name == "red's flag" then
+			flag_return_color = "#ff4444"
+			end
+			if flag_name == "blue's flag" then
+				flag_return_color = "#4466ff"
+			end
+
+			minetest.chat_send_all(minetest.colorize(flag_return_color, (flag_name.." has returned.")))
 
 			for j = 1, #ctf_flag.registered_on_drop do
 				ctf_flag.registered_on_drop[j](name, flag)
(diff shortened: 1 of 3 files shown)


Thomas--S:
I'm not sure, but I think that CTF (at least the engine) also supports additional colors. Maybe a more abstract implementation that uses the colors from the ctf_pvp_engine would be possible?

JostP:
> I'm not sure, but I think that CTF (at least the engine) also supports additional colors. Maybe a more abstract implementation that uses the colors from the ctf_pvp_engine would be possible?

Maybe you mean this file? I got colors for text here. But I am not sure that I am that expert to understand what I should change.

LoneWolfHT:
`red` is probably an unknown variable. Should make that a string

JostP:
> `red` is probably an unknown variable. Should make that a string

I tried with:

```
flag_return_color = ctf_colors.colors(red)
flag_return_color = ctf_colors.colors("red")
flag_return_color = ctf_colors(red)
flag_return_color = ctf_colors("red")
flag_return_color = ctf_colors.colors[1]
flag_return_color = ctf_colors.colors["1"]
```

and I have been getting crashes every time. Current solution works and I see no need to complicate it.

Thomas--S:
`ctf_colors.colors["red"]` and `ctf_colors.colors.red` should work.

> Current solution works and I see no need to complicate it.

Imagine you want to change the exact color. You'd have to find every place where it is used if you don't use the more abstract solution I proposed.

Thanks for your efforts!

JostP:
> `ctf_colors.colors["red"]` and `ctf_colors.colors.red` should work.
> 

Well I don' get a crash, but I can't say that it works. I tried with your both suggestions for `pick flag` message, capture flag message isn't changed.

Zaslonska slika 2020-12-10 16-02-01

Thomas--S:
It's not working because the color in `ctf_colors` is with the prefix `0x` and Minetest expects the prefix `#` here.

Would this function help you? https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_colors/hud.lua#L1

i.e. `ctf_colors.get_color("<playername>").css`

JostP:
> It's not working because the color in `ctf_colors` is with the prefix `0x` and Minetest expects the prefix `#` here.
> 
> Would this function help you? https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_colors/hud.lua#L1
> 
> i.e. `ctf_colors.get_color("<playername>").css`

I am not really sure what should I use instead `"<playername>"`.
I tried with:
```
ctf_colors.get_color(attacker.team).css
ctf_colors.get_color(ctf.team(attacker.team)).css
ctf_colors.get_color(ctf.team).css
ctf_colors.get_color(ctf.team(tplayer.team)).css
```
But I get a crash or white, not-colorized  message.

Thomas--S:
Basically, you have three options here:

1. Check if you have access to the player name as string instead of the team or
2. use the color as shown earlier by directly acessing the table (instead of calling the function). You can then replace "0x" with "#" or
3. write a new function `ctf_colors.get_team_color(<team name>)` for the ctf_colors mod.

JostP:
Done, if works perfectly now :))
This is what I get when I change colors in `ctf_colors.colors`:
Zaslonska slika 2020-12-12 11-33-14

Can be reviewed and merged now.

LoneWolfHT:
`red` and `blue` in those messages should be colored accordingly IMO

JostP:
> accordingly

What do you mean?

LoneWolfHT:
`red` colored red. `blue` colored blue, etc

JostP:
> `red` colored red. `blue` colored blue, etc

It already is, check file changes. Screenshot above with orange color for red and purple color for blue is just to show that colors change if you change color hex numbers in ctf_colors.colors (https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf/ctf_colors/init.lua).

LoneWolfHT (approved):

PR #716, opened 2020-12-10
Prevent sharp shooter to use the shotgun!

AnyBlue:
Prevent sharp shooters to use shotgun


Calinou:
Superseded by https://github.com/MT-CTF/capturetheflag/pull/717.

Issue #268, opened 2018-11-25
Make bounties more scalable


ClobberXD:
Bounty prizes are calculated using the following algorithm: `(score * KD) / 5000`. To prevent extreme bounty prizes, there's a hard upper and lower limit of 500 and 50 respectively. What happens here is that unless a player's score and KD are within a certain magical range, the bounty on their heads is always 50 or 500, because the algorithm generally returns a number outside the permitted range.

It would be great if the bounty was somehow made to scale within the 50-500 range in the first place. The lower limit should also be reduced to 5 or 10.

AKryukov92:
Is it possible to attach actual player rankings to this issue?
We just need other formula to scale bounties more accurately.
To develop such formula, we need some real data to test  hypotheses.

bewlay:
What about only tying it to K/D? With something like `(kd  - 1) * 100` It would start as 50 at k/d 1.5 and max out as 500 at k/d 6 (maybe higher?). I think most players have a k/d in between that, the linear scaling would probably be fine.

JostP:
What about  `(score/deaths) * 2` (average points per death * 2)? 
You can find score/death ratio here: https://forum.minetest.net/viewtopic.php?f=10&t=13157&p=337571#p337521

xenonca:
Can close I think?

PR #720, opened 2020-12-14
Make bounties more scalable

JostP:
Issue #268 

> Bounty prizes are calculated using the following algorithm: (score * KD) / 5000. To prevent extreme bounty prizes, there's a hard upper and lower limit of 500 and 50 respectively. What happens here is that unless a player's score and KD are within a certain magical range, the bounty on their heads is always 50 or 500, because the algorithm generally returns a number outside the permitted range.
> 
> It would be great if the bounty was somehow made to scale within the 50-500 range in the first place. The lower limit should also be reduced to 5 or 10.

I changed to `(score * 2)/deaths` (average points per death * 2) so is is a lot more scalable.
Zaslonska slika 2020-12-14 08-49-30
Zaslonska slika 2020-12-14 08-49-45

Bounties for top 10 players:

Zaslonska slika 2020-12-14 09-14-57

### DIFF
--- a/mods/ctf/ctf_bounties/init.lua
+++ b/mods/ctf/ctf_bounties/init.lua
@@ -23,6 +23,20 @@ local function bounty_player(target)
 	local prev = bountied_player
 	bountied_player = target
 
+	-- Players with less than 20000 points --> bounty 50
+	--
+	-- ########################################################
+	--
+	-- Players between 20000 points and 50000 points:
+	--
+	--                  Score
+	-- bounty_score = ----------, or 500 (whichever is lesser)
+	--                  deaths
+	--
+	-- ########################################################
+	--
+	-- Players with more than 50000 points:
+	--
 	--                 Score * 2
 	-- bounty_score = -----------, or 500 (whichever is lesser)
 	--                  deaths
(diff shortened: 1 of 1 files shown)


JostP:
Updated `give players with less than 50k pts lower bounty`

mods/ctf/ctf_bounties/init.lua
```diff
@@ -23,15 +23,39 @@ local function bounty_player(target)
...
 	local pstat = ctf_stats.player(target)
 	if pstat.deaths == 0 then
 		pstat.deaths = 1
 	end
-	bounty_score = (pstat.score * (pstat.kills / pstat.deaths)) / 10000
+
+	if pstat.score <= 20000 then
+		bounty_score = 50
+	end
+	if ((pstat.score > 20000) and (pstat.score <= 50000)) then
```
LoneWolfHT, line 52: This could be

```lua
if score <= 20000 then
    ...
elseif score > 50000 then
    ...
else -- score can only be > 20000 and <= 50000
    ...
end
```
Unless I'm thinking about it wrong that'll speed things up a tiny bit due to less calculations

JostP: :+1: true

JostP: I just didn't know how else if is used in lua

JostP: Updated, I haven't tested wit elseif, but I suppose it works.

mods/ctf/ctf_bounties/init.lua
```diff
@@ -23,15 +23,37 @@ local function bounty_player(target)
...
 
 	local pstat = ctf_stats.player(target)
 	if pstat.deaths == 0 then
 		pstat.deaths = 1
 	end
-	bounty_score = (pstat.score * (pstat.kills / pstat.deaths)) / 10000
+
+	if pstat.score <= 20000 then
+		bounty_score = 50
+	elseif ((pstat.score > 20000) and (pstat.score <= 50000)) then
```
LoneWolfHT, line 51: This elseif will only be run if `pstat.score > 20000`. So all you need is the 50000 comparison

LoneWolfHT (changes requested):

mods/ctf/ctf_bounties/init.lua
JostP: ok, fixed

Issue #402, opened 2019-04-07
Clarify why medkit healing was interrupted in the message


Calinou:
When I first used medkits, it wasn't clear to me why healing was interrupted every so often. It turns out one can't move, attack or be attacked while using a medkit, otherwise the healing will be reverted.

The message that appears when healing is interrupted should make this clear, stating the reason why it was interrupted. It could also appear in the middle of the screen as to be more visible.

ClobberXD:
Appearing in the middle of the screen for better visibility is a good idea.

Emojigit:
Is this OK?
```
diff --git a/mods/ctf/ctf_match/buildtime.lua b/mods/ctf/ctf_match/buildtime.lua
index c902df4..5caead3 100644
--- a/mods/ctf/ctf_match/buildtime.lua
+++ b/mods/ctf/ctf_match/buildtime.lua
@@ -64,7 +64,9 @@ end)
 
 minetest.register_on_punchplayer(function(_, hitter)
 	if ctf_match.is_in_build_time() then
-		minetest.chat_send_player(hitter:get_player_name(), "Match hasn't started yet!")
+		if hitter:is_player() then
+			minetest.chat_send_player(hitter:get_player_name(), "Match hasn't started yet!")
+		end
 		return true
 	end
 end)
diff --git a/mods/pvp/medkits/init.lua b/mods/pvp/medkits/init.lua
index 40ad7d3..40cf29f 100644
--- a/mods/pvp/medkits/init.lua
+++ b/mods/pvp/medkits/init.lua
@@ -46,6 +46,23 @@ local function start_healing(stack, player)
 	return stack
 end
 
+
+-- Interrupt reason can be:
+-- damage: get damage from another player
+-- move: move more then 1m
+-- die: die because any reason
+local function reason_handler(reason)
+	if reason == "damage" then
+		return " because someone damaged you!"
+	elseif reason == "move" then
+		return " because you moved!"
+	elseif reason == "die" then
+		return "because you died!"
+	else
+		return "!"
+	end
+end
+
 -- Called after regen is complete. Remove additional effects
 -- If interrupted == true, revert to original HP and give back one medkit.
 local function stop_healing(player, interrupted)
@@ -55,7 +72,7 @@ local function stop_healing(player, interrupted)
 	players[name] = nil
 	if interrupted then
 		minetest.chat_send_player(name, minetest.colorize("#FF4444",
-			"Your healing was interrupted!"))
[...]

PR #729, opened 2021-01-03
Add medkit interrupt reason

Emojigit:
Close https://github.com/MT-CTF/capturetheflag/issues/402
Add medkit interrupt reason, also have a minor bugfix.


LoneWolfHT (approved):

Emojigit:
There is a minor bug. https://github.com/MT-CTF/capturetheflag/pull/732

PR #733, opened 2021-01-05
Delete Old / Useless Files

Avyukt1:
The `sniper` class was removed, some files still are taking space, to reduce the size of the game, we should remove the old files.


TSafa-23:
I do believe that the reason these skins remain is because there is a possible future re-implementation of the sniper class

Avyukt1:
Okay, then I should close this PR.

LoneWolfHT:
The sniper skin was borrowed from the rocketeer skin

PR #736, opened 2021-01-05
Add game description in game.conf

Avyukt1:


LoneWolfHT:
This should go in the readme.md unless description.txt has some special significance?

appgurueu:
`description.txt` should not be used, use `game.conf` instead

Avyukt1:
> `description.txt` should not be used, use `game.conf` instead

That doesn't work in MT. The `game.conf` only works in the ContentDB

Panquesito7:
> `game.conf` only works in ContentDB

Nope. `game.conf` designed for both ContentDB and Minetest.
Also note that `description.txt` is deprecated:

- For games, use `game.conf` instead.
- For mods, use `mod.conf` instead.

So the description should be added in `game.conf` as @appgurueu said.

description.txt
```diff
@@ -0,0 +1 @@
+Use swords, guns, and grenades to combat the enemy and capture their flag before they capture yours.
```
Panquesito7, line 1: I guess `description.txt` should be removed...pretty much nobody uses Minetest 0.4.x series now.

Panquesito7 (commented):
PR title shall be changed to `Add game description in game.conf`.

PR #737, opened 2021-01-05
Add "Capture Rate" statistic

realOneplustwo:
Statistic to show how good a player is at capturing the flag.

### DIFF
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -36,6 +36,7 @@ local function return_as_chat_result(to, target)
 
 	if stat then
 		local kd = stat.kills
+		local cr = stat.captures / stat.attempts
 		if stat.deaths > 1 then
 			kd = kd / stat.deaths
 		end
@@ -46,6 +47,7 @@ local function return_as_chat_result(to, target)
 			"\nBounty kills: " .. stat.bounty_kills ..
 			" | Captures: " .. stat.captures ..
 			" | Attempts: " .. stat.attempts ..
+			"\nCapture rate: " .. math.floor(cr * 100) .. "%" ..
 			"\nScore: " .. math.floor(stat.score)
 	end
 	return result
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -42,22 +42,22 @@ local function return_as_chat_result(to, target)
 			kd = kd / stat.deaths
 		end
 		result = result ..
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_stats/chat.lua
```diff
@@ -46,6 +47,7 @@ local function return_as_chat_result(to, target)
 			"\nBounty kills: " .. stat.bounty_kills ..
 			" | Captures: " .. stat.captures ..
 			" | Attempts: " .. stat.attempts ..
+			"\nCapture rate: " .. math.floor(cr * 100) .. "%" ..
```
Panquesito7, line 50: Shouldn't this be like

```diff
- 			"\nCapture rate: " .. math.floor(cr * 100) .. "%" ..
+ 			"\nCapture rate: %" .. math.floor(cr * 100)
```

realOneplustwo (commented):
Should I keep it at math.floor or should I change it to math.round?

mods/ctf/ctf_stats/chat.lua
realOneplustwo: That would make it look like "Capture rate: %100", instead of "Capture rate: 100%", which is what I am going for.

Panquesito7: OH yeah...thought my suggestion was the other way round. 😄

appgurueu:
> Should I keep it at math.floor or should I change it to math.round?

Lua doesn't have `math.round`, use `math.floor(number + 0.5)` if you want it.

Panquesito7:
Can someone re-run the LuaCheck checks?

mods/ctf/ctf_stats/chat.lua
```diff
@@ -37,23 +37,26 @@ local function return_as_chat_result(to, target)
 
 	if stat then
 		local kd = stat.kills
+		local cr = stat.captures / stat.attempts
 		if stat.deaths > 1 then
 			kd = kd / stat.deaths
 		end
 		result = result ..
-			minetest.colorize("#63d437", "Kills: ") ..
+      minetest.colorize("#63d437", "Kills: ") ..
```
LoneWolfHT, line 45: This should be double-indented past line 44's indentation (same for the other similar lines)

LoneWolfHT (changes requested):

mods/ctf/ctf_stats/chat.lua
realOneplustwo: Fixed.

PR #740, opened 2021-01-06
Remove hook if player has changed their class after firing the hook

Panquesito7:
Things added/changed:

- Remove hook if player has changed their class after firing the hook.
  - This prevents users from firing the hook, change their class and teleport to their destination.

Tested with MT `5.4.0-c490a57a3-c940a57a3-dirty`.

### DIFF
--- a/mods/ctf/ctf_classes/ranged.lua
+++ b/mods/ctf/ctf_classes/ranged.lua
@@ -86,11 +86,10 @@ minetest.registered_entities["shooter_hook:hook"].on_step = function(self, dtime
 		return
 	end
 
-	-- Remove hook if player has changed
-	-- their class after firing the hook.
+	-- Remove hook if player changes class after throwing it
 	if not ctf_classes.get(self.user).properties.allow_grapples then
 		minetest.chat_send_player(self.user,
-			"You can't fire grapples and change your class!")
+			"Grapples don't work if you change class!")
 		self.object:remove()
 		return
 	end
--- a/mods/ctf/ctf_classes/ranged.lua
+++ b/mods/ctf/ctf_classes/ranged.lua
@@ -89,12 +89,8 @@ minetest.registered_entities["shooter_hook:hook"].on_step = function(self, dtime
 	-- Remove hook if player has changed
 	-- their class after firing the hook.
 	if not ctf_classes.get(self.user).properties.allow_grapples then
-		local player = minetest.get_player_by_name(self.user)
-		if player then
-			player:get_inventory():add_item("main", "shooter_hook:grapple_hook")
-		end
 		minetest.chat_send_player(self.user,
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
What code removes the regular grapple? You should be able to make use of that

Panquesito7:
> What code removes the regular grapple?

Is this what you're looking for?

https://github.com/MT-CTF/capturetheflag/blob/92203a0c83b5d158fb5ffe2e9ead90a7210c1d99/mods/ctf/ctf_classes/ranged.lua#L49-L55

LoneWolfHT:
> > What code removes the regular grapple?
> 
> Is this what you're looking for?
> 
> https://github.com/MT-CTF/capturetheflag/blob/92203a0c83b5d158fb5ffe2e9ead90a7210c1d99/mods/ctf/ctf_classes/ranged.lua#L49-L55

For an example:
What code removes a knight's sword when they switch to medic class?

mods/ctf/ctf_classes/ranged.lua
```diff
@@ -85,5 +85,19 @@ minetest.registered_entities["shooter_hook:hook"].on_step = function(self, dtime
...
+
+	-- Remove hook if player has changed
+	-- their class after firing the hook.
+	if not ctf_classes.get(self.user).properties.allow_grapples then
+		local player = minetest.get_player_by_name(self.user)
+		if player then
+			player:get_inventory():add_item("main", "shooter_hook:grapple_hook")
+		end
+		minetest.chat_send_player(self.user,
+			"Your class can't use that weapon! Change your class at spawn")
```
LoneWolfHT, line 97: IIRC when you switch class the grapple just gets removed. You should probably do the same here if that's the case

TSafa-23: Also IIRC there is a /class command, may consider adding that to the message as well

LoneWolfHT: I'm having a bad day today. My suggestion, properly worded, would be to just completely remove the highlighted code

Panquesito7: No, because this code is called `on_step`, and it is after firing the hook.

LoneWolfHT: Lines 93 to 97 return the hook and then tell the player they can't use it.
I think it would be better if the grapple wasn't returned (The player gets a fresh one when they switch to shooter class) and instead was just deleted.
IMO The message also isn't needed because it's very hard to get into this situation unless you know what you're doing. So it should be self-explanatory when you don't teleport

Panquesito7: Alright, sorry for the misunderstanding. I'll change the message and not return the hook. Thanks for explaining. 🙂

PR #742, opened 2021-01-06
Fix immunity ending when hitting teammate

realOneplustwo:


LoneWolfHT (approved):
Could use a second review in case I missed something

PR #749, opened 2021-01-13
Fix bug with medkits not allocating deaths and score after death properly

TSafa-23:
See description
How to test:
Consume medkit
Kill player in one hit

Tested, should work

### DIFF
--- a/mods/pvp/medkits/init.lua
+++ b/mods/pvp/medkits/init.lua
@@ -134,7 +134,6 @@ minetest.register_on_player_hpchange(function(player, hp, reason)
 	if hp < 0 then
 		if players[name] then
 			players[name] = nil -- Don't use stop_healing(), it uses set_hp() and won't allocate deaths or score properly
-			minetest.chat_send_player(name, minetest.colorize("#FF4444", "Your healing was interrupted"..reason_handler("damage")))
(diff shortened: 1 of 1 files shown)


mods/pvp/medkits/init.lua
```diff
@@ -130,9 +130,12 @@ end)
 -- If player takes damage while healing,
 -- stop regen and revert back to original state
 minetest.register_on_player_hpchange(function(player, hp, reason)
+	local name = player:get_player_name()
 	if hp < 0 then
-		if players[player:get_player_name()] then
-			stop_healing(player, "damage")
+		if players[name] then
+			players[name] = nil -- Don't use stop_healing(), it uses set_hp() and won't allocate deaths or score properly
+			minetest.chat_send_player(name, minetest.colorize("#FF4444", "Your healing was interrupted"..reason_handler("damage")))
```
JostP, line 137:
```suggestion
			interrupted = "damage"
```
This could work better, but no idea if it works

TSafa-23: Since stop_healing() isn't called, it would just be setting the variable value without accomplishing anything, I do think I should remove the line though, doesn't seem very useful

PR #752, opened 2021-01-15
Add nodes that affect speed

realOneplustwo:
1. Sticky cobwebs will both slow you down and almost nullify your ability to jump.
2. Quicksand will slow you down if you walk on it.
3. Asphalt will increase your speed if you walk on it.


mods/other/speed_nodes/init.lua
```diff
@@ -0,0 +1,94 @@
+minetest.register_node("speed_nodes:cobweb", {
```
TSafa-23, line 1:
```suggestion
local players = {}
minetest.register_node("speed_nodes:cobweb", {
```
IIRC the mod won't work without this, also luacheck hates you because of this missing

realOneplustwo: "players" is from the physics mod

TSafa-23: Is it global though?

realOneplustwo: Yes, I globalized it.

Panquesito7 (changes requested):
Please fix [LuaCheck](https://github.com/MT-CTF/capturetheflag/pull/752/checks?check_run_id=1712267766) warnings.

realOneplustwo:
Not necessary. I used a table from another mod (the physics mod), but I did not explicitly define it in my mod, hence LuaCheck starts complaining. But it still works.

Panquesito7:
> But it still works

It may work, but we also want the code to be clean and understandable. 🙂

realOneplustwo:
So should I edit the physics mod as well so that I can add some function to check if an override layer exists?

Panquesito7:
> So should I edit the physics mod as well so that I can add some function to check if an override layer exists?

Sounds good to me. 👍

mods/ctf/ctf_map/ctf_map_vote/init.lua
```diff
@@ -0,0 +1,102 @@
+-- code by oneplustwo
```
Panquesito7, line 1: If this is your code, then you should provide a `README.md` explaining what the mod does, how does it work, etc.. You also need to add a `LICENSE` file.

mods/ctf/ctf_map/ctf_map_vote/init.lua
```diff
@@ -0,0 +1,102 @@
...
+    local total = 0
+    for _, type in pairs(voted[map_name]) do
+        total = total + type
+    end
+    return type
+end
+
+minetest.register_chatcommand("upvote", {
+    func = function(name, param)
+        if minetest.get_player_by_name(name) == nil then
```
Panquesito7, line 51:
```suggestion
        if not minetest.get_player_by_name(name) then
```

mods/ctf/ctf_map/ctf_map_vote/init.lua
```diff
@@ -0,0 +1,102 @@
...
+            minetest.chat_send_player(name, minetest.colorize("Your voting cooldown is still active!"))
+        else
+            minetest.chat_send_player(name, minetest.colorize("You must have at least 500 score"))
+        end
+    end,
+})
+
+minetest.register_chatcommand("downvote", {
+    func = function(name, param)
+        if minetest.get_player_by_name(name) == nil then
```
Panquesito7, line 68:
```suggestion
        if not minetest.get_player_by_name(name) then
```

mods/ctf/ctf_map/ctf_map_vote/init.lua
```diff
@@ -0,0 +1,102 @@
...
+            minetest.chat_send_player(name, minetest.colorize("Your voting cooldown is still active!"))
+        else
+            minetest.chat_send_player(name, minetest.colorize("You must have at least 500 score"))
+        end
+    end,
+})
+
+minetest.register_chatcommand("unvote", {
+    func = function(name, param)
+        if minetest.get_player_by_name(name) == nil then
```
Panquesito7, line 85:
```suggestion
        if not minetest.get_player_by_name(name) then
```

mods/ctf/ctf_map/ctf_map_vote/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = ctf_map_vote
+depends = ctf_map_core, ctf_stats
```
Panquesito7, line 2:
```suggestion
depends = ctf_map_core, ctf_stats

```

Panquesito7 (commented):
One question: shouldn't the mod `ctf_map_vote` be added in a separate PR?

realOneplustwo:
I accidentally pushed to the wrong branch, my bad.

Panquesito7:
Try not to make new PR's, update the current PR so it'll be easier for others to review it.

PR #753, opened 2021-01-18
Added map rating feature

realOneplustwo:
Use /rate [rating] to rate the current map. [rating] must be 1 to 5. To cancel your rating, use /rate with no parameters.
Very WIP. Extensive testing needed.


LoneWolfHT:
Remember to add newlines at the end of your files

JostP:
Add  [random message](https://github.com/MT-CTF/capturetheflag/blob/2e3ffe8be0ccdc7ffad228604887f2777561a796/mods/other/random_messages/init.lua#L50), so player will know how to vote.

mods/other/random_messages/init.lua
```diff
@@ -79,6 +79,7 @@ function random_messages.read_messages()
 		"Hitting your enemy does more damage than not hitting them.",
 		"Press right mouse button or double-tap the screen to activate scope while wielding a sniper rifle.",
 		"The 'Updates' tab in your inventory will show some of the latest updates to CTF",
+		"Use /rate [rating] to rate the current map on a scale of 1 to 5. If /rate is used alone, it will remove your rating.",
```
Panquesito7, line 82:
```suggestion
		"Use /rate [rating] to rate the current map on a scale of 1 to 5. If /rate is used without parameters, it will remove your rating.",
```

mods/ctf/ctf_map/ctf_map_rating/init.lua
```diff
@@ -0,0 +1,112 @@
+-- code by oneplustwo
```
Panquesito7, line 1: Maybe add a `README.md` file explaining what this mod does, and a `LICENSE` file (this is required and extremely recommended).

mods/ctf/ctf_map/ctf_map_rating/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = ctf_map_rating
+depends = ctf_map_core, ctf_stats
```
Panquesito7, line 2: Maybe add a description saying what the mod does?

Panquesito7 (changes requested):
Please fix [LuaCheck](https://github.com/MT-CTF/capturetheflag/runs/1730300632) warnings.

mods/ctf/ctf_map/ctf_map_rating/init.lua
realOneplustwo: This isn't a standalone mod though, are you sure it's required?

mods/ctf/ctf_map/ctf_map_rating/mod.conf
realOneplustwo: Pretty self-explanatory imo.

mods/ctf/ctf_map/ctf_map_rating/init.lua
Panquesito7: It is required, all mods should have a license set.

realOneplustwo: But it's part of a modpack.

xenonca:
Ideally I would love to see the option to rate a map in the maps catalogue, along with what the current map rating is, but I'm aware this is probably even more complicated to implement

mods/ctf/ctf_map/ctf_map_rating/init.lua
xenonca: I don't think it's required either, if not stated otherwise it's licensed under the same license as the game (LGPLv2.1+)

xenonca:
**Problem:** Type `/rate`, this will return "Rating must be an integer". However then you can't rate a map with an integer cause the rating cooldown is active.

PR #754, opened 2021-01-18
Add nodes that affect speed

realOneplustwo:
New PR because I messed up the old one.
1. Sticky cobwebs will both slow you down and almost nullify your ability to jump.
2. Quicksand will slow you down if you walk on it.
3. Asphalt will increase your speed if you walk on it.


mods/other/physics/init.lua
```diff
@@ -66,3 +66,10 @@ function physics.remove(name, layer)
 	players[name][layer] = nil
 	update(name)
 end
+
+function physics.layer_exists(name, layer)
+	-- Basic sanity checks uwu
```
Panquesito7, line 71:
```suggestion
	-- Sanity checks
```

mods/other/speed_nodes/README.md
```diff
@@ -0,0 +1,13 @@
+# Speed Nodes
+
+Created by oneplustwo.
```
Panquesito7, line 3:
```suggestion
Created by [realOneplustwo](https://github.com/realOneplustwo).
```

mods/other/speed_nodes/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+License: MIT.
```
Panquesito7, line 5:
```suggestion
## License

Copyright (C) 2021 realOneplustwo

Speed Nodes is licensed under MIT.
See [`LICENSE`](https://github.com/MT-CTF/capturetheflag/blob/master/mods/other/speed_nodes/LICENSE) for more information.
```

mods/other/speed_nodes/init.lua
```diff
@@ -0,0 +1,94 @@
+minetest.register_node("speed_nodes:cobweb", {
```
Panquesito7, line 1:
```suggestion
-- Capture the Flag mod: speed_nodes

minetest.register_node("speed_nodes:cobweb", {
```

mods/other/speed_nodes/README.md
realOneplustwo: Would a relative path be better?

TSafa-23: It may be better to just link to the MIT license, also this isn't really necessary

realOneplustwo:
I think this mod would be better used for a specific map instead of adding it to the entire repo, since it just adds nodes.

PR #762, opened 2021-01-25
Make the door's window more wider

winniepee:
screenshot_20210123_163229
The screenshot is an example for the textures.


Panquesito7 (commented):
Why?

moreavy:
We may not need them

Toby1710:
It doesn't look very nice... The texture that is now better.

moreavy:
The team doors in ctf refer to the original `steel door` in the Minetest Game. Make Pull in the [Minetest Game Repository](https://github.com/minetest/minetest_game) instead of this one.

winniepee:
> It doesn't look very nice... The texture that is now better.

I am OK with that, I already knew this will fail.

winniepee:
> We may not need them. If you need so.. change your textures
> 
> ```
> But if you change your textures.. only you may be able to hit bypass the doors while others  wouldn't.
> ```

The textures doesn't affect anything at all

Texture = texture
Node = node

Anyone can hit bypass the doors because of the bug issues, not the textures.

moreavy:
> > We may not need them. If you need so.. change your textures
> > ```
> > But if you change your textures.. only you may be able to hit bypass the doors while others  wouldn't.
> > ```
> 
> The textures doesn't affect anything at all
> 
> Texture = texture
> Node = node
> 
> Anyone can hit bypass the doors because of the bug issues, not the textures.

Read https://github.com/MT-CTF/capturetheflag/pull/762#issuecomment-766717018

> The team doors in ctf refer to the original `steel door` in the Minetest Game. Make Pull in the [Minetest Game Repository](https://github.com/minetest/minetest_game) instead of this one.

LoneWolfHT:
> The team doors in ctf refer to the original `steel door` in the Minetest Game. Make Pull in the [Minetest Game Repository](https://github.com/minetest/minetest_game) instead of this one.

We don't follow latest MTG. It's best made here

moreavy:
Btw why do we need to make door's windows wider? What's the need @winniepee ?

PR #763, opened 2021-01-28
Add all Bounty Kills of all players to summary

DiamondPlane:
Adds to the top of the summary bounty kills
image


Panquesito7 (approved):

JostP:
Not sure, but I don't think that this is not really needed.

Toby1710:
> Not sure, but I don't think that this is really needed.

Attempts, kills, deaths, etc are there but no Bounty Kills

PR #768, opened 2021-01-31
Improve marker in-chat message

JostP:
Before:

Zaslonska slika 2021-01-31 12-01-24

Now:

Zaslonska slika 2021-01-31 12-54-08

Tested and works.

### DIFF
--- a/mods/ctf/ctf_marker/init.lua
+++ b/mods/ctf/ctf_marker/init.lua
@@ -54,19 +54,8 @@ function ctf_marker.add_marker(name, tname, pos, str)
 		end
 		minetest.log("action", name .. " placed a marker at " ..
 				minetest.pos_to_string(pos) .. ": '" .. str .. "'")
-
-		local marker_team = "#ffffff"
-		if tname == "red" then
-			marker_team = ctf_colors.colors["red"]
-		end
-
-		if tname == "blue" then
-			marker_team = ctf_colors.colors["blue"]
-		end
-		marker_team = "#" .. tostring(marker_team):sub(3, 8)
-
 		minetest.chat_send_player(pname,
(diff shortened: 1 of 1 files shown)


moreavy (approved):
Small changes but nice ones.

JostP:
> Small changes but nice ones.

It was quite hard to see old marker messages in chat. Now they stand out a bit more I hope.

It also adds marker string to in-chat message, so you immediately see it, and if marker tells important information, you will look around to find it.

xenonca:
I think it's not necessary and looks worse than before. You can only see your teammates markers anyway so no need to color it. And it looks more like a chat message now. Instead that message should be moved to a HUD element like points.

moreavy:
But the color specifies the marker messages. It is at least better than the plain white color. 😛

xenonca:
I would probably find it less annoying if this would be displayed in the HUD instead but the chat is already so bloated with other "status messages". I don't see a point in coloring, also it's not plain white but light blue. However including the marker string is probably a good idea.

JostP:
> I would probably find it less annoying if this would be displayed in the HUD instead but the chat is already so bloated with other "status messages". I don't see a point in coloring, also it's not plain white but light blue. However including the marker string is probably a good idea.

My main reason to create PR was to add marker string to message. If you think that colorizing it is not a good idea, I can use default message for whole message. Adding red or blue was meant to make marker mo visible.

moreavy:
> If you think that colorizing it is not a good idea, I can use default message for whole message.
Not don't. It looks good **IMO**

xenonca:
The light blue was probably fine, and I like the marker string being included.

JostP:
Zaslonska slika 2021-01-31 12-54-08

xenonca:
See #642, that was what I meant

PR #769, opened 2021-01-31
Edit the marker name placed with knight sword

DiamondPlane:
For example: DiamondPlane's sword marker
JostP suggested changes so it will be now:
DiamondPlane's marker like when you just type /m

### DIFF
--- a/mods/ctf/ctf_classes/melee.lua
+++ b/mods/ctf/ctf_classes/melee.lua
@@ -105,7 +105,7 @@ minetest.register_tool("ctf_classes:sword_bronze", {
 		sword_special_timer[pname] = 20
 		sword_special_timer_func(pname, 20)
 
-		minetest.registered_chatcommands["m"].func(pname, "Marked with "..pname.."'s sword")
+		minetest.registered_chatcommands["m"].func(pname, ""..pname.."'s sword marker")
 	end,
 	on_secondary_use = function(itemstack, user, pointed_thing)
 		if pointed_thing then
--- a/mods/ctf/ctf_classes/melee.lua
+++ b/mods/ctf/ctf_classes/melee.lua
@@ -105,7 +105,7 @@ minetest.register_tool("ctf_classes:sword_bronze", {
 		sword_special_timer[pname] = 20
 		sword_special_timer_func(pname, 20)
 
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_classes/melee.lua
```diff
@@ -105,7 +105,7 @@ minetest.register_tool("ctf_classes:sword_bronze", {
 		sword_special_timer[pname] = 20
 		sword_special_timer_func(pname, 20)
 
-		minetest.registered_chatcommands["m"].func(pname, "Marked with "..pname.."'s sword")
+		minetest.registered_chatcommands["m"].func(pname, ""..pname.."'s sword marker")
```
moreavy, line 108:
```suggestion
		minetest.registered_chatcommands["m"].func(pname, ""..pname.."'marked with sword")
```

"player's sword marker" doesn't sound good.

JostP:
I think that you can completely remove "Marked with " "'s sword" strings. They make message too long

mods/ctf/ctf_classes/melee.lua
JostP, line 108:
```suggestion
		minetest.registered_chatcommands["m"].func(pname, " placed"..pname.." a marker")
```

Toby1710: Here you forgot to remove '

JostP:
You can also remove timer in this PR

Zaslonska slika 2021-02-01 10-04-54

Toby1710:
> You can also remove timer in this PR
> 
> Zaslonska slika 2021-02-01 10-04-54

I don't know, maybe they will do it.

mods/ctf/ctf_classes/melee.lua
```diff
@@ -105,7 +105,7 @@ minetest.register_tool("ctf_classes:sword_bronze", {
 		sword_special_timer[pname] = 20
 		sword_special_timer_func(pname, 20)
```
Panquesito7, line 106: Well, if you want to remove the timer, perhaps removing these two lines will do the trick.

LoneWolfHT:
Note that I don't want the marker triggered on rightclick of a player to be removed

PR #770, opened 2021-01-31
Team chest: better initial stuff

JostP:
Before

Zaslonska slika 2021-01-31 15-59-21

Now:
Zaslonska slika 2021-01-31 16-00-01


LoneWolfHT:
This takes a bit out of the chest looting and mining incentive

TSafa-23:
You could say that, but it is only two team doors, often I see way more than two being used.

LoneWolfHT:
Guess we can try it then

PR #773, opened 2021-02-01
Crafting Change

moreavy:
These are a few changes to remove *bronze* in the crafting recipes because the old *copper* and *tin* were removed. Tested and working without bugs.

### DIFF
--- a/mods/ctf/ctf_crafting/init.lua
+++ b/mods/ctf/ctf_crafting/init.lua
@@ -61,6 +61,15 @@ crafting.register_recipe({
 	level  = 1,
 })
 
+-- Diamond x9 <== Diamond Block
+crafting.register_recipe({
+  type = "inv",
+  output = "default:diamond 9",
+  items = "default:diamondblock",
+  always_known = true,
+  level  = 1
+})
+
 -- Furnace <== group:stone x8
 crafting.register_recipe({
 	type   = "inv",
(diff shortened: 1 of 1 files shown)


moreavy:
Should I also add diamond recipe from the diamond block?
```lua
crafting.register_recipe({
  type = "inv",
  output = "default:diamond",
  items = "default:diamondblock",
  always_known = true,
  level  = 1
})
```

Panquesito7:
Can someone re-run LuaCheck checks?

LoneWolfHT:
Done

moreavy:
Please suggest for changes or additions.

mods/ctf/ctf_crafting/init.lua
```diff
@@ -198,15 +180,15 @@ crafting.register_recipe({
 
 crafting.register_recipe({
 	type   = "inv",
-	output = "ctf_traps:dirt 1",
+	output = "ctf_traps:dirt 4",
```
Toby1710, line 183:
```suggestion
	output = "ctf_traps:dirt 5",
```
I recommend to make it 5 because if you put 5 dirt + coal lump, the amount should be the same

LoneWolfHT:
Why are you making the trap crafts cheaper?

moreavy:
- Traps are not used by common players, very few players craft them.
- IMO, unwalkable nodes should be even cheaper because they are __unwalkable__.
- We get just 1 node while we use 5 nodes to craft it. That's too much IMO.
- Some map don't even have the require material to craft them.

LoneWolfHT (approved):
Okie, guess this is good then

PR #782, opened 2021-02-06
Master(kill assist) bugfix

Anthony-De:


LoneWolfHT:
On second thought this should not be in `ctf`. One sec...

LoneWolfHT:
I'm familiar with the code now and I don't see this crashing easily so I guess I'll merge

Issue #793, opened 2021-02-11
You get kill assist points if you attack at build time


savilli:
You can't do any damage but still get points.
No build time check in https://github.com/MT-CTF/capturetheflag/blob/master/mods/pvp/kill_assist/init.lua#L13.
The same is for respawn immunity.

PR #795, opened 2021-02-11
Don't count kill assist on build time or player immunity

savilli:
Fix https://github.com/MT-CTF/capturetheflag/issues/793
Introduce ctf.register_can_attack callbacks that can stop any ctf related attack logic.

### DIFF
--- a/mods/ctf/ctf/teams.lua
+++ b/mods/ctf/ctf/teams.lua
@@ -446,12 +446,8 @@ function ctf.register_on_killedplayer(func)
 	table.insert(ctf.registered_on_killedplayer, func)
 end
 
-ctf.registered_can_attack = {}
-function ctf.register_can_attack(func)
-	if ctf._mt_loaded then
-		error("You can't register callbacks at game time!")
-	end
-	table.insert(ctf.registered_can_attack, func)
+function ctf.can_attack(player, hitter, time_from_last_punch, tool_capabilities, dir, damage, ...)
+	return true
 end
 
 ctf.registered_on_attack = {}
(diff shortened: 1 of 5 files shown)


mods/ctf/ctf/teams.lua
```diff
@@ -446,12 +446,20 @@ function ctf.register_on_killedplayer(func)
...
-ctf.registered_on_punchplayer = {}
-function ctf.register_on_punchplayer(func)
+ctf.registered_can_attack = {}
+function ctf.register_can_attack(func)
 	if ctf._mt_loaded then
 		error("You can't register callbacks at game time!")
 	end
-	table.insert(ctf.registered_on_punchplayer, func)
+	table.insert(ctf.registered_can_attack, func)
+end
```
LoneWolfHT, line 455: Just create a ctf.can_attack() function. Mods will still be able to add things to it if needed. Will result in clearer code too

LoneWolfHT (changes requested):

mods/ctf/ctf/teams.lua
savilli: >  Mods will still be able to add things to it if needed
How so? What mod should do if it want to reject the damage?

LoneWolfHT:
```lua
local oldfunc = ctf.can_attack
ctf.can_attack = function()
    -- New code here, can call the old function with oldfunc()
end
```

savilli: We already have a common way to register callback using register_ functions. It's used in minetest api as well as in ctf (ctf.register_on_killedplayer, ctf.register_on_punchplayer). Why you want to do it another way? In my view two different ways to register callback will only add confusion and decrease code readability.

LoneWolfHT: The register_on() callbacks are used to let mods change behavior of something else. In this case you want a mod to change it's behavior based on what something else says. In a situation like that you use the method I just mentioned.
See minetest.is_protected() for an example. CTF also has a few functions doing the same. Although I forget where exactly they are.

savilli: It's absolutely not clear how to distinguish two cases. But minetest.is_protected is a valid case. If you insist on that way I'll do like is_protected.

PR #801, opened 2021-02-17
Modified team door textures

realOneplustwo:
Many people are already using this in texture packs, setting it as default would make it fair for everyone.


Panquesito7 (commented):
To be honest, I like the current textures. We should keep the default textures; anyone who wants to use it should use a texture pack.

realOneplustwo:
Many don't even know how to use texture packs. Having only some players to have better visibility through team doors gives an unfair advantage in my opinion.

TSafa-23:
Just a comment and not a serious suggestion, you could probably use a grayscale texture and use texture modifiers for red/blue, save a few of those wonderful bytes

PR #802, opened 2021-02-17
Change 'base' to 'spawn' when player can't use a weapon

Toby1710:
It says "Your class can't use this weapon! Change your class at spawn". Maybe better is "Change your class at base"?


Panquesito7 (commented):
Well, you spawn at your base, so both are actually OK.

PR #698, opened 2020-10-30
Disable build-where-somebody-stands serverside

appgurueu:
Counters pillaring.

PR #803, opened 2021-02-18
Limit node placement

### LINKED ISSUE #819: Pillaring race condition bug
A bug frequently observed with fast pillarers is as follows:

* Pillarer pillars quickly using only paxelable blocks (such as cobblestone)
* Medic paxels pillar
* Pillarer sends node placement at top of pillar to server
* Server processes paxeling: removes node being placed to, terminates at air (race condition)
* Node placement is done, pillarer happily continues pillaring (without even switching pillars!)

### LINKED ISSUE #820: Rate limit pillaring
Pillaring is hardly ratelimited (possibly ratelimited through client FPS / network level). This can certainly be exploited using hacked clients or alternatively in the form of spamplacing blocks using insane values for `place_repeat_time`.

appgurueu:
Second try (see #698), as I'm still frustrated with how node placement is abused. Simple solution: Limits node placement to a sane speed. Can be increased to 10 nodes per second if too slow. ~~Won't apply to team doors unless [this fix](https://github.com/MT-CTF/capturetheflag/pull/698/files#diff-3de2a0a8a27905263bcf2f5a3213265e015b7408874b6ed27a307aade5d91ecfR357-R370) is implemented too~~. Fixes #819. Fixes #820.

### DIFF
--- a/mods/mtg/doors/init.lua
+++ b/mods/mtg/doors/init.lua
@@ -354,14 +354,20 @@ function doors.register(name, def)
 				meta:set_string("owner_team", tname)
 			end
 
+			local copy = table.copy
+			local newnode = minetest.get_node(pos)
+			for _, on_placenode in pairs(minetest.registered_on_placenodes) do
+				if on_placenode(copy(pos), copy(newnode), placer, copy(node), ItemStack(itemstack), copy(pointed_thing)) then
+					return itemstack
+				end
+			end
 			if not (creative and creative.is_enabled_for and creative.is_enabled_for(pn)) then
 				itemstack:take_item()
 			end
 
 			minetest.sound_play(def.sounds.place, {pos = pos})
 
-			on_place_node(pos, minetest.get_node(pos),
-				placer, node, itemstack, pointed_thing)
+			on_place_node(pos, newnode, placer, node, itemstack, pointed_thing)
 
 			return itemstack
 		end
--- a/mods/pvp/place_limit/License.txt
+++ b/mods/pvp/place_limit/License.txt
@@ -1,0 +1,19 @@
+Copyright (c) 2021  appgurueu
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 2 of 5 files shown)


TSafa-23:
Moreover curious, does this give you the node back if you can't place it?

Calinou:
A fast clicker may be able to place 6 nodes or even 7 nodes, so 5 per second might be a bit frustrating.

appgurueu:
> A fast clicker may be able to place 6 nodes or even 7 nodes in one second, so limiting to 5 nodes per second might be a bit frustrating to those who can click fast enough.

As said, the limit should probably be bumped to 10/7 placements per second.

> Also, there should be a chat message or HUD notification when trying to place nodes too fast.

I haven't come up with a non-spammy way to convey this yet.

appgurueu:
> Moreover curious, does this give you the node back if you can't place it?

If `on_placenode` is properly executed, yes. Docs say if it returns `true`, the item should not be taken.

LoneWolfHT:
What's the difference between get_us_time() and os.time()?

Panquesito7 (changes requested):
The mod is missing a license and a `README.md` file. For the license: I recommend using [MIT](https://mit-license.org/).

appgurueu:
> What's the difference between get_us_time() and os.time()?

A simple call to `os.time()` will return time with second precision, which is absolutely useless for this application (as multiple places per second should be possible).

appgurueu:
> The mod is missing a license and a `README.md` file. For the license: I recommend using [MIT](https://mit-license.org/).

Quite bloaty, isn't it?

LoneWolfHT:
Readme.md isn't required. But a license is a good idea. I always forget that

appgurueu:
Closes #819

Panquesito7 (changes requested):
You should provide a `LICENSE.txt` file (see the [MIT](https://mit-license.org) license website).

appgurueu:
@moreavy note that the specific limitation is still subject to discussion. There should however be some, even if it's quite "liberal" (like 15, 20 or even 30 blocks per second), as otherwise, the client is hardly ratelimited.

moreavy:
> @moreavy note that the specific limitation is still subject to discussion. There should however be some, even if it's quite "liberal" (like 15, 20 or even 30 blocks per second), as otherwise, block placement is not restricted at all. Hacked clients can abuse this by spamming large cobble structures in the blink of an eye.

I understood that, but try keeping the node placement limits per seconds to be less, so that it won't affect the normal gameplay.

appgurueu:
This is ready for merge. The poll on Discord has decided for <= 5 blocks per second, which is what this PR implements.

If this turns out to be too restricting, it can be easily tweaked (`init.lua#3`).

Anthony-De:
Looks good. 👍

PR #804, opened 2021-02-20
Prevent players to throw frag/smoke grenade on build time

DiamondPlane:


mods/pvp/grenades/grenades.lua
```diff
@@ -16,6 +16,14 @@ local fragdef = {
...
 			return
 		end
+                if ctf_match.is_in_build_time() then
+		local player = minetest.get_player_by_name(name)
+                if player then
+			player:get_inventory():add_item("main", "grenades:frag")
+                        return
+		        minetest.chat_send_player(name, "Match hasn't started yet!")
+                end
+        end
```
Panquesito7, line 26: Wrong indentation.

mods/pvp/grenades/grenades.lua
```diff
@@ -94,6 +102,13 @@ grenades.register_grenade("grenades:smoke", {
 	on_explode = function(pos, pname)
 		local player = minetest.get_player_by_name(pname)
 		if not player or not pos then return end
+                if ctf_match.is_in_build_time() then
+                if player then
+			player:get_inventory():add_item("main", "grenades:smoke")
+                        return
+		        minetest.chat_send_player(pname, "Match hasn't started yet!")
+                end
+        end
```
Panquesito7, line 111: Same here as above.

Panquesito7 (approved):
👍 apart from the above (tested).

Toby1710:
Hmm... Can you suggest changes me to make the indentation correct, please?

TSafa-23 (changes requested):
Be sure to fix redundancies/indentation otherwise it likely won't be merged. Also, does this prevent grenades from being thrown or only giving them back if they were thrown before the build time is up?

Toby1710:
https://github.com/MT-CTF/capturetheflag/pull/804#discussion_r579849054 @TSafa-23  I tested with this, it gave me error. But on frag it didn't

TSafa-23:
Was more of an example because you asked to see one btw, just so you know how it roughly should look

Toby1710:
> Was more of an example because you asked to see one btw, just so you know how it roughly should look

Should I leave the code that I did?

DiamondPlane:
I have a problem with this PR, I'll create a new one.

PR #810, opened 2021-02-24
Make your last hitter get a kill if you suicide any possible way

### LINKED ISSUE #756: It is possible to hit and kill yourself
I discovered it a while ago, and it leads to very strange kill messages. I mean how is that even possible? Should you get the message "You are in your team!" ?

How to reproduce:
Using a dragonfire script:
```lua
minetest.register_globalstep(function()
	local player = minetest.localplayer
	if not player then return end
	local playercao = player:get_object()
	local control = player:get_control()
	if control.dig then
		playercao:punch()
	end
end)
```
Do not ask me how I found this out.

savilli:
Fix #756

### DIFF
--- a/mods/ctf/ctf_classes/ranged.lua
+++ b/mods/ctf/ctf_classes/ranged.lua
@@ -73,6 +73,11 @@ check_grapple("shooter_hook:grapple_hook")
 -- Override grappling hook entity to check if player has flag before teleporting
 local old_grapple_step = minetest.registered_entities["shooter_hook:hook"].on_step
 minetest.registered_entities["shooter_hook:hook"].on_step = function(self, dtime, ...)
+	-- User left the game. Life is no longer worth living for this poor hook
+	if not self.user then
+		self.object:remove()
+	end
+
 	-- Remove entity if player has flag
 	-- This is to prevent players from firing the hook, and then punching the flag
 	if ctf_flag.has_flag(self.user) then


mods/pvp/anticoward/init.lua
```diff
@@ -15,18 +15,7 @@ ctf.register_on_attack(function(player, hitter,
 		local pname = player:get_player_name()
 		local hname = hitter:get_player_name()
 
-		local hp = player:get_hp() - damage
-		if hp <= 0 then
```
savilli, line 19: Dead code since on_attack is not called if player:get_hp() - damage <= 0

savilli:
Testing would be nice
Test cases:
- Honest death in battle 
- Combatlogging
- Run from a battle and suicide using fail damage
- Run from a battle and suicide using a grenade

xenonca:
Why would the last hitter get the points? I don't know if that is very fair in every case. I'd prefer a solution where you're simply unable to kill yourself (unless it's a nade or so)

savilli:
> Why would the last hitter get the points? I don't know if that is very fair in every case. I'd prefer a solution where you're simply unable to kill yourself (unless it's a nade or so)

How it's not fair? If you die by falling your your last hitter will get kill. Kill yourself using a nade is a suicide too. What is the reason why it is not the same?

savilli:
Just a quick summary from the discussion.

In the old days if you didn't want to give your enemy a kill you could run away from a battle and suicide. Anticoward system was designed to prevent such tactic. If you suicide using fail damage your last hitter will get a kill.
But if you die from hitting yourself or nading yourself your enemy still doesn't get a kill because anticoward system doesn't see it as a suicide. This pr will make that your last hitter get a kill if you suicide ANY possible way.

You can forbid players to hit themself but after this pr it will be completely useless so I don't see a reason to do it.

xenonca:
> You can forbid players to hit themself but after this pr it will be completely useless so I don't see a reason to do it.

Well, because I think that that's the proper solution, to prevent players from hitting themselves. Also it might not necessarily be useless if we find another way of suiciding we haven't thought of yet

savilli:
> prevent players from hitting themselves.

No need to prevent players from hitting themselves just because. I don't think that hitting yourself and nading yourself are different and need different approach to handle.

realOneplustwo:
Grenading yourself was, I believe, an intended consequence of adding grenades. Being able to hit yourself is not an intended feature, the two are vastly different.

realOneplustwo:
What problem does this PR solve that cannot be solved through simpler means?

savilli:
> Grenading yourself was, I believe, an intended consequence of adding grenades. Being able to hit yourself is not an intended feature, the two are vastly different.

It's not relevant
> What problem does this PR solve that cannot be solved through simpler means?

This PR is pretty simple

xenonca:
It is very relevant as killing yourself with a nade (ideally killing some enemies with it too) can be a tactical reason to suicide. On the other side a PR is usually not good **just** because it's simple.

realOneplustwo:
The simplest way to fix #756 is simply to not let the player hit herself.

savilli:
> The simplest way to fix #756 is simply to not let the player hit herself.

It's more complicated since you need to distinguish hitting yourself using a sword or a grenade

PR #811, opened 2021-02-25
Give 10 captures on makepro command

DiamondPlane:
When you use /makepro <name> then it doesn't give captures and you can't have pro chest. I made it that it gives 10 captures

### DIFF
--- a/mods/ctf/ctf_stats/chat.lua
+++ b/mods/ctf/ctf_stats/chat.lua
@@ -240,10 +240,10 @@ minetest.register_chatcommand("makepro", {
 			modified = true
 		end
 
-        if stats.captures < 10 then
-            stats.captures = 10
-            modified = true
-        end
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_stats/chat.lua
```diff
@@ -240,6 +240,12 @@ minetest.register_chatcommand("makepro", {
 			modified = true
 		end
 
+		if stats.captures < 10 then
+			stats.captures = 10
+			modified = true
+		end
+
+
```
LoneWolfHT, line 248: Don't need two empty lines here

LoneWolfHT:
Pushed the code without the extra line as you don't allow me to edit your PRs

DiamondPlane:
> Pushed the code without the extra line as you don't allow me to edit your PRs

I can allow.

PR #812, opened 2021-02-25
Prevent players to throw frag/smoke grenade on build time

DiamondPlane:
I made a new one because https://github.com/MT-CTF/capturetheflag/pull/804#issuecomment-785733553

### DIFF
--- a/mods/pvp/grenades/grenades.lua
+++ b/mods/pvp/grenades/grenades.lua
@@ -17,6 +17,13 @@ local fragdef = {
 			return
 		end
 
+        if ctf_match.is_in_build_time() then
+           			minetest.get_player_by_name(name):get_inventory():add_item("main", "grenades:frag")
+                                      return
+            minetest.chat_send_player(pname, "Match hasn't started yet!")
+        end
+    end
+
 		local player = minetest.get_player_by_name(name)
 
 		local radius = 10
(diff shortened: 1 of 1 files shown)


mods/pvp/grenades/grenades.lua
```diff
@@ -17,6 +17,12 @@ local fragdef = {
 			return
 		end
 
+        if ctf_match.is_in_build_time() then
+			minetest.get_player_by_name(name):get_inventory():add_item("main", "grenades:frag")
+                                      return
+            minetest.chat_send_player(name, "Match hasn't started yet!")
```
LoneWolfHT, line 23: This is never called because of the return statement above it

mods/pvp/grenades/grenades.lua
```diff
@@ -95,6 +101,14 @@ grenades.register_grenade("grenades:smoke", {
 		local player = minetest.get_player_by_name(pname)
 		if not player or not pos then return end
 
+        if ctf_match.is_in_build_time() then
+            if player then
+            player:get_inventory():add_item("main", "grenades:smoke")
+                                      return
+            minetest.chat_send_player(pname, "Match hasn't started yet!")
```
LoneWolfHT, line 108: This is never called because of the return statement above it

LoneWolfHT (changes requested):
This needs to use tabs instead of spaces

mods/pvp/grenades/grenades.lua
DiamondPlane: Can't edit

PR #815, opened 2021-02-27
Good news for miners: Mine to get score! 

Emojigit:
## Content of this PR
Mine these ores to get score!
- coal: 1 score
- gold and iron: 2 score
- diamond and mese: 3 score
## Why I do this
I become a miner (seldom kills) in "The Wall" and cleared a mine. But finally, I just get a few scores because I seldom kill. **THAT IS NOT FAIR!**
So, I work on this idea to help miners to get scores.

### DIFF
--- a/mods/ctf/ctf_mining/init.lua
+++ b/mods/ctf/ctf_mining/init.lua
@@ -1,19 +1,12 @@
 minetest.register_on_dignode(function(pos, oldnode, digger)
-	print(dump(pos))
-	print(dump(oldnode))
-	print(digger)
 	if not digger:is_player() then
 		return
 	end
 	local nname = oldnode.name
 	local pname = digger:get_player_name()
 	local ndef = minetest.registered_nodes[nname]
-	print(dump(ndef.groups))
-	print(ndef.groups.ores)
 	if ndef and ndef.groups and ndef.groups.ores then
 		local main, match = ctf_stats.player(pname)
-		print(dump(main))
-		print(dump(match))
 		if main and match then
 			main.score  = main.score  + ndef.groups.ores
 			match.score = match.score + ndef.groups.ores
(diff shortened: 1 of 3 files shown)


winniepee:
Why would players should mine to get scores? Isn't this kinda weird?

Emojigit:
> Why would players mine to get scores? Isn't this kinda weird?

In some maps, like "The wall", mining is an important part of playing. Miners provides minerals that can craft into generates, swords, guns, etc. If no miners then no win. But miners get no points!

xenonca:
> Why would players should mine to get scores? Isn't this kinda weird?

Because this also helps your team win? Giving score for coal might be a bit too much but mining iron/mese/dia should indeed be rewarded IMO

winniepee:
> > Why would players should mine to get scores? Isn't this kinda weird?
> 
> Because this also helps your team win? Giving score for coal might be a bit too much but mining iron/mese/dia should indeed be rewarded IMO

This will encourage players camping (And yeah, why coal should be included?)

xenonca:
How will this encourage camping? You might want to get these ores anyway, the only difference is that you'll be rewarded for it. Once they're dug the points are gone anyway, no need to "camp" anymore.

realOneplustwo:
Getting the ore is a reward in itself, point gain is not necessary.

DiamondPlane:
I wouldn't do this or even if you'd add I recommend to make only if you dig mese and diamonds.

DiamondPlane:
About gold: there no gold in CTF and even if there would be - you can't craft anything.

appgurueu:
Code-wise not the way to go. Mod name is poorly chosen as well as group name. I'm not familiar with the API, but immediate `request_save()`? Sounds laggy.

savilli:
I don't think we need that

Emojigit:
I will rework that once I get a developer support. This PR will remain as a conversation.

savilli:
No interest from core developers. Closing it for now, can be reopened later.

PR #816, opened 2021-02-27
Seperate "Build Time" and "Cannot dig" pillar digging failed message

Emojigit:


mods/ctf/ctf_classes/medic.lua
```diff
@@ -191,8 +191,11 @@ minetest.register_tool("ctf_classes:paxel_bronze", {
 		if pointed_thing.type == "node" then
 			local pname = placer:get_player_name()
 
-			if not isdiggable(minetest.get_node(pointed_thing.under).name) or ctf_match.is_in_build_time() then
-				minetest.chat_send_player(pname, "Can't dig node or build time active")
+			if not isdiggable(minetest.get_node(pointed_thing.under).name) then
+				minetest.chat_send_player(pname, "Can't dig node!")
+				return minetest.item_place(itemstack, placer, pointed_thing)
+			elseif ctf_match.is_in_build_time() then
+				minetest.chat_send_player(pname, "Build time active!")
```
DiamondPlane, line 198:
```suggestion
				minetest.chat_send_player(pname, "You can't dig node when build time is active!")
```
Would be this better?

xenonca: Not really IMO, but it should also be added as a HUD message (`hud_events`)

DiamondPlane: That's not to me

PR #817, opened 2021-02-28
mark hit and kill yourself as suicide

### LINKED ISSUE #756: It is possible to hit and kill yourself

Emojigit:
Close #756 
I have no Dragonfire client so this change did not test yet.

### DIFF
--- a/mods/ctf/ctf/teams.lua
+++ b/mods/ctf/ctf/teams.lua
@@ -498,6 +498,10 @@ minetest.register_on_punchplayer(function(player, hitter,
 		end
 
 		if hp - damage <= 0 then
+			if pname == hname then
+				tool_capabilities.damage_groups = tool_capabilities.damage_groups or {}
+				tool_capabilities.damage_groups.suicide = true
+			end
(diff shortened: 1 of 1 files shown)


realOneplustwo:
How does this solve the underlying problem?

Emojigit:
> underlying

What is "underlying"?

realOneplustwo:
The problem is isn't that the kill feed isn't accurate, it is that you can hit and kill yourself.

appgurueu:
The underlying problem has been fixed by an engine change. I suggest this PR be closed as well @LoneWolfHT.

PR #826, opened 2021-03-12
Sword marker: Decrease cooldown, fix trivial bug

JostP:
slika

PR decreases cooldown to 3s instead of 20s and removes bug: currently when you open team chest or doors with sword in hand you get "can't place marker yet" message:
Zaslonska slika 2021-03-12 16-52-44

Tested and everything works, is ready to be merged.
Zaslonska slika 2021-03-12 17-19-06

P.S. PR also makes that timer `timeleft` is checked every second. Currently it checked every 10 seconds. ~~I hope it won't make more lag.~~

### DIFF
--- a/mods/ctf/ctf_classes/melee.lua
+++ b/mods/ctf/ctf_classes/melee.lua
@@ -29,7 +29,7 @@ end, true)
 
 
 local sword_special_timer = {}
-local SWORD_SPECIAL_COOLDOWN = 10
+local SWORD_SPECIAL_COOLDOWN = 20
 local function sword_special_timer_func(pname, timeleft)
 	sword_special_timer[pname] = timeleft
 
@@ -57,7 +57,7 @@ minetest.register_tool("ctf_classes:sword_bronze", {
 		local pname = placer:get_player_name()
 		if not pointed_thing then return end
 
-		if (sword_special_timer[pname] and placer:get_player_control().sneak) then
+		if sword_special_timer[pname] and placer:get_player_control().sneak then
 			minetest.chat_send_player(pname, "You have to wait "..sword_special_timer[pname].."s to place marker again")
 
 			if pointed_thing.type == "node" then
--- a/mods/ctf/ctf_classes/melee.lua
+++ b/mods/ctf/ctf_classes/melee.lua
@@ -29,7 +29,7 @@ end, true)
 
 
 local sword_special_timer = {}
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_classes/melee.lua
```diff
@@ -29,12 +29,12 @@ end, true)
...
 local sword_special_timer = {}
-local SWORD_SPECIAL_COOLDOWN = 40
+local SWORD_SPECIAL_COOLDOWN = 10
 local function sword_special_timer_func(pname, timeleft)
 	sword_special_timer[pname] = timeleft
 
-	if timeleft - 10 >= 0 then
-		minetest.after(10, sword_special_timer_func, pname, timeleft - 10)
+	if timeleft - 1 >= 0 then
+		minetest.after(1, sword_special_timer_func, pname, timeleft - 1)
```
LoneWolfHT, line 37: This should be set to 2, it doesn't need to be run every second.

JostP: alright

LoneWolfHT:
Going from 20 seconds to 3 seems a bit drastic. How about 6 seconds?

JostP:
I can change to 6, but why we even have timer?
I mean you can place markers via chat infinite fast.

It is really annoying when you miss something with sword's marker and then wait 20 sec to place again.

LoneWolfHT:
Actually the current timeout should be 3 seconds. SWORD_SPECIAL_COOLDOWN is only used for the marker that loops through all online players

https://github.com/MT-CTF/capturetheflag/blob/4ef12801090219d9f1e9173b445ab607a72faab9/mods/ctf/ctf_classes/melee.lua#L72-L96

LoneWolfHT:
OH that's a bug
The interval is 10 seconds so you can't have any timeout less than that

LoneWolfHT:
Ah it was set to 20 before. Missed that somehow

JostP:
> OH that's a bug
> The interval is/was 10 seconds so you can't have any timeout less than that

Yes, I was wondering why timeout was 10 although I set it to 3 lol

mods/ctf/ctf_classes/melee.lua
JostP: Does running every second make more lag than running every 2 seconds?

LoneWolfHT:
'tis now 4 seconds and 20 seconds for the more resource intensive marker

PR #827, opened 2021-03-12
Code duplication cleanup

appgurueu:
Stop making fun of https://github.com/Gerold55/ma_pops_furniture/pull/6


appgurueu:
Note that this also registers indestructible variants of additional nodes, such as obsidian etc.

mods/ctf/ctf_map/ctf_map_core/nodes.lua
```diff
@@ -76,814 +76,47 @@ do
...
-			if nodedef.description then
-				nodedef.description = "Indestructible " .. nodedef.description
-			end
-			nodedef.groups = {immortal = 1}
-			minetest.register_node(":ctf_map:wool_" .. color, nodedef)
-			minetest.register_alias("ctf_map:" .. color, "ctf_map:wool_" .. color)
+for name, def in pairs(minetest.registered_nodes) do
+	local mod, nodename = name:match"(..-):(.+)"
+	local prefix = mod_prefixes[mod]
+	if nodename and prefix and name ~= "default:torch" and not (def.groups and (def.groups.immortal or def.groups.not_in_creative_inventory)) then
```
savilli, line 107: Why not poor torch tho?

appgurueu: The torch has to be handled a bit differently, I should probably add a comment to clarify that. See https://github.com/MT-CTF/capturetheflag/pull/827/files#diff-3d7b7717c70b400efd3cfb4bbec5c0131018fdf9e77cc516314ba65de5535b93R119-R122.

savilli: Does it really need to be handled differently? What happens if we remove name ~= "default:torch" with https://github.com/MT-CTF/capturetheflag/pull/827/files#diff-3d7b7717c70b400efd3cfb4bbec5c0131018fdf9e77cc516314ba65de5535b93R119-R122 as well?

appgurueu: > Does it really need to be handled differently? What happens if we remove name ~= "default:torch" with https://github.com/MT-CTF/capturetheflag/pull/827/files#diff-3d7b7717c70b400efd3cfb4bbec5c0131018fdf9e77cc516314ba65de5535b93R119-R122 as well?

It won't work then: Torches determine in on_place which variant to place. Accordingly, on_place has to be altered as well for torches.

mods/mtg/default/torch.lua
```diff
@@ -13,105 +13,85 @@ local function on_flood(pos, oldnode, newnode)
...
+				animation = {type = "vertical_frames", aspect_w = 16, aspect_h = 16, length = 3.3}
+		}},
+		paramtype = "light",
+		paramtype2 = "wallmounted",
+		sunlight_propagates = true,
+		walkable = false,
+		liquids_pointable = false,
+		light_source = 12,
+		groups = {choppy=2, dig_immediate=3, flammable=1, attached_node=1, torch=1},
+		drop = "default:torch",
```
appgurueu, line 70: register_torch should probably set the drop

mods/ctf/ctf_map/ctf_map_core/nodes.lua
savilli: I think it's better to rewrite on_place in such a way that it determines what to place by itself. For example we can use def.drop to construct wall and ceiling forms.

appgurueu: Sounds reasonable.

appgurueu:
* Torch variants don't appear in creative inventory anymore
* Torch `on_place` has been made "data-driven" following @savilli's suggestion
* Nodes with the `mortal` group or `buildable_to` don't get immortal variants

Regarding luacheck: The trailing whitespace is unrelated to this PR.

PR #828, opened 2021-03-14
Add "[Team Chat]" text to team chat and markers

DiamondPlane:


xenonca:
I think you can close this PR

PR #829, opened 2021-03-15
Fix running cooldown when changing to the same class

Panquesito7:
Things added/changed:

- Fix running cooldown when changing to the same class.

### DIFF
--- a/mods/ctf/ctf_classes/api.lua
+++ b/mods/ctf/ctf_classes/api.lua
@@ -76,9 +76,10 @@ function ctf_classes.set(player, new_name)
 	meta:set_string("ctf_classes:class", new_name)
 	ctf_classes.update(player)
 
-	if old_name ~= new_name then
-		ctf_classes.set_cooldown(player:get_player_name())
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
Does your inventory update with initial stuff when you switch to the same class?

Panquesito7:
> initial stuff when you switch to the same class

No, the inventory isn't modified. 🙂

mods/ctf/ctf_classes/api.lua
```diff
@@ -76,7 +76,9 @@ function ctf_classes.set(player, new_name)
 	meta:set_string("ctf_classes:class", new_name)
 	ctf_classes.update(player)
 
-	ctf_classes.set_cooldown(player:get_player_name())
+	if old_name ~= new_name then
+		ctf_classes.set_cooldown(player:get_player_name())
```
savilli, line 80: Just return if old_name == new_name. We don't need any of registered_on_changed callbacks if we don't change anything

PR #830, opened 2021-03-16
Prevent players to spam with empty messages in team chat

DiamondPlane:


mods/ctf/ctf_chat/init.lua
```diff
@@ -256,6 +256,10 @@ minetest.register_chatcommand("t", {
 				chatplus.log("<" .. name .. "> ** ".. param .. " **")
 			end
 
+			if param == "" then
+			minetest.chat_send_player(name, "You did not write anything, see /help t")
+			end
```
Panquesito7, line 261:
```suggestion
			if param == "" then
			    minetest.chat_send_player(name, "You did not write anything, see /help t")
			end
```

mods/ctf/ctf_chat/init.lua
```diff
@@ -256,6 +256,10 @@ minetest.register_chatcommand("t", {
 				chatplus.log("<" .. name .. "> ** ".. param .. " **")
 			end
 
+			if param == "" then
+			    minetest.chat_send_player(name, "You did not write anything, see /help t")
+			end
+
```
xenonca, line 262:
```suggestion
```
This will still send a message into the team chat. Instead add this to line 250:
```
if param == "" then
	return false, "-!- Empty team message, see /help t"
end
```

xenonca (changes requested):

mods/ctf/ctf_chat/init.lua
DiamondPlane: @xenonca Really? I checked and it didn't.

DiamondPlane: Edited! Thanks

mods/ctf/ctf_chat/init.lua
```diff
@@ -256,6 +256,10 @@ minetest.register_chatcommand("t", {
 				chatplus.log("<" .. name .. "> ** ".. param .. " **")
 			end
 
+			if param == "" then
+		return false, "-!- Empty team message, see /help t"
+			end
```
Panquesito7, line 261: Please adjust your code here.
```suggestion
			if param == "" then
		        return false, "-!- Empty team message, see /help t"
			end
```

mods/ctf/ctf_chat/init.lua
xenonca: I think you need to insert that code in line 250, else you'll be logging empty team messages without a reason.

xenonca: > I think you need to insert that code in line 250, else you'll be logging empty team messages without a reason.

:point_up:

xenonca:
I think you need to insert that code in line 250, else you'll be logging empty team messages without a reason.

Panquesito7 (approved):

xenonca (approved):

PR #841, opened 2021-03-23
Sprint gets reset to full on new game

Minetest-j45:
im not sure what the difference is between ```ctf_match.register_on_new_match(function())``` and ```ctf.register_on_new_game(function())``` so im not sure which i need to use(feel free to make suggestions if you know) but if this works, then great

### DIFF
--- a/mods/other/sprint/init.lua
+++ b/mods/other/sprint/init.lua
@@ -124,15 +124,15 @@ minetest.register_on_leaveplayer(function(player)
 end)
 
 ctf_match.register_on_new_match(function()
-	local players = minetest.get_connected_players()
-	for _, player in ipairs(players) do
+	local connected_players = minetest.get_connected_players()
+	for _, player in ipairs(connected_players) do
 		players[player:get_player_name()].stamina = STAMINA_MAX
 	end
 end)
 
 ctf.register_on_new_game(function()
-	local players = minetest.get_connected_players()
-	for _, player in ipairs(players) do
+	local connected_players = minetest.get_connected_players()
+	for _, player in ipairs(connected_players) do
 		players[player:get_player_name()].stamina = STAMINA_MAX
 	end
 end)
--- a/mods/other/sprint/init.lua
+++ b/mods/other/sprint/init.lua
@@ -122,3 +122,17 @@ end)
 minetest.register_on_leaveplayer(function(player)
 	players[player:get_player_name()] = nil
 end)
+
+ctf_match.register_on_new_match(function()
+	local players = minetest.get_connected_players()
+	for _, player in ipairs(players) do
+		players[player:get_player_name()].stamina = STAMINA_MAX
+	end
+end)
+
+ctf.register_on_new_game(function()
+	local players = minetest.get_connected_players()
+	for _, player in ipairs(players) do
+		players[player:get_player_name()].stamina = STAMINA_MAX
+	end
+end)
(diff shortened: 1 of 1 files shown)


mods/other/sprint/init.lua
```diff
@@ -122,3 +122,17 @@ end)
 minetest.register_on_leaveplayer(function(player)
 	players[player:get_player_name()] = nil
 end)
+
+ctf_match.register_on_new_match(function()
+	local connected_players = minetest.get_connected_players()
+	for _, player in ipairs(connected_players) do
+		players[player:get_player_name()].stamina = STAMINA_MAX
```
Panquesito7, line 129:
```suggestion
		player[player:get_player_name()].stamina = STAMINA_MAX
```

mods/other/sprint/init.lua
```diff
@@ -122,3 +122,17 @@ end)
...
+	end
+end)
+
+ctf.register_on_new_game(function()
+	local connected_players = minetest.get_connected_players()
+	for _, player in ipairs(connected_players) do
+		players[player:get_player_name()].stamina = STAMINA_MAX
```
Panquesito7, line 136:
```suggestion
		player[player:get_player_name()].stamina = STAMINA_MAX
```

Minetest-j45:
wait what... no? im confused

Minetest-j45:
see line 119, how they reset it on respawn... it is ```players[player:get_player_name()].stamina```, not ```player[player:get_player_name()].stamina```

mods/other/sprint/init.lua
Panquesito7, line 136: Where is the variable/parameter `players` defined?

mods/other/sprint/init.lua
Panquesito7, line 129: Same here.

mods/other/sprint/init.lua
Minetest-j45: line 22

mods/other/sprint/init.lua
Minetest-j45: same answer as above

appgurueu:
Two things:
1. Is `register_on_new_game` needed? Why is `register_on_new_match` insufficient?
2. If both are needed, make your function - which is the same for both - a named local one and pass the same function to both instead of copying & pasting

Minetest-j45:
See my first comment i made, when i opened this, as i said... I dont know the difference between them so i out both. If you know, feel free to change my pr

Anthony-De:
@Minetest-j45 
`register_on_new_game` only ever used on server start.
You only need to use `register_on_new_match`.

mods/other/sprint/init.lua
```diff
@@ -122,3 +122,17 @@ end)
...
+	end
+end)
```
Anthony-De, line 138:
```suggestion
minetest.after(0, function()
	ctf_match.register_on_new_match(function()
		for _, player in pairs(players) do
			player.stamina = STAMINA_MAX
		end
	end)
end)
```

Anthony-De (changes requested):

Anthony-De:
There is a little issue, the hudbar doesn't update till the player starts sprinting.

Panquesito7:
> update till the player starts sprinting

That should be fixed in another PR, I think.

mods/other/sprint/init.lua
appgurueu: `minetest.after(0, ...)` should not be used, use `minetest.register_on_mods_loaded(...)` instead. Why is it needed here?

Anthony-De: Because ctf_match isn't defined when this file runs.

Minetest-j45:
is this (with AnthonyDE's changes) ready to merge?

Minetest-j45:
oops. i accidentally closed this PR and didnt notice

Panquesito7 (approved):

PR #842, opened 2021-03-26
Add C/A

moreavy:


xenonca:
Please provide a proper description of what this PR does, it also adds an optional C/A requirement for the pro chest section.
I still believe that C/A will more likely discourage to even attempt to get the flag -> discourage capturing in general and doesn't really reflect any player skills.

moreavy:
Capturing is EZ, and _**Pro Players**_ should be able to capture the flag in at least 4 attempts.

Calinou:
I also think people should never be penalized for *attempting* to capture the flag, even if they always fail. Please don't encourage making the game campy :slightly_smiling_face:

LoneWolfHT:
Closing as I'm only seeing self-support

PR #844, opened 2021-03-27
place_limit: Use HUD notifications & quota instead of time between placements

appgurueu:
Makes place_limit use HUD notifications & a quota.


appgurueu:
Two changes to place_limit in one PR, in order to avoid merge conflicts.

LoneWolfHT (changes requested):
maps submodule is changed by this

PR #849, opened 2021-03-31
Announce bounties only to the enemy team(s)

appgurueu:
Relaying such an announcement to the bountied team ~~would~~ might be considered sabotage.


LoneWolfHT:
Ignoring this:
> Relaying such an announcement to the bountied team would might be considered sabotage.

PR #852, opened 2021-04-01
Adds TNT from mtg

Anthony-De:
I took the code from this PR then tweaked it till it worked, then optimized it for CTF usage.
https://github.com/MT-CTF/capturetheflag/pull/558


DiamondPlane:
TNT would be bad for CTF :/

RyanZem:
It will make holes in the maps, we already have players who is making holes, so no need TNT.

Calinou:
I think the only way TNT could be viable is by making it non-craftable and only findable as a special item in chests. This would make it rare and therefore hard to abuse.

Also, it shouldn't be able to destroy protected items in a base.

PR #854, opened 2021-04-06
Immortal map nodes: Switch to tool group blacklist

appgurueu:
Originally intended as a fix for the non-connecting fences reported, this switches to a tool group blacklist instead of whitelisting node-only groups (which missed groups such as the fence one).


appgurueu:
Changes contained in #856

PR #870, opened 2021-04-25
Medkits: Don't give back if map is skipped

appgurueu:


savilli:
ctf.register_on_new_game?

appgurueu:
> ctf.register_on_new_game?

on_new_match I think, but thanks for the pointer! Should have read the docs.

savilli:
> > ctf.register_on_new_game?
> 
> on_new_match I think. Thanks for the pointer! Should have read the docs.

Currently it's the same but on_new_match makes more sense indeed

savilli:
ctf_match.register_on_new_match

PR #871, opened 2021-04-26
Prevent teammates from knowing who no longer has the bounty when the bountied player is disconnected and isn't in a team

Panquesito7:
Things added/changed:

- Prevent teammates from knowing who no longer has the bounty
when the bountied player is disconnected and isn't in a team.
  - Didn't test, but it should work properly. The code was taken from this piece of code:
  https://github.com/MT-CTF/capturetheflag/blob/7733c2cb6786e8657bcc797712a9e132efdf6d73/mods/ctf/ctf_bounties/init.lua#L5-L8

### DIFF
--- a/mods/ctf/ctf_bounties/init.lua
+++ b/mods/ctf/ctf_bounties/init.lua
@@ -24,6 +24,7 @@ local function announce_all()
 end
 
 local function bounty_player(target)
+	local bountied = ctf.player(bountied_player)
 	local prev = bountied_player
 	bountied_player = target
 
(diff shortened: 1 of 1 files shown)


savilli:
But why teammates shouldn't know about who no longer has the bounty? It doesn't matter at that point

Panquesito7:
> shouldn't know about who no longer has the bounty

Currently, if my teammate gets a bounty, I'm not able to see who no longer has the bounty, nor who has the bounty now. If somebody loses their bounty (and they're not in any team, or they left), I shouldn't be able to see that. Hope I get to understand.

savilli:
> Currently, if my teammate gets a bounty, I'm not able to see who no longer has the bounty

It's not true. You can see who had the bounty

Panquesito7:
> You can see who had the bounty

This PR is intentionally for players who left and are not in a team. Also, see https://github.com/MT-CTF/capturetheflag/pull/849.

savilli:
> > You can see who had the bounty
> 
> This PR is intentionally for players who left and are not in a team. Also, see #849.

You don't say anything new. What problem are you trying to solve anyway?

PR #875, opened 2021-04-30
Spectator Privacy

xenonca:
- Do not show spectators in join messages (leave messages are handled by the spectator mod itself)
- Do not show spectators in /team or /status
Other relevant PR: [here](https://github.com/MT-CTF/servermods/pull/23)

### DIFF
--- a/mods/ctf/ctf/teams.lua
+++ b/mods/ctf/ctf/teams.lua
@@ -218,8 +218,12 @@ function ctf.join(name, team, force, by)
 			" has joined team " .. minetest.colorize(tcolor, team)
 	end
 
-	minetest.chat_send_all("*** " .. join_msg)
-	minetest.log("action", name .. " joined team " .. team)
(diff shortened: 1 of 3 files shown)


Lucyucy (approved):

savilli:
Why just don't add spectators to teams? In case there is other code with ctf.team()

xenonca:
AFAIK joining players must be allocated to a team, not sure though

savilli:
ctf.player(name).team can be nil. It can cause some crashes tho, but we need to fix them anyways

xenonca:
I'm not really familiar with that code, I suppose those fixes should go into a different PR

Emojigit:
You can also add a "spectator team" which does nothing with flags and the gameplay

xenonca:
I assume at this point it's better to wait for CTF 3.0 to be released and rework this...

Panquesito7 (approved):

xenonca:
incompatible with v3

PR #878, opened 2021-05-06
Rework melee damage

savilli:
Introduce get_damage_modifier, change player hp in register_on_punchplayer.
Avoid using register_on_player_hpchange callback as it doesn't account kill assists correctly.

### DIFF
--- a/mods/ctf/ctf_classes/melee.lua
+++ b/mods/ctf/ctf_classes/melee.lua
@@ -1,33 +1,3 @@
-minetest.register_on_player_hpchange(function(player, hp_change, reason)
-	if reason.type ~= "punch" or not reason.object or not reason.object:is_player() then
-		return hp_change
-	end
-
-	local class = ctf_classes.get(reason.object)
-
-	if class.properties.melee_bonus and reason.object:get_wielded_item():get_name():find("sword") then
-		local change = hp_change - class.properties.melee_bonus
-
-		if player:get_hp() + change <= 0 and player:get_hp() + hp_change > 0 then
-			local wielded_item = reason.object:get_wielded_item()
-
-			for i = 1, #ctf.registered_on_killedplayer do
-				ctf.registered_on_killedplayer[i](
-					player:get_player_name(),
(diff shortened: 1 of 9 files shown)


mods/ctf/ctf_classes/melee.lua
```diff
@@ -1,32 +1,18 @@
...
-
-			for i = 1, #ctf.registered_on_killedplayer do
-				ctf.registered_on_killedplayer[i](
-					player:get_player_name(),
-					reason.object:get_player_name(),
-					wielded_item,
-					wielded_item:get_tool_capabilities()
-				)
-			end
+	if class.properties.melee_bonus and hitter:get_wielded_item():get_name():find("sword") then
```
LoneWolfHT, line 6: class.properties.melee_bonus should be used to control by how much damage is increased (Problem from my old code you ported). Should also support negative values too in case we want to decrease medic melee damage slightly.

mods/ctf/ctf_classes/melee.lua
```diff
@@ -1,32 +1,18 @@
...
+		if time_from_last_punch > 1 then
+			time_from_last_punch = 1
+		elseif time_from_last_punch < 0.5 then
+			time_from_last_punch = 0.5
```
LoneWolfHT, line 10: Would be nice if we could get this code to penalize serious spam clicking, I think it does the opposite right now, something I thought I fixed at some point, maybe not though

mods/ctf/ctf_classes/melee.lua
savilli: Decreasing damage is troublesome. Perhaps we should reject minetest engine calculation and change hp by ourselves then

LoneWolfHT: Don't need it if it's too hard to do, at a first look I'd assume negative damage would work though

savilli: Negative damage would break a lot of things for sure

PR #880, opened 2021-05-08
Fix ctf_respawn_delay ghost mode bug

savilli:
Players could turn into ghosts under certain conditions


mods/ctf/ctf_traps/init.lua
```diff
@@ -63,7 +63,10 @@ minetest.register_node("ctf_traps:damage_cobble", {
 			if placerobj then
 				digger:punch(placerobj, 10, {damage_groups = {fleshy = 7}}, vector.new(0, 1, 0))
 			else
-				digger:set_hp(digger:get_hp() - 7)
+				local hp = digger:get_hp()
+				if hp > 0 then
+					digger:set_hp(digger:get_hp() - 7)
```
savilli, line 68: We still need it because otherwise death would be counted twice

PR #882, opened 2021-05-14
Make all players use the same smoke texture

appgurueu:
Uses a `dynamic_media_add`-based hack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# Dynamic textures
+mods/pvp/grenades/dynamic_textures/*
 ## Generic ignorable patterns and files
 *~
 .*.swp
--- a/mods/pvp/grenades/grenades.lua
+++ b/mods/pvp/grenades/grenades.lua
@@ -84,6 +84,29 @@ grenades.register_grenade("grenades:frag", table.copy(fragdef))
 
 -- Smoke Grenade
 
+-- HACK to ensure all players use the same texture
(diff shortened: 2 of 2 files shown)


LoneWolfHT:
What does this look like when you use a transparent smoke texture?

Lucyucy:
With a transparent smoke texture the smoke is invisible

appgurueu:
> With a transparent smoke texture the smoke is invisible

Guess I'll have to throw in a `[noalpha` then...

Lucyucy:
https://user-images.githubusercontent.com/80708819/118393319-dbfdc100-b63e-11eb-8f3f-8dabc2125499.png

Looks like this now

appgurueu:
Alternative is just appending a `[makealpha` texture modifier. That works, but exposes ugly Minetest Z-sorting bugs. Would prefer this.

DiamondPlane:
I tested it. Hmm, it looks now like this:
screenshot

appgurueu:
> I tested it. Hmm, it looks now like this:
> screenshot

How it looks depends on the current smoke texture.

appgurueu:
@savilli I don't see how this has become irrelevant.

appgurueu:
> I don't think we should use a hack. If anything [6d5f046](https://github.com/MT-CTF/capturetheflag/commit/6d5f046d41c28e7243876f6ab349b0710ed51d22) approach should be used.

The mentioned approach is heavily limiting though.

PR #881, opened 2021-05-14
Increase paxel digging limit

Lucyucy:
Increase paxel digging limit to 50

PR #883, opened 2021-05-14
Paxel: Ignore vertical distance, lower dist. limit

appgurueu:
An alternative to #881


LoneWolfHT:
The pillar removal doesn't go sideways AFAIK so you can just remove all distance related code

appgurueu:
> The pillar removal doesn't go sideways AFAIK so you can just remove all distance related code

Wouldn't this incentivize just paxeling and then running away? I thought it was intended that you stand close to the tower you are paxeling.

LoneWolfHT:
Oh forgot about that

PR #890, opened 2021-05-24
Disable Towering While Fighting

air-glitch:
I am `disco-dancer` in-game. I would just like to disable towering while fighting because many noobs tower up when I am killing them.

### DIFF
--- a/mods/pvp/anticoward/init.lua
+++ b/mods/pvp/anticoward/init.lua
@@ -181,7 +181,7 @@ minetest.register_on_punchplayer(function (player, hitter, time_from_last_punch,
 end)
 
 minetest.register_on_placenode(function(pos, newnode, placer, oldnode, itemstack, pointed_thing)
-	if no_towering[placer:get_player_name()] then
+	if no_towering[placer:get_player_name()] and (placer:get_pos().y = (pos.y + 1)) then
 		minetest.set_node(pos, oldnode)
 	end
 end)
--- a/mods/pvp/anticoward/init.lua
+++ b/mods/pvp/anticoward/init.lua
@@ -169,3 +169,19 @@ ctf.register_on_new_game(function()
 	end
 	potential_cowards = {}
 end)
+
+--disable towering while fighting
+local no_towering = {}
+
+minetest.register_on_punchplayer(function (player, hitter, time_from_last_punch, tool_capabilities, dir, damage)
+	no_towering[player:get_player_name()] = true
+	minetest.after(5, function ()
+		no_towering[player:get_player_name()] = false
+	end)
+end)
+
+minetest.register_on_placenode(function(pos, newnode, placer, oldnode, itemstack, pointed_thing)
+	if no_towering[placer:get_player_name()] then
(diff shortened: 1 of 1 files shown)


moreavy (approved):
Test: Working, Great Job :+1:

appgurueu:
1. This belongs in the place_limit mod IMO
2. Do not call remove node, but set the old node instead. Have a look at how `place_limit` does it. Also, `return true` for the node to not be wasted.

moreavy:
This code disables towering while fighting, but actually, it also disable blocking the way in tunnels too while it shouldn't.

air-glitch:
May I know on which axis (`x`, `y` or `z`) do people tower up?

appgurueu:
> May I know on which axis (`x`, `y` or `z`) do people tower up?

y, of course?

mods/pvp/anticoward/init.lua
```diff
@@ -181,7 +181,7 @@ minetest.register_on_punchplayer(function (player, hitter, time_from_last_punch,
 end)
 
 minetest.register_on_placenode(function(pos, newnode, placer, oldnode, itemstack, pointed_thing)
-	if no_towering[placer:get_player_name()] then
+	if no_towering[placer:get_player_name()] and (placer:get_pos().y = (pos.y + 1)) then
```
appgurueu: Syntax error, I assumed you intended `==`? Anways, this won't work as expected; you'd have to do a range check for that.

appgurueu (changes requested):

mods/pvp/anticoward/init.lua
air-glitch: I've already fixed `==` bug.

moreavy:
> > May I know on which axis (`x`, `y` or `z`) do people tower up?
> 
> y, of course?

Wrong, `z` axis is the one where players tower up.

mods/pvp/anticoward/init.lua
```diff
@@ -169,3 +169,19 @@ ctf.register_on_new_game(function()
...
+
+minetest.register_on_punchplayer(function (player, hitter, time_from_last_punch, tool_capabilities, dir, damage)
+	no_towering[player:get_player_name()] = true
+	minetest.after(5, function ()
+		no_towering[player:get_player_name()] = false
+	end)
+end)
+
+minetest.register_on_placenode(function(pos, newnode, placer, oldnode, itemstack, pointed_thing)
+	if no_towering[placer:get_player_name()] and (placer:get_pos().y == (pos.y + 1)) then
```
moreavy, line 184:
```suggestion
	if no_towering[placer:get_player_name()] and (placer:get_pos().z == pos.z) then
``` I think this should work.

Comment (moreavy) [inline mods/pvp/anticoward/init.lua:184]:
```suggestion
	if no_towering[placer:get_player_name()] and ((math.floor(placer:get_pos().y) == (pos.y + 1)) or (math.ceil(player:get_pos().y) == (pos.y + 1)) then
```

PR #893, opened 2021-06-06
Reset sprint on new match

### LINKED ISSUE #892: Sprint should reset between matches
Currently, the amount of sprint a player has left is not reset to full between matches. I believe this would be logical to change, so that upon starting a new game players have full sprint capability..

Panquesito7:
Things added/changed:

- Reset sprint on new match
  - Closes #892.


LoneWolfHT:
Ah, it's currently only reset on respawn

PR #896, opened 2021-06-12
Add a thing that shows the player's HP that you healed

RyanZem:
Example:
healing

### DIFF
--- a/mods/ctf/ctf_bandages/init.lua
+++ b/mods/ctf/ctf_bandages/init.lua
@@ -48,7 +48,7 @@ minetest.register_craftitem("ctf_bandages:bandage", {
 				hud_event.new(pname, {
 					name  = "ctf_bandages:heal",
 					color = 0xC1FF44,
-					value = name .. " healed you!",
+					value = name .. " healed you!\nYour HP:"..object:get_hp(),
 				})
 			end
 		else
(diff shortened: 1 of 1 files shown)


appgurueu:
Pretty sure you must keep it on the same line or it won't work (events work with hardcoded offsets of a single line, I think). If you really need the newline, you should probably add a second HUD event.

LoneWolfHT:
> Pretty sure you must keep it on the same line or it won't work (events work with hardcoded offsets of a single line, I think). If you really need the newline, you should probably add a second HUD event.

It looks like it's working in the screenshot?

appgurueu:
> > Pretty sure you must keep it on the same line or it won't work (events work with hardcoded offsets of a single line, I think). If you really need the newline, you should probably add a second HUD event.
> 
> It looks like it's working in the screenshot?

The screenshot only shows a single event.

RyanZem:
> > > Pretty sure you must keep it on the same line or it won't work (events work with hardcoded offsets of a single line, I think). If you really need the newline, you should probably add a second HUD event.
> > 
> > 
> > It looks like it's working in the screenshot?
> 
> The screenshot only shows a single event.

I edited as you asked.

RyanZem:
Example: 
s

xenonca:
I don't see a need for this tbh, you can always have a look at your hearts/hp in your HUD.

PR #899, opened 2021-06-17
Update random messages

RyanZem:
Removed "Hitting your enemy does more damage than not hitting them." because this message is not needed


Panquesito7 (approved):

LoneWolfHT:
It's comic relief :(

wwar-XeStro:
It is needed, maybe not for us but for new player, how will they know which one does more damage? Also i think we need a "damaging your enemies is better than not damaging them" message so they will understand the basics

PR #901, opened 2021-06-19
Add antiwalkthrough mod

appgurueu:
This prevents players from freely glitching through blocks.

### DIFF
--- a/mods/pvp/antiwalkthrough/init.lua
+++ b/mods/pvp/antiwalkthrough/init.lua
@@ -3,14 +3,14 @@ local collisionbox = {
     type = "fixed",
     fixed = {}
 }
-local fixed = collisionbox.fixed
+local boxes = collisionbox.fixed
 for i = 1, 3 do
-    fixed[i] = {-0.5, -0.5, -0.5, 0.5, 0.5, 0.5}
-    fixed[i][i] = 0.49
-    fixed[i+3] = {-0.5, -0.5, -0.5, 0.5, 0.5, 0.5}
-    fixed[i+3][i+3] = -0.49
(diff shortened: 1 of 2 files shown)


DiamondPlane:
It works good but I think teammates can trap teammates, like if you can't dig that node or don't have a pickaxe.

appgurueu:
> It works good but I think teammates can trap teammates, like if you can't dig that node or don't have a pickaxe.

Well I could disable placing blocks where teammates - excluding the placer - stand.

mods/pvp/antiwalkthrough/init.lua
```diff
@@ -0,0 +1,86 @@
...
+-- Item entity must be provided with our copy of the registered nodes table, which provides regular collision box definitions
+local def = minetest.registered_entities["__builtin:item"]
+local on_step = def.on_step
+function def.on_step(...)
+    -- HACK temporarily replace the registered_nodes table
+    minetest.registered_nodes = original_nodes
+    on_step(...)
+    minetest.registered_nodes = registered_nodes
+end
+minetest.register_entity(":__builtin:item", def)
```
Panquesito7, line 86: Missing an endline. This could cause problems with some OS'ses and IDEs.

```suggestion
minetest.register_entity(":__builtin:item", def)
```

Panquesito7 (changes requested):
Maybe provide a `README.md` file explaining what the mod does and a license file, too.

appgurueu:
This could have been merged into v3 perfectly fine.

savilli:
We have decided we don't need that

appgurueu:
> We have decided we don't need that


Who are "we"?

PR #905, opened 2021-06-24
Make stairs diggable with paxel

Lucyucy:
Adds cobble stairs and desert cobble stairs to local function "isdiggable" and "remove_pillar"

### DIFF
--- a/mods/ctf/ctf_classes/medic.lua
+++ b/mods/ctf/ctf_classes/medic.lua
@@ -150,9 +150,7 @@ local function isdiggable(name)
 		name:find("glass" ) or name:find("ice"  ) or
 		name:find("snow"  )
 	)
-	or name:find("stairs:") and (
-		name:find("stair_cobble") or name:find("stair_desert_cobble")
-	)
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_classes/medic.lua
```diff
@@ -150,6 +150,9 @@ local function isdiggable(name)
 		name:find("glass" ) or name:find("ice"  ) or
 		name:find("snow"  )
 	)
+	or name:find("stairs:") and (
+		name:find("stair_cobble") or name:find("stair_desert_cobble")
```
Panquesito7, line 154: Aren't there any other stairs used in maps besides these two?

mods/ctf/ctf_classes/medic.lua
```diff
@@ -150,6 +150,9 @@ local function isdiggable(name)
...
+	)
```
LoneWolfHT, line 155:
```suggestion
	or name:find("stairs:")
```

PR #913, opened 2021-07-17
add illumination mod

ksandronline:
I think it will be very useful to be able to light up the space around you. It will be especially convenient to use it in tunnels and underground maps. And also at night in the forest. I'm playing with the standard brightness settings. And my nights are very dark, almost nothing is visible. But it's easy to steal a flag on a dark night. When an enemy object is in your hands, everyone already knows where you are, so it is advantageous to glow around you, it is not dangerous that your enemies will notice you. And run fast, seeing the road well.


ksandronline:
Снимок экрана_2021-07-17_09-35-54

PR #914, opened 2021-07-17
add light_source team doors

ksandronline:
light_source = 3, as chest


DiamondPlane:
Why should the door glow?

LoneWolfHT:
Doors do not need to glow

PR #915, opened 2021-07-24
enhancement join message

ksandronline:
helloip

I have seen a lot of dissatisfied reviews that join and leave messages have become similar. I suggest changing the join message by adding new functionality to it.


RyanZem:
"All checks have failed"

ksandronline:
RyanZem: yes im see thanks. I don't think that this will be added to the shared server.

RyanZem:
You can close this PR already.

PR #924, opened 2021-08-29
optionally use 9 sliced background (fix #862)

### LINKED ISSUE #862: Optionally use 9-slice background textures
I noticed while running tests on my texture pack that ctf:
1. Uses the same background texture as Minetest Game
2. Does not use 9-slice backgrounds in formspecs, _unlike_ new Minetest Game releases

This means that backgrounds in texture packs targeting newer MTG releases can break when used in CTF... MTG already has a switch for backwards-compatible background handling, so it would be nice to see 9-slice textures offered here as well.

AFCMS:
@Df458
fix #862

### DIFF
--- a/mods/mtg/default/init.lua
+++ b/mods/mtg/default/init.lua
@@ -10,10 +10,22 @@ default.LIGHT_MAX = 14
 
 -- GUI related stuff
 minetest.register_on_joinplayer(function(player)
-	player:set_formspec_prepend([[
-			bgcolor[#080808BB;true]
-			background[5,5;1,1;gui_formbg.png;true]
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
I think this is fixed in CTFv3, which only changes MTG code via external mods so it can be kept up-to-date

AFCMS:
> I think this is fixed in CTFv3, which only changes MTG code via external mods so it can be kept up-to-date

Off-topic and dumb question, but where is the CTFv3 repo?

LoneWolfHT:
> Off-topic and dumb question, but where is the CTFv3 repo?

Little hidden but https://github.com/LoneWolfHT/capturetheflag/tree/3.0

(Note that you will probably need to clone it into a separate MT install, it's a 90% rewrite)

PR #925, opened 2021-09-03
Revert "Rework melee damage"

LoneWolfHT:
Reverts MT-CTF/capturetheflag#878


Calinou:
Out of curiosity, what was wrong with the rework?

LoneWolfHT:
Server is still on old engine code, we're waiting until ruben can update it

PR #933, opened 2021-11-28
Remove pick_stone from initial_stuff

RyanZem:
Stone pickaxe is not needed because it will give us 2 pickaxes and we don't need stone pickaxe because steel pickaxe already exists.


savilli:
It's not guaranteed that steel pickaxe will be available at every map. Better to handle that with item tiers.

PR #940, opened 2021-11-30
Update random messages

RyanZem:


Panquesito7 (approved):

mods/other/random_messages/init.lua
```diff
@@ -57,6 +57,10 @@ function random_messages.read_messages()
 		"Use /summary (Or /s) to check scores of the current match and the previous match.",
 		"Strengthen your team by capturing enemy flags.",
 		"Hitting your enemy does more damage than not hitting them.",
+		"Use /top50 command to see the leaderboard.",
+		"Use /top50 <mode:technical modename> to see the leaderboard on another mode. For example: /top50 mode:classes.",
+		"To check someone's rank on another mode use /r <mode:technical modename> <playername>."
+		.. "For example: /r mode:classic randomplayer.",
```
savilli, line 63: I think we should use nade_fight mode in the examples, because it's the most confusing.

RyanZem: Added.

appgurueu:
This should not have been merged.
1. The messages duplicate chatcommand help;
2. There should be a single message instead informing users about /help (ideally this should happen in some kind of tutorial though);
3. The last message misses a space between the first and second sentence, hidden by the string concat (it is way too long and shouldn't have been included, actually).

PR #948, opened 2021-12-06
Add desert and cobble stairs back to the crafting

RyanZem:


savilli:
Why?

RyanZem:
> Why?

You can make with it better base and make stairs like to a bridge or something.

PR #950, opened 2021-12-08
Add cobble, wood, torches and team door to the team chest

RyanZem:
(there no ctf_traps mod so I can't add it)

### DIFF
--- a/mods/ctf/ctf_teams/team_chest.lua
+++ b/mods/ctf/ctf_teams/team_chest.lua
@@ -64,9 +64,9 @@ for _, chest_color in pairs(colors) do
 		inv:set_size("pro", 4 * 7)
 		inv:set_size("helper", 1 * 1)
 		inv:add_item("main", ItemStack("default:cobble 99"))
-       	inv:add_item("main", ItemStack("default:wood 99"))
-       	inv:add_item("main", ItemStack("default:torch 30"))
-       	inv:add_item("main", ItemStack("ctf_teams:door_steel 3"))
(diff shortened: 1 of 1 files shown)


RyanZem:
> Why?

Many players want it and it was suggested. The reason is that we can build our base by not even trying to find something in treasure chests, just by opening the team chest. It's actually a PR so it's ok if you don't want to merge it. :)

savilli:
> > Why?
> 
> Many players want it and it was suggested. The reason is that we can build our base by not even trying to find something in treasure chests, just by opening the team chest. It's actually a PR so it's ok if you don't want to merge it. :)

Where was it suggested?

RyanZem:
> > > Why?
> > 
> > 
> > Many players want it and it was suggested. The reason is that we can build our base by not even trying to find something in treasure chests, just by opening the team chest. It's actually a PR so it's ok if you don't want to merge it. :)
> 
> Where was it suggested?

here: https://discord.com/channels/447819017391046687/448136308108296213/917840446011613184 but I already coded it before it was suggested.

appgurueu:
Not happy with the current state of this:

1. Should be configurable, not "hardcoded"
2. Why did you put that in the code, at that place? Where is it handled currently?

savilli:
We've decided it should be per-mode

savilli (changes requested):
We've decided it should be per-mode

savilli:
Closed in favor of 9f0186e8bf9f81a208a727cc90dd01ae1b1e42ff

PR #952, opened 2021-12-09
Move "winner" and "duration" texts in the summary

BrownHedgehog78:
"Fixes" this:
image

### DIFF
--- a/mods/ctf/ctf_modebase/summary.lua
+++ b/mods/ctf/ctf_modebase/summary.lua
@@ -224,7 +224,7 @@ function ctf_modebase.summary.show_gui_sorted(name, rankings, special_rankings, 
 	if formdef.winner then
 		formspec.elements.winner = {
 			type = "label",
-			pos = {"center", 0.5},
+			pos = {4, 0.5},
 			label = formdef.winner,
 		}
 	end
--- a/mods/ctf/ctf_modebase/summary.lua
+++ b/mods/ctf/ctf_modebase/summary.lua
@@ -224,7 +224,7 @@ function ctf_modebase.summary.show_gui_sorted(name, rankings, special_rankings, 
 	if formdef.winner then
 		formspec.elements.winner = {
 			type = "label",
-			pos = {4, 0.5},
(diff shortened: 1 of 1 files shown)


savilli:
How does it look after the change?

BrownHedgehog78:
> How does it look after the change?

image

savilli:
How about you move "winner" to the left? I don't like free space there.

BrownHedgehog78:
> How about you move "winner" to the left? I don't like free space there.

Like it was before?

savilli:
> > How about you move "winner" to the left? I don't like free space there.
> 
> Like it was before? image

That looks good

Panquesito7:
> > How does it look after the change?
> 
> image

For me, it looks better like this. The winner text being on the left leaves too much space on the right which makes it look bad, IMO.

BrownHedgehog78:
uh, where should it be then?

mods/ctf/ctf_modebase/summary.lua
```diff
@@ -232,7 +232,7 @@ function ctf_modebase.summary.show_gui_sorted(name, rankings, special_rankings,
 	if formdef.duration then
 		formspec.elements.duration = {
 			type = "label",
-			pos = {"center", 0.5},
+			pos = {14.5, 0.5},
```
LoneWolfHT, line 235: I think you can use "right" here

LoneWolfHT:
> > How does it look after the change?
> 
> image
> 

Is that middle element using `"center"`?

BrownHedgehog78:
> > > How does it look after the change?
> > 
> > 
> > image
> 
> Is that middle element (Player x captured) using `"center"`?

Yes

mods/ctf/ctf_modebase/summary.lua
BrownHedgehog78: No, it crashes the server

savilli:
Closed in favor of d80f2f1a00b51e43cd8c35d1298714b636e97d7a

PR #982, opened 2022-01-31
Sprint: Show "gone" background icons

appgurueu:
Does what it says on the tin. `sprint_stamina_icon_gone.png` may be removed if this is to be only a customization for TPs.


appgurueu:
> This is adding submodules, I think your local copy still has some v2 stuff

Fixed.

PR #987, opened 2022-02-07
Classes Rework P.1

LoneWolfHT:
* New class select GUI
* New ctf_gui API
* Rework for scouts
* fix damage cobble not damaging shooters


mods/apis/ctf_gui/dev.lua
```diff
@@ -0,0 +1,39 @@
...
+
+	assert(file, error)
+
+	file:write(formspec)
+
+	file:close()
+
+	local function interval()
+		if formspec:sub(1, 3) == "[f]" then
+			local result, form = pcall(loadstring(formspec:sub(4)), formcontext)
```
appgurueu, line 17: There is absolutely no need to use loadstring here?

LoneWolfHT: Does `pcall` run strings as lua code?

LoneWolfHT: [f] indicates you are writing lua code that returns the formspec string

appgurueu: > Does `pcall` run strings as lua code?

`pcall` runs a function. When you call `loadstring`, you are "compiling" Lua code.

appgurueu:
I recommend you to have a look at [Epidermis formspec implementation](https://github.com/appgurueu/epidermis/blob/master/formspec.lua) for inspiration; it's simple yet effective way to show a formspec to a player with a callback attached to it.

Note: There is no need to use an explicit `context[playername]` table if you use my approach; all upvalues to the callback will simply be the "context". This is both faster and more comfortable to use.

mods/apis/ctf_gui/dev.lua
appgurueu: > [f] indicates you are writing lua code that returns the formspec string

Then you should instead allow passing/storing a function that does this. Passing Lua code around as prefixed strings and compiling it upon execution is bad for both performance and code quality.

LoneWolfHT: This is for live formspec editing, not release use (see the non _dev function for that).
I'll make this clearer.

appgurueu: > This is for live formspec editing, not release use (see the non _dev function for that). I'll make this clearer.

I wouldn't want this for live formspec editing either. You'll have some kind of Lua prompt and should just be entering the function there. Having to put it in a string is annoying.

LoneWolfHT: Down below it creates `file_edit.txt` in the world directory. It loads that file as a string every second, converts it to lua code if that's what it says it is, and runs that lua code for the formspec, which it then displays

LoneWolfHT:
*The one time I was too lazy to diff my changes against the main remote*

PR #997, opened 2022-03-16
Establish vote count ordering

appgurueu:
Currently vote counts aren't ordered. `pairs` traversal means any order is possible.

This is "fixed" here. Vote counts are output in ascending order. It does so by iterating over the lengths in the correct order and checking whether there are votes. Not an issue for a max of 5 rounds. Sorting would likely be less efficient and more of a hassle here.

Needs testing.


savilli:
LGTM

PR #998, opened 2022-03-19
Fix vote count ordering: Include votes for 0

appgurueu:
Off by one error: 0 [votes] has to be included


savilli:
I like the things without fix

appgurueu:
> I like the things without fix

Yeah sorry, I missed this.

PR #999, opened 2022-03-20
Fix void grenade allowing player through barrier

Zardshard:


savilli:
Thanks for your contribution!
I think we should use a more generic approach. We should disallow using void nades to cross any wall, for example, the barrier in the wall map.

savilli:
Closed in favor of 31a30905098bae71fcbc30b9c9d06c4daa91b27b

PR #1003, opened 2022-04-27
Allow maps to specify a spawn direction for each team

grorp:
screenshot of the GUI

### Context

This is an alternative to #1002. #1002 produces bad results on some maps:

screenshot of a bad spawn direction 1
screenshot of a bad spawn direction 2

Since there's probably no algorithm that works for all maps, I think it's better to specify the spawn directions manually.

### DIFF
--- a/mods/ctf/ctf_map/mapedit_gui.lua
+++ b/mods/ctf/ctf_map/mapedit_gui.lua
@@ -414,7 +414,55 @@ function ctf_map.show_map_save_form(player, scroll_pos)
 				end)
 			end,
 		}
-		idx = idx + 1.5
+		idx = idx + (0.7 / 2) + 0.3 + (0.3 / 2)
+
+		local look_pos = context[player].teams[teamname].look_pos
+
+		elements[teamname .. "_look_pos"] = {
+			type = "label",
+			label = "Look position: " .. (look_pos and vector.to_string(look_pos) or "auto"),
+			pos = {0.2, idx},
+			-- "The first line of text is now positioned centered exactly at the position specified."
+			-- https://github.com/minetest/minetest/blob/480d5f2d51ca8f7c4400b0918bb53b776e4ff440/doc/lua_api.txt#L2929
+		}
+		idx = idx + (0.3 / 2) + 0.1 + (0.7 / 2)
+
+		local btn_width = (9 - (ctf_gui.SCROLLBAR_WIDTH + 0.1) - 0.1) / 2
+
+		elements[teamname.."_look_pos_auto"] = {
+			type = "button",
+			label = "Auto",
+			pos = {0.2, idx - (ctf_gui.ELEM_SIZE.y / 2)},
+			size = {btn_width, ctf_gui.ELEM_SIZE.y},
+
+			func = function(pname, fields)
+				context[pname].teams[teamname].look_pos = nil
+
+				minetest.after(0.1, function()
+					ctf_map.show_map_save_form(pname, minetest.explode_scrollbar_event(fields.formcontent).value)
+				end)
+			end,
+		}
+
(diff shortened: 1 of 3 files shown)


grorp:
Two maps with spawn directions specified:

[abandoned_battlefield.zip](https://github.com/MT-CTF/capturetheflag/files/8577426/abandoned_battlefield.zip)
[big_ocean.zip](https://github.com/MT-CTF/capturetheflag/files/8577427/big_ocean.zip)

LoneWolfHT:
What if you use the map center calculations by default?

savilli:
Agreed, we need reasonable defaults

grorp:
Old maps now default to the map center calculation, but for new maps you still have to specify the spawn directions manually. Is that OK?

LoneWolfHT:
> Old maps now default to the map center calculation, but for new maps you still have to specify the spawn directions manually. Are you okay with that?

It will get annoying with 3+ teams. But I guess I can fix this later when I get around to making the editor automatically set the flag positions

mods/ctf/ctf_map/mapedit_gui.lua
```diff
@@ -399,7 +419,32 @@ function ctf_map.show_map_save_form(player, scroll_pos)
...
+		elements[teamname .. "_spawn_dir"] = {
+			type = "label",
+			label = "Spawn direction: " .. name_dir(context[player].teams[teamname].spawn_dir),
+			pos = {0.2, idx},
+			-- "The first line of text is now positioned centered exactly at the position specified."
+			-- https://github.com/minetest/minetest/blob/480d5f2d51ca8f7c4400b0918bb53b776e4ff440/doc/lua_api.txt#L2929
+		}
+		idx = idx + (0.3 / 2) + (0.7 / 2)
+
+		elements[teamname.."_set_spawn_dir"] = {
```
savilli, line 433: Looks complicated but still inefficient. For some maps, diagonal direction may be the best. Let's just add the point where players will look instead of directions.

mods/ctf/ctf_modebase/features.lua
```diff
@@ -66,18 +66,25 @@ end
...
 
 	local pos = vector.offset(ctf_map.current_map.teams[tname].flag_pos,
 		math.random(-1, 1),
 		0.5,
 		math.random(-1, 1)
 	)
-	player:set_pos(pos)
+	local rotation = vector.dir_to_rotation(ctf_map.current_map.teams[tname].spawn_dir)
+
+	local function apply()
```
savilli, line 78: Random radius can be quite big. You should calculate the direction using the player's actual position.

savilli:
Why do map makes have to specify the spawn directions? You always can calculate the default if it's not specified.

grorp:
@savilli, I think I've applied your suggestions:

- You don't specify a look direction, but a position to look at.
- New maps also default to the map center calculation.
- The look direction is calculated using the player's actual position and it isn't snapped to an axis anymore.

screenshot

savilli:
Looks good to me, but I think it's better to calculate the flag center at map loading. I suggest to put there https://github.com/MT-CTF/capturetheflag/blob/092a8025a2a37b71817c6cd733fe9b98982612bc/mods/ctf/ctf_map/map_meta.lua#L162 smth like
```lua
map['flag_center'] = calc_flag_center(map)
```

grorp:
I did it this way because the code is much easier to understand if everything relating to one feature is in the same place and the difference in performance is probably negligible. Also, the flag center isn't needed when the map is loaded for editing.

grorp:
Since that doesn't really matter, I'll just do it the way you want :-)

savilli:
Works, looks good

LoneWolfHT (approved):

PR #1005, opened 2022-05-21
Shooters Mod

AerfudaX:


AerfudaX:
wrong

mods/ctf/ctf_modes/ctf_mode_shooters/textures/1
```diff
@@ -0,0 +1 @@
+
```
savilli, line 1: ?

magnetar47 (commented):
Pretty sure `ctf_melee_dagger.png` and `ctf_melee_mace.png` have nothing to do with this game-mode...

magnetar47:
Thank you for putting in the effort to create this new game mode. I'm not trying to put down this idea, but given the current state of guns, I don't think they're particularly good as the primary (and the only) weapon.

This gamemode can be enjoyable (or even playable) only when guns perform a little better, IMO. Guns are near unplayable even when there's a little lag, and they also beg for substantial improvements in the physics department. :)

LoneWolfHT:
It is possible to get a shooter mode added, but it will take a lot of work on the part of the author

LoneWolfHT:
I'd recommend releasing this as a mod first, and if it goes well and people want it merged into the main game I can do the porting myself

PR #1015, opened 2022-06-22
Introduce orb of safety

farooqkz:
This PR is going to introduce ~~two~~ one orb in game:

 - Orb of Safety which teleports the user to one of teammates randomly
 - ~~Orb of Protection which teleports the user to the teammate carrying flag if any~~


mods/ctf/ctf_orbs/init.lua
```diff
@@ -17,37 +17,27 @@ minetest.register_craftitem("ctf_orbs:safety_orb", {
...
-			itemstack.take_craftitem(1)
-			return itemstack
 		end
-
+		local teammates = {}
+		local i = 1
+		for player, _ in pairs(ctf_teams.online_players[pteam].players) do
+			if pname ~= player then
+				teammates[i] = player
+				i = i + 1 -- Why no += in Lua? :|
```
Zardshard, line 31: If you want to avoid the `i = i + 1`, you could use `table.insert`. But, still, why is there no `+=`?

farooqkz: thanks. done

mods/ctf/ctf_orbs/init.lua
```diff
@@ -0,0 +1,85 @@
+minetest.register_craftitem("ctf_orbs:safety_orb", {
+	description = "Teleport to one of teammates randomly",
+	short_description = "Orb of Safety",
+	inventory_image = "orb_of_safety.png",
+	stack_max = 1,
+	sound = {
+		use = "use.ogg"
+	},
```
Zardshard, line 8: According to Minetest's [documentation](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L256) The `orb_of_safety.png` should be in a folder called `textures`. Similarly, `use.ogg` should be in a folder called `sounds`. Also, according to [the section on textures](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L413), `orb_of_safety.png` should be called something more like `ctf_orbs_safety_orb.png`. Similarly, [the section on sounds](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L868) suggests naming `use.ogg` `ctf_orbs_use.ogg`

farooqkz: okay will fix that soon

Zardshard:
I should probably warn you that this may be rejected. Orbs of Safety allow you to be nearly immortal by letting you teleport out of a bad situation. Apples used to be somewhat similar, in that, assuming you could click fast enough, you could survive some very bad situations. In the case of apples, their use was [rate-limited](https://github.com/MT-CTF/capturetheflag/commit/48b300dd2b47cf8a31e1d49abeec6a2575705abc) and then [rate-limited some more](https://github.com/MT-CTF/capturetheflag/commit/8f358dba7379f05522e206c847028686b7bfb37d). Of course, YMMV.

farooqkz:
On some more thought, yeah I think it doesn't make sense for CTF. At least not in this way. I will think about one better way. But this PR meant to add two orbs, one is orb of safety which I conclude it is not a good idea and the other orb of protection which teleports "user" to the teammate carrying enemy flag. I think that makes much more sense for CTF as it encourages teamwork. What do you think @LoneWolfHT @savilli ?

LoneWolfHT:
Incentives for teamwork should fit in with the rest of the game. This feels a bit cheap, especially in the non-NadeFight modes

savilli:
I agree that it probably doesn't fit in the game

farooqkz:
Okay closing it then.

PR #1016, opened 2022-06-22
Fix treasure string conversion

Zardshard:


Zardshard:
The only time this is used (I believe) is by maps that specify the `treasures` field. No maps do this currently. For testing, you can replace a map's `treasures = ` with `treasures = default:lava_source;1;10;1;0.2;1;default:water_source;1;10;1;0.2;1;`. There should be water and lava blocks in some of the chests on that map.

LoneWolfHT (approved):

mods/ctf/ctf_modebase/treasure.lua
```diff
@@ -21,12 +21,12 @@ function ctf_map.treasure.treasure_from_string(str)
...
+	for name, min_count, max_count, max_stacks, rarity in str:gmatch("([^%;]+);(%d*);(%d*);(%d*);([%d.]*);%d;") do
 		out[name] = {
-			min_count  = min_count  or 1,
-			max_count  = max_count  or 1,
-			max_stacks = max_stacks or 1,
-			rarity     = rarity     or 0.5,
+			min_count  = min_count        or 1,
+			max_count  = max_count        or 1,
+			max_stacks = max_stacks       or 1,
+			rarity     = tonumber(rarity) or 0.5,
```
savilli, line 29: Why there is a need for `tonumber` here, but there is no need before?

magnetar47:
> The only time this is used (I believe) is by maps that specify the `treasures` field. No maps do this currently. For testing, you can replace a map's `treasures = ` with `treasures = default:lava_source;1;10;1;0.2;1;default:water_source;1;10;1;0.2;1;`. There should be water and lava blocks in some of the chests on that map.

I believe quite a few maps used to specify custom treasures earlier, and they all worked alright to the best of my knowledge. Does this fix something in specific?

mods/ctf/ctf_modebase/treasure.lua
```diff
@@ -21,12 +21,12 @@ function ctf_map.treasure.treasure_from_string(str)
 
 	local out = {}
 
-	for name, min_count, max_count, max_stacks, rarity in str:gmatch("([^%;]+);(%d*);(%d*);(%d*);(%d*);%d;") do
+	for name, min_count, max_count, max_stacks, rarity in str:gmatch("([^%;]+);(%d*);(%d*);(%d*);([%d.]*);%d;") do
```
magnetar47, line 24: This stands out to me as the main change this PR proposes. Please help me understand what this does.

magnetar47: Oh, I see. I'm surprised it didn't match floats before, given that the range is literally between 0 and 1 like you said. 👀

mods/ctf/ctf_modebase/treasure.lua
savilli: Only for that? Doesn't the code later expect min_count, max_count, max_stacks, and rarity to be numbers?

mods/ctf/ctf_modebase/treasure.lua
```diff
@@ -21,12 +21,12 @@ function ctf_map.treasure.treasure_from_string(str)
...
+	for name, min_count, max_count, max_stacks, rarity in str:gmatch("([^%;]+);(%d*);(%d*);(%d*);(%d*%.%d*);%d;") do
 		out[name] = {
-			min_count  = min_count        or 1,
-			max_count  = max_count        or 1,
-			max_stacks = max_stacks       or 1,
-			rarity     = tonumber(rarity) or 0.5,
+			min_count  = tonumber(min_count)  or 1,
+			max_count  = tonumber(max_count)  or 1,
+			max_stacks = tonumber(max_stacks) or 1,
+			rarity     = tonumber(rarity)     or 0.5,
```
magnetar47, line 29: Was this section of code not working without `tonumber` till now? I thought these vars were implicitly converted to numbers either in `string.match` (L24) or in the comparison lines (L26-L29)?

savilli:
Now it doesn't parse strings like `default:lava_source;1;10;1;1` with rarity = 1. Don't overengineer it, `([%d.]*)` was good.

mods/ctf/ctf_modebase/treasure.lua
savilli: No, these vars weren't numbers before now. It only worked because lua library is quite loyal to pass strings instead of numbers.

magnetar47: > No, these vars weren't numbers before now. It only worked because lua library is quite loyal to pass strings instead of numbers.

If they weren't converted to numbers, did the vars get set to 1 (0.5 for `rarity`) automatically?

savilli: ofc, they were strings or nil

magnetar47: Oh that sucks.

PR #1018, opened 2022-07-05
Improve the map editor

Zardshard:
- Make the process more self-explanatory
  - Improve error messages
  - Give hints on how to proceed
- Improve GUI
  - Make reopening of GUI after selecting coordinates more consistent.
  - Visually display coordinate selections with `worldedit`
  - Automatically place flag at chosen flag location
  - Placing flag automatically sets flag location
  - Add button to cancel map editing
  - Rearrange map editor formspec
- Write documentation

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -12,7 +12,7 @@ globals = {
 	"dropondie", "grenades",
 
 	"chatcmdbuilder", "crafting", "hpbar", "playertag", "random_messages",
-	"skybox", "throwable_snow",
+	"skybox", "throwable_snow", "worldedit",
 
 	"default", "doors", "player_api", "sfinv", "binoculars",
 
--- a/mods/ctf/ctf_map/README.md
+++ b/mods/ctf/ctf_map/README.md
@@ -2,85 +2,122 @@
 
 ## Creating a new map
 
-### Youtube tutorial
-
-
 ### 1. Dependencies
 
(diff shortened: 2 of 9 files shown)


LoneWolfHT (approved):
Code looks good, I can test before merge

xenonca:
Tested and works, this seems to bring in some really nice improvements!
A few comments:
- Eventually rename the button "Start Editing" to something like "Edit Selected Map", I think this is more clear about what it does. Could also highlight the "Create New Map" by tinting that button just a little bit (green? blue?).
- Be consistent with message coloring. It welcomes you (and finishes map editing) with orange messages but continues with some white ones like "Done getting positions!".
- If flags get placed automatically when you punch a node to select the flag's position I suppose they should be removed from the creative inventory, don't think we need two ways to place flags

I love the documentation on how to use the map editor, was working on a doc myself but this already seems more than sufficient.

Zardshard:
Great ideas! I implemented two of them.
Some comments on the other two:
- Renaming "Start Editing" to "Edit Selected Map" makes its use clearer, but I'm afraid it would require renaming "Resume Editing" to something like "Resume Editing Selected Map". Otherwise, "Resume Editing" would appear to resume editing the last selected map. But maybe it already implies that without any name changes... What a tough problem!
-  I like the idea of tinting the "Create New Map" button a certain color. However, implementing it is a little more complicated than I want to spend time on.

PR #1045, opened 2022-08-15
ctf_chat: Prevent /msg from bypassing filter

magnetar47:
Solution seems hacky to me. Surely `filter` must do this automatically? Maybe the `override_chatcommand` messes it up? I have no clue how this worked in v2 or earlier.

Tested; works.

### DIFF
--- a/mods/ctf/ctf_chat/init.lua
+++ b/mods/ctf/ctf_chat/init.lua
@@ -10,6 +10,12 @@ minetest.override_chatcommand("msg", {
 			return false, "The player " .. sendto .. " is not online."
 		end
 
+		-- Run the message through filter if it exists
+		if filter and not filter.check_message(name, message) then
+			filter.on_violation(name, message)
+			return false
+		end
+
 		-- Message color
 		local color = minetest.settings:get("ctf_chat.message_color") or "#E043FF"
 		local pteam = ctf_teams.get(name)
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -12,7 +12,7 @@ globals = {
 	"dropondie", "grenades",
 
 	"chatcmdbuilder", "crafting", "hpbar", "playertag", "random_messages",
-	"skybox", "throwable_snow", "worldedit",
+	"skybox", "throwable_snow", "worldedit", "filter",
 
 	"default", "doors", "player_api", "sfinv", "binoculars",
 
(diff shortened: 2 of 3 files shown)


magnetar47:
I've tested this PR after disabling `filter` mod as well - the game runs just fine, even when I use `/msg`. Any ideas as to how I can satisfy the CI?

Thomas--S:
You probably have to add filter here: https://github.com/MT-CTF/capturetheflag/blob/master/.luacheckrc#L3

magnetar47:
> You probably have to add filter here: https://github.com/MT-CTF/capturetheflag/blob/master/.luacheckrc#L3

I thought as much, but there are other mods using it without causing luacheck to fail. I'll add it anyway, if that's the only way to appease the CI.

Thomas--S (approved):
Didn't test it in game, but code looks good to me.

mods/ctf/ctf_chat/init.lua
```diff
@@ -10,6 +10,12 @@ minetest.override_chatcommand("msg", {
 			return false, "The player " .. sendto .. " is not online."
 		end
 
+		-- Run the message through filter if it exists
+		if filter and not filter.check_message(name, message) then
```
Panquesito7, line 14: You might want to check if the mod is loaded by using `minetest.get_modpath()`. What do you think? 🤔


__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :question: I have a question

magnetar47: Doesn't `if filter` do the same thing? Would it be possible that this evaluates to `true` when the mod may not be loaded? Also, the optional depends makes sure that if filter exists, it's loaded before this mod.

magnetar47:
For some context:

https://github.com/minetest-mods/filter/blob/b3d58cd83d4e52eeedf1bb730fdfb569a89b1257/init.lua#L180-L184

`filter` automatically injects this check in all registered chat-commands, and also overrides the `minetest.register_chatcommand` function to automatically inject this into chatcommands that'll be registered afterwards.

What I believe is happening here, is that the original `func` of `/msg` is injected with the filter check, and then this chat-command is over-ridden here:
https://github.com/MT-CTF/capturetheflag/blob/54cb7a6f67d64f07e68389a22159cf8f1c8a6045/mods/ctf/ctf_chat/init.lua#L3-L27

I guess one could also fix this by having `filter` overriding `minetest.override_chatcommand` to inject the checker? It may not sit well with mods, that explicitly want to override the chat-command to bypass this checker for one particular chat-command, for example.

Related: minetest-mods/filter#11

LoneWolfHT (approved):

PR #1046, opened 2022-08-17
[NO SQUASH] Prevent initial stuff being dropped upon player death

### LINKED ISSUE #1041: Player inventory loadouts
It'd be extremely time-saving to have "loadouts", where the initial stuff is automagickally rearranged in the player's hotbar according to each player's preferences. I (and I'm sure many others) waste many precious seconds re-arranging their hotbar after respawn.

I can think of two approaches:

#### Implicit - Live recording

Simply remember each player's hotbar setup in that map (and in that class, if applicable), and auto-arrange initial items accordingly. Maybe we could make it so that the hotbar arrangement is recorded only upto 5s after (re)spawn, as it may not be worth it to continuously watch each player's hotbar throughout the match. Easier to implement.

#### Explicit - Separate inv page

Separate inv page that when opened shows a full inv worth of slots, and a separate "hotbar" row. initial items of the current map (and class, if applicable) is added to the inv, from which player can take items and arrange in the "hotbar" row. Then they can click on "Save" and that arrangement will be followed henceforth. Little more effort to implement, but *could* be worth it IMO.

I'll maybe try to make a prototype formspec to demonstrate what I mean.

Thoughts, suggestions, and discussion welcome.

magnetar47:
Closes #1041.

I've also added a method `ctf_modebase.player.get_initial_stuff(player)`. I can think of more use-cases where this information is necessary, apart from this PR.

### How it werks

In `dropondie`, the invlist to be dropped as items is first removed of all `initial_stuff` for that player. The `initial_stuff` is moved to a new invlist of equal size, and the indices of these initial items are preserved from the original invlist. What remains in the original invlist is then dropped.

The second part of the PR (which is currently WIP) is to not keep giving the initial stuff on every respawn, as they're preserved across deaths now.

### Proposed changes
- [x] Add method `ctf_modebase.player.get_initial_stuff(player)` - includes initial items from both map and class.
  - returns `{ itemstack, itemstack2, itemstack3 }`
- [x] Don't drop initial stuff upon death
- [ ] Prevent initial stuff from being given every respawn
- [ ] Testing. More testing. MOAR testing

### Testing procedure

1. Launch CTF. `singleplayer` *should* be fine.
2. Run the following commands:
  2.1. `/grantme ctf_admin`
  2.2. `/ctf_next -f mode:classes caverns`
3. Arrange your inventory in your own preferred way **OR** in any random bizzare way.
4. Die. Just die; how you die doesn't matter - no one cares about you. :P
5. Once you respawn, make sure you have the default items in the same slots as you had left them. Make sure you have exactly 20 torches and 5 sticks.
6. Try these different scenarios after each respawn:
  6.1. Throw out 10 torches, so that you're left with only 10 torches.
  6.2. Collect more torches so that you have more than 20 cobble.
  6.3. Don't add or remove any torches in your inventory, but split the existing torches into multiple stacks.
7. Go to step 3. Keep dying until you're either satisfied with the PR or fed up with life.

Werk-in-progress. Don't start testing right now - I'll let you know when the PR is ready.

### DIFF
--- a/mods/ctf/ctf_modebase/player.lua
+++ b/mods/ctf/ctf_modebase/player.lua
@@ -15,6 +15,14 @@ local function get_initial_stuff(player, f)
 	end
 end
 
+function ctf_modebase.player.get_initial_stuff(player)
+	local initial_stuff = {}
+	get_initial_stuff(player, function(item)
+		table.insert(initial_stuff, item)
+	end)
+	return initial_stuff
+end
+
 function ctf_modebase.player.give_initial_stuff(player)
 	minetest.log("action", "Giving initial stuff to player " .. player:get_player_name())
 
--- a/mods/pvp/dropondie/init.lua
+++ b/mods/pvp/dropondie/init.lua
@@ -1,15 +1,47 @@
 dropondie = {}
 
-local function drop_list(pos, inv, list)
-	for _, item in ipairs(inv:get_list(list)) do
-		local obj = minetest.add_item(pos, item)
+local function drop_list(pos, player, listname)
+	local initial_stuff = ctf_modebase.player.get_initial_stuff(player)
+	local inv = player:get_inventory()
+	local invlist = inv:get_list(listname)
+	local newlist = {}
 
+	-- Move initial stuff from old list to new list
+	for list_i, item in ipairs(invlist) do
+		for stuff_i = 1, #initial_stuff do
+			-- Initialize this within loop so that the value can be changed within the loop
+			local i_item = initial_stuff[stuff_i]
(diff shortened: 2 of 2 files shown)


farooqkz:
The WIP label is not required as GH now supports WIP/Draft PRs.

tigercoding56:
i think that if you used any of the initial items you should still respawn with replacements , for example if you lost your sword (by dropping it into lava or it ran out of durability)

LoneWolfHT:
Closing for now, as you seem to be busy, feel free to reopen if you become interested in this again though

PR #1048, opened 2022-08-25
ctf_chat: Replace the sender-side response for /msg

Vagankovo:
Currently when you send a private message, you get a nondescript `Message sent.`, which makes it harder to talk. The changes will make /msg give out `To <RECEPIENT>: <MESSAGE>` instead.

### DIFF
--- a/mods/ctf/ctf_chat/init.lua
+++ b/mods/ctf/ctf_chat/init.lua
@@ -31,7 +31,7 @@ minetest.override_chatcommand("msg", {
 		str = "Message sent to " .. sendto .. ": " .. message
 
 		minetest.log("action", string.format("[CHAT] PM from %s to %s: %s", name, sendto, message))
-		
+
 		-- Send the sender-side message
 		return true, str
 	end
--- a/mods/ctf/ctf_chat/init.lua
+++ b/mods/ctf/ctf_chat/init.lua
@@ -27,10 +27,8 @@ minetest.override_chatcommand("msg", {
 		str = str .. minetest.colorize(color, ": " .. message)
 		minetest.chat_send_player(sendto, str)
 
-		-- Colorize the sender-side message
-		str = minetest.colorize (color, "To ")
-		str = str .. minetest.colorize(tcolor, name)
-		str = str .. minetest.colorize(color, ": " .. message)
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_chat/init.lua
```diff
@@ -21,14 +21,19 @@ minetest.override_chatcommand("msg", {
...
 
-		-- Colorized sender name and message
+		-- Colorize the recepient-side message and send it to the recepient
 		local str =  minetest.colorize(color, "PM from ")
 		str = str .. minetest.colorize(tcolor, name)
 		str = str .. minetest.colorize(color, ": " .. message)
 		minetest.chat_send_player(sendto, str)
 
+		-- Make the sender-side message
+		str = "To " .. name .. ": " .. message
```
magnetar47, line 31: `name` is the player name of the sender, not the receiver. See L#24 above. The PM goes to the recipient like so: `PM from <name>: ...`. So if you use `name` here, then the sender will see their own name. Recipient's name is stored in the `sendto` var (see [here](https://github.com/MT-CTF/capturetheflag/blob/88fc9e15549af1ba95d764079f1617a553baa8a6/mods/ctf/ctf_chat/init.lua#L9)).

I'd also like to see the recipient name `minetest.colorize`-d for consistency.

magnetar47 (changes requested):
Not sure if this change is worth it, but since you've made a PR yourself, then why not.

mods/ctf/ctf_chat/init.lua
Vagankovo: Thanks for the catch, I only tested in singleplayer.

As of the colorization, Vagnar said in Discord that it should stay white *as is*: https://discord.com/channels/447819017391046687/1012316794062639184/1012350458884534382

Vagankovo: Isn't this supposed to be resolved after our conversation, or did I miss something?

LoneWolfHT (approved):

mods/ctf/ctf_chat/init.lua
magnetar47: Sorry, I went AFK halfway through this conversation. :P

PR #1053, opened 2022-09-16
Altering /ctf_next(making the map param optional)

farooqkz:
This PR makes the map parameter of `/ctf_next` command optional. If one is not provided, one will be chosen randomly.

### DIFF
--- a/mods/ctf/ctf_modebase/match.lua
+++ b/mods/ctf/ctf_modebase/match.lua
@@ -79,9 +79,9 @@ function ctf_modebase.start_new_match(delay)
 end
 
 minetest.register_chatcommand("ctf_next", {
-	description = "Set a new map and mode",
+	description = "Set a new map and mode. If a new map is not specified one will be chosen randomly.",
 	privs = {ctf_admin = true},
-	params = "[-f] <mode:technical modename> <technical mapname>",
+	params = "[-f] <mode:technical modename> [technical mapname]",
 	func = function(name, param)
 		minetest.log("action", string.format("[ctf_admin] %s ran /ctf_next %s", name, param))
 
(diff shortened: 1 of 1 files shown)


magnetar47:
Would it be a good idea to have the command follow the existing map order, instead of implement a randomiser from scratch? Also, I think that the current implementation will select a map that isn't designed for the specified class.

farooqkz:
> Would it be a good idea to have the command follow the existing map order, instead of implement a randomiser from scratch? 

I didn't know a map order already exists.

> Also, I think that the current implementation will select a map that isn't designed for the specified class.

You might be right. Any idea on how to prevent this?

LoneWolfHT:
Are you aware of /ctf_skip ?

farooqkz:
> Are you aware of /ctf_skip ?

I wasn't. But now checking, it doesn't seem to change the game mode.

LoneWolfHT:
> > Are you aware of /ctf_skip ?
> 
> I wasn't. But now checking, it doesn't seem to change the game mode.

Is that a bug report?

farooqkz:
> 

Sorry. I don't understand.

LoneWolfHT:
> This PR makes the map parameter of /ctf_next command optional.
It *is* optional: `/ctf_next mode:modename`

> I wasn't. But now checking, it doesn't seem to change the game mode.
It does. You don't set the next gamemode with it though

farooqkz:
Yeah you seem to be right. Closing this PR

magnetar47:
I wonder what the point of `/ctf_next` is, if neither map nor mode is specified, especially because `/ctf_next -f` is essentially `/ctf_skip` (from what I unerstand). 🤔

LoneWolfHT:
> I wonder what the point of `/ctf_next` is, if neither map nor mode is specified, especially because `/ctf_next -f` is essentially `/ctf_skip` (from what I unerstand). 🤔

Typing practice I guess

Issue #1032, opened 2022-08-04
Formspec text clipping


magnetar47:
![screenshot_20220801_222338](https://user-images.githubusercontent.com/88258122/182772233-e4c00931-7600-496b-91c2-881f7cd9378d.png)

Crucial bit of info which is so hard to understand because some important parts have been clipped by the formspec. Would love to have word wrapping or something similar. Plis fix.

FYI, this screenshot was taken with a MT res of 1366x768.

Btw I did a quick search to see if there was already an issue on this and I couldn't find any. Please feel free to close this if there is indeed an old one.

farooqkz:
This is now solved.

PR #1066, opened 2022-12-11
Adding 2*7 slots to teamchest

farooqkz:
screenshot_20221211_152953

I thought 14 more slots are enough and there is no need to double it.


LuismiInTheZone:
Everything seems fine, just needs to get an approval from some mod

LoneWolfHT:
You should make the formspec wider to account for the addition. Not sure how that will work on smaller screens though

LoneWolfHT:
Note #1032

farooqkz:
Solved 
screenshot_20221212_161623

PR #1072, opened 2022-12-21
Fix Minetest path text

Panquesito7:
Things added/changed:

- Fix the Minetest path text in the CTF map guide.
  - The text to the `ctf_map` directory seemed to be wrong. I hope it looks better this way. 🙂


LoneWolfHT:
This looks more confusing to me

LoneWolfHT:
This should be readable as a text file too

Panquesito7:
How should I proceed with this, then? 🤔

LoneWolfHT:
If you can't come up with anything I'd say it's good as it is. Thanks for the PR though

PR #1074, opened 2022-12-24
Fix for /lb showing bounty score

farooqkz:
`/lb` sometimes shows values like `18.6666...`. This fixes that using `math.floor`


LoneWolfHT:
Might want to fix this in the bounty creation code instead

PR #1082, opened 2023-01-24
Make opened treasure chests appear open

MrRar:
Right now opened chests have cracks. This is weird. It would be better for them to look open instead. This PR makes opened chests use the open chest mesh from MTG default which is already in the CTF source tree.


LoneWolfHT:
Thanks for the PR! Sorry for the wait

MrRar:
> Thanks for the PR! Sorry for the wait

No problem. You're welcome!

PR #1090, opened 2023-02-02
add aliases for /abstain /yes and /no

farooqkz:
Solves #1089


farooqkz:
@LoneWolfHT This is so trivial and I believe should not wait long.

LoneWolfHT (approved):

Issue #1095, opened 2023-02-12
Damage cobble doesn't deal damage unless match has started


LoneWolfHT:
Enemies can dig your damage cobble before the match has started, without taking damage from it

The fix should probably just prevent digging if that match hasn't started

PR #1097, opened 2023-02-13
Fix the bug of damage cobble when match hasn't yet started

farooqkz:
With this PR, #1095 will be closed and damage cobbles cannot be removed before match start from an enemy.


farooqkz:
@LoneWolfHT ping

PR #1113, opened 2023-03-12
make bandages change horizontal look of enemies randomly

farooqkz:
This PR adds a new feature to medic bandages(not normal bandages):

 - They have small chance changing the enemy's horizontal look to a random value causing distraction for them.

### DIFF
--- a/mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
+++ b/mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
@@ -291,6 +291,21 @@ ctf_healing.register_bandage("ctf_mode_classes:support_bandage", {
 	heal_min = 4,
 	heal_max = 5,
 	rightclick_func = function(itemstack, user, pointed)
+		math.random(1, 100)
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
I think this is a little too clunky, and it takes medic's focus off of healing

farooqkz:
> I think this is a little too clunky, and it takes medic's focus off of healing

Well with this low probability, it isn't rational that a medic hits an enemy constantly till this happens to them. However in a fight when a medic is supporting a knight, this could have extra good effect for the fight in the favor of medic's team. And that's my goal.

LoneWolfHT:
It'll be too confusing for players, even if you explain it somewhere.
Subtle effects like this should be added sparingly to keep things simple

farooqkz:
So you think medic's bandage must not have any effect on enemies?

LoneWolfHT:
> So you think medic's bandage must not have any effect on enemies?

If it does have an effect it should be one that's big enough that it could/would replace their invincibility effect

PR #1115, opened 2023-03-17
Introduce flagwars game mode

farooqkz:
Tasks

 - [x] flagwar mode itself
 - [ ] Treasure spawner(gold, mese, diamond, iron)
 - [ ] Shop

Subtasks for shop:

 - [ ] texture
 - [ ] Shopping inventory

Do you think a vending machine would be fun, too?

### DIFF
--- a/mods/ctf/ctf_modes/ctf_mode_flagwars/init.lua
+++ b/mods/ctf/ctf_modes/ctf_mode_flagwars/init.lua
@@ -1,0 +1,115 @@
+local rankings = ctf_rankings.init()
+local recent_rankings = ctf_modebase.recent_rankings(rankings)
+local features = ctf_modebase(rankings, recent_rankings)
+
+local old_bounty_reward_func = ctf_modebase.bounties.bounty_reward_func
+local old_get_next_bounty = ctf_modebase.bounties.get_next_bounty
+local old_get_skin = ctf_cosmetics.get_skin
+local custom_item_levels = table.copy(features.initial_stuff_item_levels)
+
+ctf_modebase.register_mode("flagwars", {
+	treasures = {
+		["default:ladder_wood"] = {
+			max_count = 20,
+			rarity = 0.3,
+			max_stacks = 5
+		},
+		["default:torch"] = {
+			max_count = 10,
+			rarity = 0.3,
+			max_stacks = 5,
+		},
+		["default:cobble"] = {
+			max_count = 20,
+			min_count = 2,
+			max_stacks = 2,
+			rarity = 0.15,
+		},
+		["default:wood"] = {
+			max_count = 20,
+			min_count = 2,
+			max_stacks = 2,
+			rarity = 0.15,
+		},
+	},
+	crafts = {},
+	physics = {
+		sneak_glitch = true,
+		new_move = false,
+	},
+	team_chest_items = {
+		"default:cobble 40",
+		"default:wood 40",
+		"ctf_map:damage_cobble 20",
+		"ctf_map:rinforcedc_cobble 20",
+		"default:torch 10",
+		"ctf_teams:door_steel 1",
+	},
+	rankings = rankings,
+	recent_rankings = recent_rankings,
+	summary_ranks = {
+		_sort = "score",
+		"score",
+		"flag_captures",
+		"flag_attempts",
+		"kills",
+		"kill_assits",
+		"bounty_kills",
+		"deaths",
+		"hp_healed",
+	},
+	build_timer = 90,
+	is_bound_item = function(_ name)
+		return false
+	end,
+	stuff_provider = function(player)
+		local initial_stuff = {
+			"default:sword_stone",
+			"default:torch 10",
+			"default:cobble 10",
+		}
+
+		return initial_stuff
+	end,
+	initial_stuff_item_levels = custom_item_levels,
+	is_restricted_item = function(player, name)
+		return false
+	end,
+	on_mode_start = function()
+		ctf_+modebase.bounties.bounty_reward_func = ctf_modebase.bounty_algo.kd.bounty_reward_func
+		ctf_modebase.bounties.get_next_bounty = ctf_modebase.bounty_algo.kd.get_next_bounty
+		ctf_cosmetics.get_skin = function(player)
+			if not ctf_teams.get(player) then
+				return old_get_skin(player)
+			end
+
+			return old_get_skin(player)
+		end
+	end,
+	on_mode_end = function()
+		ctf_modebase.bounties.bounty.bounty_reward_func = old_bounty_reward_func
+		ctf_modebase.bounties.get_next_bounty = old_get_next_bounty
+		ctf_cosmetics.get_skin = old_get_skin
+	end,
+	on_new_match = features.on_new_match,
+	on_match_end = features.on_match_end,
+	team_allocator = features.team_allocator,
+	on_allocplayer = features.on_allocplayer,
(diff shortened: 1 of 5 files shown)


farooqkz:
1. What must we use for the spawner thing? I believe an ABM  fits it so that it'll spawn only when somebody is around.
2. I want to save the item each spawner spawns in its metadata, is it a good idea? Or perhaps separate gold, iron, and diamond spawners is a better idea?
3. Alternatively, Spawners could spawn something on punch/click. Is this a better idea?
4. Should each team's shop create a blast on destroy? Damaging nearby players and nodes/blocks?

LoneWolfHT:
> 1. What must we use for the spawner thing? I believe an ABM  fits it so that it'll spawn only when somebody is around.
> 2. I want to save the item each spawner spawns in its metadata, is it a good idea? Or perhaps separate gold, iron, and diamond spawners is a better idea?
> 3. Alternatively, Spawners could spawn something on punch/click. Is this a better idea?
> 4. Should each team's shop create a blast on destroy? Damaging nearby players and nodes/blocks?

1: Anything but an ABM please, it fits but is horribly inefficient. I'd suggest using a globalstep that loops through (a) table(s) of dropper nodes every x seconds
2. Metadata is not saved in maps, so a separate node for each spawner would be best
3. No
4. Shops should be unbreakable

farooqkz:
Is it a good idea shops having some HP and when it reaches zero, the shop gets destroyed?

LoneWolfHT:
> Is it a good idea shops having some HP and when it reaches zero, the shop gets destroyed?

I don't think so. It's not done like that in Bedwars either AFAIK.
I'd recommend getting the original idea done first, and innovating once the mode is on the server being played

LoneWolfHT:
`mtg_default` should be `default`, I changed the folder name to make it more obvious where the mod was from, the mod.conf name is still `default`

farooqkz:
@LoneWolfHT The shop and the spawners must be placed in maps, somehow. My suggestion is a new property for maps which is a list of coordinates and nodes to place in a given game mode. What do you think?

LoneWolfHT:
> @LoneWolfHT The shop and the spawners must be placed in maps, somehow. My suggestion is a new property for maps which is a list of coordinates and nodes to place in a given game mode. What do you think?

How about a node group that causes on_construct to be called for it at around the time chests are placed? No need to record/save its position then

farooqkz:
A "node group"?

LoneWolfHT:
> A "node group"?

In the node def. It would look something like:
```lua
groups = {init_on_map_load = 1},
```

farooqkz:
> > A "node group"?
> 
> In the node def. It would look something like:
> 
> ```lua
> groups = {init_on_map_load = 1},
> ```

Excuse me but I still don't understand how such a group can make nodes with this group appear in certain places in a map.

LoneWolfHT:
> Excuse me but I still don't understand how such a group can make nodes with this group appear in certain places in a map.

I would need to add code to one of the map placement LVMs that checks for it and then calls that node's on_construct

farooqkz:
> > Excuse me but I still don't understand how such a group can make nodes with this group appear in certain places in a map.
> 
> I would need to add code to one of the map placement LVMs that checks for it and then calls that node's on_construct

But I still don't understand where do you get the placement positions from.

LoneWolfHT:
> But I still don't understand where do you get the placement positions from.

The LVMs loop through all nodes on the map in order to find places to place the treasure chests

farooqkz:
> > But I still don't understand where do you get the placement positions from.
> 
> The LVMs loop through all nodes on the map in order to find places to place the treasure chests

oh so you meant putting them randomly on the map?

LuismiInTheZone:
One tiny question about this new project for CTF v4, after it's finished, we'll need to create new maps no? So... maybe creating a tiny tutorial of how to create new maps would be ok too... Also, about the poll you made Landar... is for deleting one mode... right :(

LuismiInTheZone:
> One tiny question about this new project for CTF v4, after it's finished, we'll need to create new maps no? So... maybe creating a tiny tutorial of how to create new maps would be ok too... Also, about the poll you made Landar... is for deleting one mode... right...

I could make the tutorial by the way if it's needed

LoneWolfHT:
> One tiny question about this new project for CTF v4, after it's finished, we'll need to create new maps no? So... maybe creating a tiny tutorial of how to create new maps would be ok too... Also, about the poll you made Landar... is for deleting one mode... right...

I'm not planning on ever making a CTFv4, as that implies a near-full rewrite with my current versioning system.

This mode probably will need maps specially made for it, so I will hopefully be able to release a mapmaking tutorial with it.

As to the poll, I don't plan on deleting anything based on its results, it's just to get an idea of what I could improve

LoneWolfHT:
> oh so you meant putting them randomly on the map?

No, put them on the map and add `load_on_map_init=1` to their node groups. I'll handle the rest :P

FlamingRCCars:
> 1. What must we use for the spawner thing? I believe an ABM  fits it so that it'll spawn only when somebody is around.

Here is an some code from my item spawner, it uses node timers which only have to be started once (using an ABM or something), then when the match ends I stop them and clear objects.

`minetest.register_node("bedwars:base_spawner", {
    description = S("Base Item Spawner"),  -- Description for translation
    tiles = {"default_stone.png"}, -- Change the texture to your preference
    groups = {cracky = 3},
    drop = "", -- No drop when the node is dug,
    paramtype2 = "facedir",
    legacy_facedir_simple = true,
    is_ground_content = false,

    on_construct = function(pos)
		local timer = minetest.get_node_timer(pos)
        timer:start(2)
    end,
	on_timer = function(pos, elapsed)
        local meta = minetest.get_meta(pos)
		local time = meta:get_int("time")
		if time == nil then
			time = 0
		end
		pos.y = pos.y + 1.5
		time = time + math.round(elapsed)
		if time > 14 then
			local item = ItemStack("default:gold_ingot 1")
        	local objects = minetest.get_objects_inside_radius(pos, 2)
			local count = 0
        	for _, object in ipairs(objects) do
                local entity = object:get_luaentity()
                if entity and entity.name == "__builtin:item" then
                	local stack = ItemStack(entity.itemstring)
                	if stack:get_name() == item:get_name() then
						count = count + stack:get_count()
					end
            	end
    		end
			if count < 12 then
				minetest.add_item(pos, item)
			end
			time = 0
		end
		if time == 3 or time == 7 or time == 10 or time == 14 then
			local item = ItemStack("default:steel_ingot 1")
			local objects = minetest.get_objects_inside_radius(pos, 2)
			local count = 0
        	for _, object in ipairs(objects) do
                local entity = object:get_luaentity()
                if entity and entity.name == "__builtin:item" then
                	local stack = ItemStack(entity.itemstring)
                	if stack:get_name() == item:get_name() then
						count = count + stack:get_count()
					end
            	end
    		end
			if count < 64 then
				minetest.add_item(pos, item)
			end
		end
		meta:set_int("time", time)
        return true
    end,
})`

LoneWolfHT:
I think a globalstep looping over positions would still be faster. If you want to use node timers I'd suggest comparing its speed against that of a globalstep

FlamingRCCars:
> 

I'm sure that global steps would probably be faster in this specific use case, where you only have a few spawners, but node timers start becoming faster once you have 10+, since they are handled in C, and they can be configured to only run when you want. I'll test this theory later, but there isn't really a good way to measure how long it takes multiple node timers to execute.

PR #1117, opened 2023-03-18
Add craftable fence

Lucyucy:
This change adds craftable wooden fences. Craft fences from any wood planks + sticks.

### DIFF
--- a/mods/ctf/ctf_modebase/crafting.lua
+++ b/mods/ctf/ctf_modebase/crafting.lua
@@ -120,6 +120,13 @@ crafting.register_recipe({
 	always_known = true,
 })
 
+-- Fence x4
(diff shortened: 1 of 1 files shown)


LuismiInTheZone:
O well, that was quick

farooqkz:
What are fences good for in CTF?

Lucyucy:
> What are fences good for in CTF?

I created this PR in response to bBADs suggestion. Fences are useful to build strong bases whithout blocking the view. They are easier to craft than doors and represent a practical extension.

farooqkz:
While fences are better than glasses for this purpose. They are still not good enough. I will open another issue for that.

farooqkz:
@LoneWolfHT LGTM and trivial

Issue #1094, opened 2023-02-08
Remove christmas grenades


LuismiInTheZone:
Christmas ended like one month ago, and I suggested to be eliminated, and mostly everyone said yes (and is what should happen...). So... Can christmas grenades be deleted?

LoneWolfHT:
I was going to make it a seasonal mode but I decided to focus on ranked resets instead, so I'll probably leave it until I get a better idea of my schedule

LoneWolfHT:
I'll look at reverting this at some point, not sure I'll find the time to follow through with my bigger plan anytime soon. I'll probably take the change to sync CTF's grenade mod with the one I published on CDB too

PR #1121, opened 2023-03-23
Restore old grenade textures

a-blob:
Restore previous grenade textures and names before the Xmas event. I kept the Xmas textures so they can be easily switched back.


LuismiInTheZone:
I don't know if you did this because my suggestion but I meant another thing... I told this to Landar long time ago and he said that there were more important things to do #1094

tigercoding56:
i agree

a-blob:
Oops. Probably shouldn't go around renaming branches.

a-blob:
I'll just leave this closed for now.

PR #1126, opened 2023-03-30
Two new random messages

farooqkz:


LuismiInTheZone:
It's the "In nade fights, team up with someone using void grenade and split the kill's score!" necessary because, the other one is useful but this one... I think is what the player wants

farooqkz:
> In the phrase "In nade fights, team up with someone using void grenade and split the kill's score!" I think isn't necessary because I think is what the player wants, the other one is useful but this one...

I don't follow

LoneWolfHT:
> I don't follow

The goal of teaming up is to be able to trap one or more enemies, making it easy to kill them with normal grenades. Your wording makes it seem like the goal is to get less score for your kills, you don't really even need to mention that part

farooqkz:
> > I don't follow
> 
> The goal of teaming up is to be able to trap one or more enemies, making it easy to kill them with normal grenades. Your wording makes it seem like the goal is to get less score for your kills, you don't really even need to mention that part

okay I will do another commit. But you split the score anyway.

farooqkz:
Done

LoneWolfHT:
Tweaked the first because visual properties don't matter to players who aren't interested in bounties, and will now show up as a nice surprise to players that try out the `/lb` command

Tweaked the second to give a reason for teaming up with a void nade user

PR #1129, opened 2023-03-31
Update build_timer.lua for classic

LuismiInTheZone:
I think this is a bit to much of time for classic, also I saw that if I only put 150 it will only change to classic, and you don't need to do much more problems

### DIFF
--- a/mods/ctf/ctf_modebase/build_timer.lua
+++ b/mods/ctf/ctf_modebase/build_timer.lua
@@ -1,6 +1,6 @@
 local hud = mhud.init()
 
-local DEFAULT_BUILD_TIME = 60 * 3
+local DEFAULT_BUILD_TIME = 150
 
 local timer = nil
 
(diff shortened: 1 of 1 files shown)


LuismiInTheZone:
> This should have been done the same way the other modes change their build time.
> 
> https://github.com/MT-CTF/capturetheflag/blob/0a47d3785ef5e62dcc69e5ad30ca5a913dcc7104/mods/ctf/ctf_modes/ctf_mode_classes/init.lua#L88
> 
> This got a pretty solid downvote on Discord so I'll close for now though

Huh, in my tests from minetest ctf only changed the classic timer...

LoneWolfHT:
> Huh, in my tests from minetest ctf only changed the classic timer...

It probably does, since it's currently the only mode that uses the default value. If we ever get new modes that go with the default this would affect/be affected by them though.
Thoughtless initial design on my part

Issue #1127, opened 2023-03-31
Poison grenades are underpowered


farooqkz:
I think they must deal more damage

LoneWolfHT:
This will probably have to be a poll, I'm seeing a lot of negative reactions to your PR in Discord

farooqkz:
> This will probably have to be a poll, I'm seeing a lot of negative reactions to your PR in Discord

Has the other poll ended yet?

farooqkz:
\o/

ixPost:
> > I think they must deal more damage
> 
> I agree.

Actually they should if they had a less harresing fog effect like using semitransparent circles of smoke.(Counts form smoke screen grenades to)

PR #1130, opened 2023-04-02
Update grenades.lua: make poison grenade damage more per sec

farooqkz:
This is to test. Perhaps a solution to #1127


LuismiInTheZone:
No, smoke grenade is very good, isn't need to be better, and everyone at discord is saying that no so...

PR #1131, opened 2023-04-03
Change the texture of the grenade

LuismiInTheZone:
This was from a suggestion of some weeks ago (https://discord.com/channels/447819017391046687/448136308108296213/1088196475969871953) and everyone wanted to put the old texture from the grenade...


LoneWolfHT:
The shooter mod's license for that texture will also need to be added. This PR isn't currently high on my list so it'll probably sit for a bit if you need my help with that.

I'm still unsure about this, I wonder if people wishing for the old texture should just add it to a texturepack

src4026:
> I'm still unsure about this, I wonder if people wishing for the old texture should just add it to a texturepack

What about a poll for it? 
Also, this PR is like #1121 but only the old frag nade texture, which I feel can be included in #1121 if the conclusion of this PR gets resolved.

LuismiInTheZone:
I think this PR should be close because many things, I think everyone knows the reasons, the first one being that there's now a new texture.

PR #1133, opened 2023-04-10
Two new random messages

farooqkz:


LoneWolfHT:
I don't want the random messages to advertise third-party tools, and maybe not even first-party ones, it feels out of place there.

The second help message doesn't feel helpful, I think it's usually bad to get close enough to a knight/medic to grenade them if you're a shooter, unless they're low enough health to die instantly, and even then the grenade isn't much different from other methods of damage, it doesn't temporarily negate healing received for example

farooqkz:
I can remove the message about ctf notify. Regarding the other message, I've found it very effective to remove the medic from the combination making the combination weak and very easily destroyable.

LoneWolfHT:
I'm not sure that's noteworthy enough to put in a random message

farooqkz:
> I'm not sure that's noteworthy enough to put in a random message

I'm just proposing the strategy *with* counter strategy. If this isn't noteworthy, so is the message about combination of medic and knight.

LoneWolfHT:
The shooter's grenade isn't a counter-strategy, dealing damage has no strategy to it.

I'm going to close this, but would accept random messages saying things like 'Target the medic first unless the knight is super low' or 'Shooters can launch a grenade from their rifle, which is especially useful against groups of players'

farooqkz:
> The shooter's grenade isn't a counter-strategy, dealing damage has no strategy to it.
> 
> I'm going to close this, but would accept random messages saying things like 'Target the medic first' or 'Shooters can launch a grenade from their rifle, which is especially useful against groups of players'

The idea I was trying to express is this:

When a mixed group of 1-2 medics and many knights come to attach your team, you can't kill the knights. Obviously, because the medics have their backs. You can't kill the medics "first" either using only knight's sword because as soon as you get there with your teammates, they will kill you all. It is really hard to kill the medic first unless there is just one medic and 2-3 knights concentrate their damage on the lone medic which is rare. Here comes the shooter's grenade. While the enemy group and your group are fighting, the medic(s) can get damaged a LOT instantly and then the knights from your team will quickly kill the medic(s).

LoneWolfHT:
That's a good and valid strategy, but it's for a rather specific case and doesn't fit in a random message

PR #1137, opened 2023-04-18
Another new random message and an edit

LuismiInTheZone:
I saw that random messages said nothing about the new thing of /mp, and since the new map race day came out, now blueberries are in-game too, so... (I'm sure there are a lot more that we are missing...)


LuismiInTheZone:
> Blueberries? Since when? I'm so outdated.

The new map, race day, has blueberries (new map in-game)

MoreKat:
`"As we said before, /m is for adding a mark for your team, but you can also do /mp (player) to send a marker to a specific teammate",`
Should be `marker` and `/mp <player>`

LuismiInTheZone:
> `"As we said before, /m is for adding a mark for your team, but you can also do /mp (player) to send a marker to a specific teammate",` Should be `marker` and `/mp <player>`

Okie, I'll do it in a moment

PR #1143, opened 2023-04-28
Add K/D to ranking command result

farooqkz:
Testing server:

```
test.minetest.ir:30000
```


farooqkz:
The server is not available anymore. But @LoneWolfHT, why is this waiting so long? I think it's pretty trivial

farooqkz:
@LoneWolfHT This is trivial as well

LoneWolfHT:
Tweaked locally and merged https://github.com/MT-CTF/capturetheflag/commit/359495a25cf5e0716477e95ff5c14c50a9a25610

I want to eventually have a per-player setting allowing customization of that second for loop's contents

PR #1145, opened 2023-05-06
Add sounds when player drops flag

a-blob:
I have no idea what I'm doing.


a-blob:
Still have to change the sounds when the player drops the flag. The code is using the old trumpet sound for now as a test.

farooqkz:
The sounds look good to me but I'm not a professional anyway.

mods/ctf/ctf_modebase/features.lua
```diff
@@ -116,6 +116,27 @@ local function celebrate_team(teamname)
...
+				gain = 1.0,
+				pitch = 1.0,
+			}, true)
+		else
+			minetest.sound_play("ctf_modebase_drop_flag_positive", {
+				to_player = pname,
+				gain = 1.0,
+				pitch = 1.0,
+			}, true)
+		end
```
LoneWolfHT, line 136:
```suggestion
        if pteam then
		  if pteam == teamname then
			  minetest.sound_play("ctf_modebase_drop_flag_negative", {
				  to_player = pname,
				  gain = 1.0,
				  pitch = 1.0,
			  }, true)
		  else
			  minetest.sound_play("ctf_modebase_drop_flag_positive", {
				  to_player = pname,
				  gain = 1.0,
				  pitch = 1.0,
			  }, true)
		  end
		end
```

LoneWolfHT:
Changing it to use the default sounds for now, with gain/pitch shifting. It sounds ok, but feel free to propose alternatives

PR #1148, opened 2023-05-18
Add stone swords to blacklist

FlamingRCCars:
Blacklist stone swords from going in the team chest, since they don't serve a purpose.

https://discord.com/channels/447819017391046687/448136308108296213/1105827145902731365


farooqkz:
The change looks good and is really trivial.

PR #1149, opened 2023-05-18
Prevent placement of markers while sniping

FlamingRCCars:
When zooming and using rifles, you will not leave markers.

### DIFF
--- a/mods/ctf/ctf_modebase/markers.lua
+++ b/mods/ctf/ctf_modebase/markers.lua
@@ -1,3 +1,14 @@
+local blacklist = {
+	"ctf_ranged:smg",
+	"ctf_ranged:smg_loaded",
+	"ctf_ranged:rifle",
+	"ctf_ranged:rifle_loaded",
+	"ctf_ranged:sniper_magnum",
+	"ctf_ranged:sniper_magnum_loaded",
+	"ctf_mode_classes:ranged_rifle",
+	"ctf_mode_classes:ranged_rifle_loaded"
+}
+
 local hud = mhud.init()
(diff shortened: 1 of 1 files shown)


farooqkz:
From what I understood, this PR prevents marker placement while wielding a variety of guns which I think is not something people want.

FlamingRCCars:
13-0 voted yes on this feature: https://discord.com/channels/447819017391046687/448136308108296213/1082317387845615676

FlamingRCCars:
> From what I understood, this PR prevents marker placement while wielding a variety of guns which I think is not something people want.

13-0 voted yes on this feature: https://discord.com/channels/447819017391046687/448136308108296213/1082317387845615676

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -246,6 +257,13 @@ minetest.register_globalstep(function(dtime)
 		if controls.zoom then
 			local marker_text = false
 
+			stack = player:get_wielded_item()
+			for _, itemstring in ipairs(blacklist) do
+				if stack:get_name():match(itemstring) then
+					return
```
LoneWolfHT, line 263: This will cancel zoom checks for all other players. Add a variable to track if they're holding a gun and put the following code in an if statement checking that variable

LoneWolfHT (changes requested):

mods/ctf/ctf_modebase/markers.lua
FlamingRCCars: I didn't fully understand what you meant, but I believe the changes I pushed are correct.

Issue #1073, opened 2022-12-21
Make team doors automatically close if you Sprint+Open them


LoneWolfHT:

LoneWolfHT:
The idea with Sprint + Open is that you're sprinting, so the door can close again relatively quickly, but not in your face. Players used to regular doors would also be less likely to be caught off guard by the new behavior

farooqkz:
I agree players are used to regular doors. But I don't agree because of that we must not implement improvements. In the beginning of CTFv3 most players were used to CTFv2 and many suggested having two servers: one for v2 and the other for v3.

We can have new doors which implement the new behavior and make them replace regular doors slowly.

farooqkz:
@LoneWolfHT Any new thought regarding this?
My opinion is almost as it was before. But I should add that I believe for a majority of players, this feature won't be useful. As they are not sprinting while running out of the base.

LoneWolfHT:
Can try listing the different implementations in a poll

farooqkz:
I think this idea is not good. JMA currently has auto close option and players can enable/disable it with settings. We have a PR for this too: https://github.com/MT-CTF/capturetheflag/pull/1151

PR #1151, opened 2023-05-25
Making team doors automatically close after opening.

Jo5629:
Add function so door moves after two seconds. Probably needs some reworking but it is a working robust version.


farooqkz:
As I stated in the relevant issue, I strongly disagree with this as a "sudden big change" for something very common in the game. Read the logs in #1073

farooqkz:
@Jo5629 Happy first PR to the CTF codebase :D

Jo5629:
1. Fair enough. I see what you mean. It will throw off tons of the players that have been on the server for a bit, and it will get annoying to some of them.
2. Thanks farooqkz for the shoutout of my first PR.

farooqkz:
> 1. Fair enough. I see what you mean. It will throw off tons of the players that have been on the server for a bit, and it will get annoying to some of them.
> 
>     2. Thanks farooqkz for the shoutout of my first PR.

That's my opinion. And I am not the one making decisions.

LoneWolfHT:
Should probably do a poll for this

LoneWolfHT:
Since this is a setting (now?) it probably doesn't need a poll unless we want to make it a default

Jo5629:
> Since this is a setting (now?) it probably doesn't need a poll unless we want to make it a default

Yeah I kinda forgot about this until a few days ago (I procrastinate a lot).
People were yelling at each other to close the doors and then I remembered this existed so I rewrote it. But it should be ready (I think).

farooqkz:
@Jo5629 Are you still interested?

farooqkz:
Since it is a setting, removing needs poll label.

Jo5629:
> Are you still interested?

A little bit. I mean, if you want to continue forward with this and/or optimize it, I don't mind.

AnonomousAccount:
This is a feature on JMA and it sucks, in my opinion

AnonomousAccount:
A better option might be to add "pressure plates", but theres no reason why a player cant just close their own door

farooqkz:
> This is a feature on JMA and it sucks, in my opinion

I think it's great. Why do you think it sucks?

> > Are you still interested?
> 
> A little bit. I mean, if you want to continue forward with this and/or optimize it, I don't mind.

I'm not sure about LV's opinion on this. But I'm very positive. As I've seen it on JMA and it's very good. Plus if someone doesn't like it, they can simply disable the setting. But in the end, LV has the say.

AnonomousAccount:
> > This is a feature on JMA and it sucks, in my opinion
> 
> I think it's great. Why do you think it sucks?

Its really buggy and confusing if multiple people try to open the same door at the same time, and it makes it waay harder to open two doors placed in series, because the first one will close before you can enter the second.

It also removes the disadvantage from being lazy and not closing the door

AnonomousAccount:
Uploading help_Cd2aef5k_1.mp4…

AnonomousAccount:
https://discord.com/channels/1002108415511896095/1351677020492271676/1367695807385436222

Issue #1158, opened 2023-06-08
Add /hp command to reveal your current location and HP to teammates


farooqkz:
This feature was requested in Discord in [this message](https://discord.com/channels/447819017391046687/448136308108296213/1115865726587256973). An example usage: You are low in HP and need medical attention. You do `/hp` and a marker similar to normal teammate-only markers will be created for your location with your name, your current HP and your maximum class HP.

src4026:
I believe this can be closed. #1159

PR #1159, opened 2023-06-10
Add HP Markers 

src4026:
In reference to #1158.
## PR:
- Added `hpmarker` parameter to the `marker_func`.
- Added the commands `/mhp` for adding the HP markers. (within the class of normal markers and are not considered different from normal markers)


src4026:
## Adding to "Issues to be resolved and improvements to be made:":
- Errors
- Improved HP marker message

farooqkz:
IMO, you are greatly overdoing it. There shouldn't be an extra hpmarker thing and commands to remove it. HP Markers are normal markers except that you mark your own position and your current hp plus your max hp is in the label. So no need for tracking HP markers separately and another command to remove them.

Also hp marker for another player was not requested and I think it is not needed. I am against adding that. We can however, include hp and max hp of the normal marker target if it's a player.

src4026:
> IMO, you are greatly overdoing it. There shouldn't be an extra hpmarker thing and commands to remove it. HP Markers are normal markers except that you mark your own position and your current hp plus your max hp is in the label. So no need for tracking HP markers separately and another command to remove them.
> 
> Also hp marker for another player was not requested and I think it is not needed. I am against adding that. We can however, include hp and max hp of the normal marker target if it's a player.

Yes, you are right. The process of cleaning up the code that I'll be doing in the near future wasn't only for that file itself, but for the HP marker mechanism as a whole. I have made these files only for temporary reference.

src4026:
@farooqkz, If I am correctly understanding your previous comment, you mean you want to make merge them into a single file with the HP markers as a part of the normal markers? I tried doing it, but I ran into some issues, which, I believe, are mostly by my incompetence of not knowing much about the code I am working on.

src4026:
> @farooqkz, If I am correctly understanding your previous comment, you mean you want to merge them into a single file with the HP markers as a part of the normal markers? I tried doing it, but I ran into some issues, which, I believe, are mostly by my incompetence of not knowing much about the code I am working on.

farooqkz:
There is already code which places markers. 

1. Find that code.
2. If it's not already the case, create a function to create markers
3. Re-implement `/m` and other commands with your new function
4. Add `/hp` which uses the marker placement function.

src4026:
> There is already code which places markers. 
> 
> 1. Find that code.
> 2. If it's not already the case, create a function to create markers
> 3. Re-implement `/m` and other commands with your new function
> 4. Add `/hp` which uses the marker placement function.

That is what I tried doing yesterday. I ran into some errors. I'll retry again, and if I get stuck, I'll post the code along with the error :+1:.

farooqkz:
Looks much better now but it's not my final review :)

farooqkz:
Okay. I recommend reusing the existing code by calling `marker_func` in `hpmarker_func`. However `marker_func` by default adds a marker according to player eye direction. I suggest adding another parameter(last parameter) to `marker_func` which is the looking direction. By default when it is called by `/m` it's nil and therefore it checks the player's looking direction. But when supplied, it uses that instead of actual looking direction of the player calling it. And you use that through `/hp`

src4026:
> Okay. I recommend reusing the existing code by calling `marker_func` in `hpmarker_func`. However `marker_func` by default adds a marker according to player eye direction. I suggest adding another parameter(last parameter) to `marker_func` which is the looking direction. By default when it is called by `/m` it's nil and therefore it checks the player's looking direction. But when supplied, it uses that instead of actual looking direction of the player calling it. And you use that through `/hp`

Hmm. I will look into it later when I can. Thanks 👍

src4026:
I reused the `marker_func` in `hpmarker_func`. I understand clearly why you want me to reuse it - DRY =P - because of my previous programming knowledge, but since I am very new to Lua, I need to get used to its syntax and proper usage. So, after I did that reusing, here's what I got. It may not be the best :)

src4026:
Note: There is an issue that if an HP marker is placed, it apparently doesn't set a marker cooldown (?). I tried seeing the code, but can't seem to find anything at first glance. I will be having a look at it for 10 minutes from now, and if I can't find it, that will be indicated my no pushes to this PR since the last one.

src4026:
I can't seem to find the exact reason why this is happening. Bye. I have got to sleep, see you all tomorrow :zzz:

farooqkz:
> Note: There is an issue that if an HP marker is placed, it apparently doesn't set a marker cooldown (?). I tried seeing the code, but can't seem to find anything at first glance. I will be having a look at it for 10 minutes from now, and if I can't find it, that will be indicated my no pushes to this PR since the last one.

Feel free to take your time and come back later. I will look into it tomorrow, God willing. And try to help if you hadn't figured it out till then.

src4026:
> > Note: There is an issue that if an HP marker is placed, it apparently doesn't set a marker cooldown (?). I tried seeing the code, but can't seem to find anything at first glance. I will be having a look at it for 10 minutes from now, and if I can't find it, that will be indicated my no pushes to this PR since the last one.
> 
> Feel free to take your time and come back later. I will look into it tomorrow, God willing. And try to help if you hadn't figured it out till then.

I tested it, and found that the timer works, but it doesn't print the message saying that "You can place markers every 5 seconds". The code seems fine. Is the way I have called the function the cause of the issue, or that I have mixed a loop block on accident with another one?

farooqkz:
Other than the code suggestion I made, I also suggest this:

``lua
if hpmarker then
    -- code A
else
   -- code B
end
```

rather than this:

```lua
if hpmarker == nil then
  -- code B
else
  -- code A
end
```

src4026:
> I tested it and LGTM
So the text not appearing is fine?

farooqkz:
> 

What text?

src4026:
> > 
> 
> What text?

The one that appears if you try to place markers before the time limit. The timer works and the markers aren't placed, but the message doesn't show for HP markers when I tested it.

src4026:
> I am looking forward to see the first code contribute of you merged :)

Me too :D

farooqkz:
> > > 
> > 
> > 
> > What text?
> 
> The one that appears if you try to place markers before the time limit. The timer works and the markers aren't placed, but the message doesn't show for HP markers when I tested it.

Ah I see.

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -187,6 +190,13 @@ local function marker_func(name, param, specific_player)
 	return true, "Marker is placed!"
 end
 
+minetest.register_chatcommand("hp", {
+	description = "Place a HP marker in your look direction",
+	params = "",
+	privs = {interact = true, shout = true},
+	func = function(name, param, specific_player) marker_func(name, param, specific_player, true) end
```
farooqkz, line 197:
```suggestion
	func = function(name, param, specific_player) return marker_func(name, param, specific_player, true) end
```

farooqkz:
@src4026 try my suggestion please

src4026:
> @src4026 try my suggestion please

Oh sorry, I didn't see.

src4026:
Ah. So it was indeed the way the function was called. I learned something :D

LoneWolfHT:
Are the issues listed in OP resolved then?

farooqkz:
yes as i recall. except that it is enabled for all modes.

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -128,56 +128,59 @@ local function marker_func(name, param, specific_player)
...
+		local pos1 = vector.offset(player:get_pos(), 0, player:get_properties().eye_height, 0)
 
-	local ray = minetest.raycast(
+		if param == "" then
+			param = "Look here!"
+		end
+		local ray = minetest.raycast(
 		pos1, vector.add(pos1, vector.multiply(player:get_look_dir(), MARKER_RANGE),
 		true, false
 	))
```
LoneWolfHT, line 145:
```suggestion

		local ray = minetest.raycast(
			pos1, vector.add(pos1, vector.multiply(player:get_look_dir(), MARKER_RANGE),
			true, false
		))
```

LoneWolfHT (commented):
One other option would be to add the player's HP to their marker if the marker is within ~2 blocks from them. So to request heals they can just mark at their feet

farooqkz:
> One other option would be to add the player's HP to their marker if the marker is within ~2 blocks from them. So to request heals they can just mark at their feet

I am unsure if I agree as it adds an extra step for placing a HP marker which is bad in the case of emergency.

farooqkz:
Another idea: Distinguish HP markers and normal markers in the chat.

src4026:
> Another idea: Distinguish HP markers and normal markers in the chat.

I was initially thinking about that too. Perhaps a cyan colour since it isn't used in CTF?

farooqkz:
> > Another idea: Distinguish HP markers and normal markers in the chat.
> 
> I was initially thinking about that too. Perhaps a cyan colour since it isn't used in CTF?

I would make it coloured like other markers but with a different message.

@LoneWolfHT What do you think?

src4026:
Proposal for message:
> \<player\> needs medical help! HP=<player's hp>

src4026:
> Proposal for message:
> \<player\> placed an HP marker. 
> \<player\> needs medical help! HP=<player's hp>

farooqkz:
"medical attention" sounds more natural to me than "medical help"

src4026:
Something like the above? Also, will HP markers be available for all the modes or be restricted to classes only?

src4026:
> "medical attention" sounds more natural to me than "medical help"

Yup 👍

LoneWolfHT:
> I am unsure if I agree as it adds an extra step for placing a HP marker which is bad in the case of emergency.

It would be one less step, because you'd be using the normal markers you're used to placing, you just place it at your feet.
The quick markers (Zoom+Mousebutton) would work for it too

LoneWolfHT:
> Something like the above? Also, will HP markers be available for all the modes or be restricted to classes only?

HP markers should only be allowed in the modes that have normal markers

LoneWolfHT:
> Another idea: Distinguish HP markers and normal markers in the chat.

Could probably do this by putting the marker's text in the chat message

farooqkz:
@src4026 @LoneWolfHT I think the idea of pointing to your feet for HP marker is a good idea. However, I think there must be the command there as well.

@LoneWolfHT If you agree, @src4026 can keep the current command and add what you suggested as well.

src4026:
The addition of the player points at their feet and places a marker I also feel is good as LV suggested. So I'll be adding that as an extension, isn't it?

farooqkz:
> The addition of the player points at their feet and places a marker I also feel is good as LV suggested. So I'll be adding that as an extension, isn't it?

You can simply check for pointed thing against players position and if it's close enough(as LV/LW suggested, less than 2 blocks), simply set the `hpmarker` to `true`.

LoneWolfHT:
Having the command still works. it might be a good idea to rename it to `/mhp` though

src4026:
Ready for review :D

src4026:
Also, I'll update the description in a minute or two. 👍

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -193,16 +190,27 @@ local function marker_func(name, param, specific_player, hpmarker)
 		else
 			pos = pointed.under
 		end
+
+	if (math.abs(pos.x - player:get_pos().x) <= 2) and (math.abs(pos.y - player:get_pos().y) <= 2) and (math.abs(pos.z - player:get_pos().z) <= 2) then
```
farooqkz:
```suggestion
	if vector.distance(pos, player:get_pos()) <= 2 then
```

src4026: Ah yes, of course MT has a way to directly find the vector distances.

src4026:
> ## Issues to be resolved and improvements to be made:
> - Colouring and chat message changes to be discussed upon.

So, is this what is left of now?

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -158,46 +160,61 @@ local function marker_func(name, param, specific_player)
...
+	else
+		return true, "Marker is placed!"
+	end
 end
 
+minetest.register_chatcommand("mhp", {
+	description = "Place a HP marker in your look direction",
+	params = "",
+	privs = {interact = true, shout = true},
+	func = function(name, param, specific_player) return marker_func(name, param, specific_player, true) end
```
LoneWolfHT, line 215: The `func` function of chatcommands only passes two arguments, did you test this code?

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -158,46 +160,61 @@ local function marker_func(name, param, specific_player)
...
+			pos = pointed.under
 		end
 
-		pos = obj:get_pos()
-		if concat then
-			message = message .. " <" .. concat .. ">"
-		end
-	else
-		pos = pointed.under
+	if vector.distance(pos, player:get_pos()) then
```
LoneWolfHT, line 192: This code currently checks if distance exists between the player and the marker pos. I think you forgot to add a `<= 2` here :P

LoneWolfHT:
> * Colouring and chat message changes to be discussed upon.

I answered this earlier:
> Could probably do this by putting the marker's text in the chat message

Coloring the marker messages differently can be done in a different PR

mods/ctf/ctf_modebase/markers.lua
src4026: IIRC that's how Farooq had made the marker changes to be to make the function reusable. There's a part of me still I'm confusion about this 😅

mods/ctf/ctf_modebase/markers.lua
src4026: Ah yes, my bad :D

src4026: But it worked when I was testing it lol. Perhaps somewhere in copy/pasting the code and changing for messed up.

src4026:
> > * Colouring and chat message changes to be discussed upon.
> 
> I answered this earlier:
> > Could probably do this by putting the marker's text in the chat message
> 
> Coloring the marker messages differently can be done in a different PR

So I suppose now all is well except that function you're talking about in the above review. :+1:

mods/ctf/ctf_modebase/markers.lua
farooqkz:
```suggestion
	func = function(name, param) return marker_func(name, param, nil, true) end
```

@src4026 did it wrong but because in Lua an argument not provided is `nil`, it didn't cause any problem and it works. This suggestion makes it better.

farooqkz (changes requested):
You need to make small changes as there are bugs.
Now, if the player tries to place HP marker using `mhp`, if they are not pointing to anything, no HP marker is placed and the player is told they are pointing nothing. You should move the code to check player distance and checking `hpmarker` for being `true` to the beginning of the function and immediately place the hp marker and return. This way first, if the player wanted to just place a HP marker, no calculation of ray stuff and such these is done in vain and second, the bug is easily solved.

mods/ctf/ctf_modebase/markers.lua
src4026: Sorry for that, commiting suggestion.

src4026:
~~There's literally a comment in the next line and I've tried to clear every whitespace~~

farooqkz:
@src4026 Ping me when you are finished.

src4026:
@farooqkz luacheck is giving an error although as far as I can see there is no trailing whitespace.

src4026:
Why did luacheck show an error where it was not located?

src4026:
Also, ty Farooq for fixing them :)

farooqkz:
For me luacheck did not false alarm. The problem could possibly be with your editor and/or its configuration. The main reason I did the changes was that you didn't seem to have fixed the bugs properly and I was too lazy too explain you what to do. So I did it myself in the hope that you realize where you were wrong by comparing my code and your code.

farooqkz:
@LoneWolfHT It's ready for review. But there's one last thing: What if the player is pointing to somewhere near(maximum two blocks away as you said) but pointing to an object? I think the HP marker must be placed only if the player is pointing to the node under.

src4026:
> For me luacheck did not false alarm. The problem could possibly be with your editor and/or its configuration. The main reason I did the changes was that you didn't seem to have fixed the bugs properly and I was too lazy too explain you what to do. So I did it myself in the hope that you realize where you were wrong by comparing my code and your code.

Yes, I do understand it now. And my editor has become a bit messy with the switching and everything.

LoneWolfHT:
> @LoneWolfHT It's ready for review. But there's one last thing: What if the player is pointing to somewhere near(maximum two blocks away as you said) but pointing to an object? I think the HP marker must be placed only if the player is pointing to the node under.

That's one solution, you could also just append the player's HP to that object marker too (`My HP: x/x` or something)

src4026:
> > @LoneWolfHT It's ready for review. But there's one last thing: What if the player is pointing to somewhere near(maximum two blocks away as you said) but pointing to an object? I think the HP marker must be placed only if the player is pointing to the node under.
> 
> That's one solution, you could also just append the player's HP to that object marker too (`My HP: x/x` or something)

So which one should I implement here? I'll do it after you guys agree on one 👍

LoneWolfHT:
The one I suggested will work even if there's an object at your feet, so that might be a better one

src4026:
I will do it tomorrow. Thanks for the suggestions and feedback :D

src4026:
I had been working on it but had some local conflicts with a lot of folder. I need more time, sorry for making wrong commitments regarding the time within which I will be able to complete it. Thanks

farooqkz:
No worries there is absolutely no rush :)

src4026:
There might be some redundancy, but I tried not resolving it for the time being. Also, I will be making one more commit, one sec :+1:

src4026:
#### Possible feature suggestion:
- If a player is pointing at another player with low HP, then the other player's HP will be appended to the marker.

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -160,50 +161,90 @@ local function marker_func(name, param, specific_player)
 
 	if pointed and pointed.type == "object" and pointed.ref == player then
 		pointed = ray:next()
+  end
+  if (vector.distance(pointed.under or pointed.ref:get_pos(), player:get_pos()) <= 2) and
+    (player:get_hp() < player:get_properties().hp_max) then
+		hpmarker = true
```
LoneWolfHT, line 167: Weird indentation here. `pointed` can also be nil here too I think

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -160,50 +161,90 @@ local function marker_func(name, param, specific_player)
...
+      if obj:is_player() then
+        concat = obj:get_player_name()
+      end
+      if entity then
+        if entity.name == "__builtin:item" then
+          local stack = ItemStack(entity.itemstring)
+          local itemdef = minetest.registered_items[stack:get_name()]
+          -- Fallback to itemstring if description doesn't exist
+          -- Only use the first line of itemstring
+          concat = string.match(itemdef.description or entity.itemstrng, "([^\n]+)")
```
LoneWolfHT, line 186: Misspelling of itemstring here. Did you modify any other parts of this? It's all shown as a change in Github so it's hard for me to tell

LoneWolfHT (changes requested):

mods/ctf/ctf_modebase/markers.lua
src4026: I did that to perhaps prevent the error that luacheck shows when a line is long.

src4026: Tested it and it is indeed `nil`. I will have to fix that :+1:

src4026: It should work now. Is the way I have done the proper way or is another better way possible?

LoneWolfHT: > I did that to perhaps prevent the error that luacheck shows when a line is long. Also, I did test the markers in-game and if I haven't missed a case, then they should work fine.

NEVER sacrifice indentation for luacheck warnings. Try putting a newline in these places 
image

LoneWolfHT: > It should work now. Is the way I have done the proper way, or is another better way possible?

That works, but are you going to have the same problem further down in the code?

src4026: Everything seems to be fine for the code below since if `hpmarker` is not `true`, it would go to default markers, where it would need to have the marker as pointed and if not pointed than would return false, which would exit the code before any such problem comes to view, isn't it? At least that is what I have understood from the code at present.

src4026:
Indentation seems to be bad still. What's wrong with my editor...making changes :+1:

src4026:
Done! I realised that the `end` above the next `if` statement messed up the indention whenever I made a newline. Phew

mods/ctf/ctf_modebase/markers.lua
src4026: Ah, so my editor kept adding weird indentations. I kept wondering how it kept going back a tab when I didn't do anything intentionally xD

mods/ctf/ctf_modebase/markers.lua
```diff
@@ -160,50 +161,91 @@ local function marker_func(name, param, specific_player)
...
-			message = message .. " <" .. concat .. ">"
-		end
+  ctf_modebase.markers.add(name, message, pos, nil, specific_player)
+  marker_cooldown:set(name, MARKER_PLACE_INTERVAL)
+  if hpmarker then
+		return true, "HP marker is placed!"
 	else
-		pos = pointed.under
+		return true, "Marker is placed!"
 	end
```
farooqkz, line 239: Indents are wrong. You've used 3 spaces instead of a tab

farooqkz (changes requested):

src4026:
Hmmm.

mods/ctf/ctf_modebase/markers.lua
src4026: Indentation is fixed now?

src4026: Are they fixed now?

LoneWolfHT: In VSCode you can run `Ctrl+Shift+P` and run `Convert Indentation to Tabs`

I'll check it when I do cleanup though

LoneWolfHT (approved):
Looks ok, I'll go through and test/cleanup before merging though

mods/ctf/ctf_modebase/markers.lua
LoneWolfHT: I think you could get a crash if hpmarker is true and pointed_thing is nil. Added a check for that for now, I'll make sure in my cleanup review though

src4026:
> @src4026 
> As a programmer/developer, you must learn the tools you use and utilizing them. This includes your Operating System, your web browser and finally... your development environment. You must learn about your development environment enough to have control over it. And you must know exactly what are you doing with it. Like what will happen if you use X key combination. You also have freedom to choose your tools. If you are not fine with an editor or IDE you use because it has bugs or the UI it has does not suit you, feel free to try another one.
> 
> For instance, I want everything to be done with my keyboard. I am not comfortable with mouse unless it can really do the job much better than a keyboard(like playing Minetest!). Therefore, I use i3 WM to switch quickly between open windows. And I use vim which its philosophy is no mouse.

I have been having a long journey of changing IDEs, and making this PR is in the middle of it, thus this chaos. I have been trying to find the editor that truly is the one I can find, andI did find it. But due to my current software conflicts, I am unable to use it, that is, Doom Emacs. I had to use VSCodium for the time being but decided to move to something else. I then chose Neovim's NvChad for a start. (I had used Neovide prior to this but due to the limitations of the apt repository, some packages were outdated). When I used nvim, the lua linter was fantastic! But the way it handled tabs was extremely weird, which happened to be the case here. I am using VSCodium at the moment with the Lua Language server installed, but the linter is something I haven't been able to figure out. Could anyone please help me with it?

src4026:
> > @src4026 
> > As a programmer/developer, you must learn the tools you use and utilizing them. This includes your Operating System, your web browser and finally... your development environment. You must learn about your development environment enough to have control over it. And you must know exactly what are you doing with it. Like what will happen if you use X key combination. You also have freedom to choose your tools. If you are not fine with an editor or IDE you use because it has bugs or the UI it has does not suit you, feel free to try another one.
> > 
> > For instance, I want everything to be done with my keyboard. I am not comfortable with mouse unless it can really do the job much better than a keyboard(like playing Minetest!). Therefore, I use i3 WM to switch quickly between open windows. And I use vim which its philosophy is no mouse.
> 
> I have been having a long journey of changing IDEs, and making this PR is in the middle of it, thus this chaos. I have been trying to find the editor that truly is the one I can find, andI did find it. But due to my current software conflicts, I am unable to use it, that is, Doom Emacs. I had to use VSCodium for the time being but decided to move to something else. I then chose Neovim's NvChad for a start. (I had used Neovide prior to this but due to the limitations of the apt repository, some packages were outdated). When I used nvim, the lua linter was fantastic! But the way it handled tabs was extremely weird, which happened to be the case here. I am using VSCodium at the moment with the Lua Language server installed, but the linter is something I haven't been able to figure out. Could anyone please help me with it? 

Also, I do understand the freedom and have much explored with my forage on Linux. Everything I use has something to it. Thank you for reminding me :)

farooqkz:
@src4026 Please do re-base only on request. It is annoying to have notifications for the repo watchers just for unnecessary rebases.

src4026:
> @src4026 Please do re-base only on request. It is annoying to have notifications for the repo watchers just for unnecessary rebases.

Oh sorry. I didn't know it was sent to watchers.

LoneWolfHT:
There's no rule against it or anything, but it annoys people that like an empty notification queue :P

I usually commit the rebase without pushing it. That way I get my rebase locally and it'll show up on the repo whenever I push other changes that watchers might be more interested in

LoneWolfHT (changes requested):
* Need to remove the `[]` around the numbers in the HP markers.
* Markers should also show HP of marked players if they're not full health

sonicSMASH:
How about an additional argument that specifies which player to show it to (like /mp <player>)

src4026:
> How about an additional argument that specifies which player to show it to (like /mp <player>)
>  

I didn't understand what you meant by this. Could you elaborate please?

sonicSMASH:
> > How about an additional argument that specifies which player to show it to (like /mp )
> 
> I didn't understand what you meant by this. Could you elaborate please?

The command /mp allows a marker to be placed, but it's only visible to the player added after the command. You could implement this into the HP markers so that everyone on your team doesn't need to see it.

src4026:
> > > How about an additional argument that specifies which player to show it to (like `/mp` )
> > 
> > I didn't understand what you meant by this. Could you elaborate please?
> 
> The command `/mp` allows a marker to be placed, but it's only visible to the player added after the command. You could implement this into the HP markers so that everyone on your team doesn't need to see it.

Ah, then I suppose it could be good as well. What are your thoughts on this @farooqkz and @LoneWolfHT ?

LoneWolfHT:
Maybe to all medics in your team. But that should be done in another PR if you want

LoneWolfHT (approved):

PR #1166, opened 2023-07-02
Add per map init.lua execution

FlamingRCCars:
This will allow adding custom things on a per-map basis.


LoneWolfHT:
This is merged, but not intended to be used for now

PR #1167, opened 2023-07-02
Add a list containing fall damage factors

FlamingRCCars:
This is a basic mod that allows you to individually set the fall damage factors of each block, I have 2 example values in here, but they can be changed. 

Voted 14-1 on CTF suggestions channel
https://discord.com/channels/447819017391046687/448136308108296213/1124496383634833509


farooqkz:
Related: https://discord.com/channels/447819017391046687/1008040841102766153/1124616741268496416

farooqkz:
Shouldn't this be in `mtg/ctf_changes`? ping @LoneWolfHT

LoneWolfHT:
Yeah, if you don't mind creating a second file for your mod in there, or just adding the code to the already existing init.lua

FlamingRCCars:
@LoneWolfHT Changes merged, should be good now.

farooqkz (approved):
The code LGTM but unsure about the value of damage.

FlamingRCCars:
Yep, those values are temporary place holders

mods/mtg/ctf_changes/init.lua
```diff
@@ -2,6 +2,11 @@ local COOLDOWN = ctf_core.init_cooldowns()
 
 local DISALLOW_MOD_ABMS = {"default", "fire", "flowers", "tnt"}
 
+local node_fall_damage_factors = {
+    {"default:snowblock", -14}, -- From a height of 13 blocks you take 5 damage
```
LoneWolfHT, line 6: This could be `["default:snowblock"] = -14,` and all other table elements could follow the same pattern

mods/mtg/ctf_changes/init.lua
```diff
@@ -177,3 +182,13 @@ minetest.override_item("default:furnace_active", {
 	can_dig = function() return true end,
 	on_destruct = furnace_on_destruct,
 })
+
+minetest.register_on_mods_loaded(function()
+    for _, entry in ipairs(node_fall_damage_factors) do
```
LoneWolfHT, line 187: You could then change this to
`for nodename, reduction in ipairs(node_fall_damage_factors) do`

LoneWolfHT (changes requested):

mods/mtg/ctf_changes/init.lua
farooqkz: Why `pairs` instead of `ipairs` here?

LoneWolfHT: `ipairs` can only loop over lists, in lists the keys are `1, 2, 3, 4, ...`, but I changed the table we're looping over to use named keys, which makes it into a normal table, not a list table. `pairs` loops over both lists and normal tables

LoneWolfHT (approved):

PR #1173, opened 2023-07-03
give score for items removed from enemies

farooqkz:
This PR gives score to the players if they put an item dropped from teammate or enemy and put it in the teamchest. The item must not be stackable.


AltZ-15:
What stops players from just cashing in all the items they've collected and talking it all back from the chest as soon as they get the points?

farooqkz:
> What stops players from just cashing in all the items they've collected and talking it all back from the chest as soon as they get the points?

It's a reasonable concern. But why they would do that?

mods/ctf/ctf_teams/team_chest.lua
```diff
@@ -4,6 +4,58 @@ local blacklist = {
...
+	["ctf_ranged:smg"] = 4,
+	["ctf_ranged:smg_loaded"] = 5,
+	["ctf_ranged:sniper_magnum"] = 8,
+	["ctf_ranged:sniper_magnum_loaded"] = 10,
+	["ctf_ranged:ammo"] = 4,
+	["default:diamond"] = 2.5,
+	["default:mese_crystal"] = 2,
+	["default:mese"] = 18,
+	["default:steel_ingot"] = 1,
+	["default:iron_lump"] = 1,
```
LoneWolfHT, line 24: Stackable

mods/ctf/ctf_teams/team_chest.lua
```diff
@@ -4,6 +4,58 @@ local blacklist = {
...
+	["ctf_map:reinforced_cobble"] = 3,
+	["ctf_map:damage_cobble"] = 3,
+	["ctf_map:unwalkable_cobble"] = 1,
+	["ctf_map:unwalkable_stone"] = 1,
+	["ctf_map:unwalkable_dirt"] = 1,
+	["default:steelblock"] = 2.5,
+	["default:bronzeblock"] = 2.5,
+	["default:obsidian_block"] = 3.5,
+	["ctf_map:spike"] = 2.5,
+	["default:apple"] = 1.5,
```
LoneWolfHT, line 49: Stackable

mods/ctf/ctf_teams/team_chest.lua
```diff
@@ -245,6 +297,15 @@ for _, team in ipairs(ctf_teams.teamlist) do
 			stack:to_string(),
 			minetest.pos_to_string(pos)
 		))
+		local meta = stack:get_meta()
+		if meta:get_string("dropped") == "true" then
+			local pname = player:get_player_name()
+			if pname then
+				local score = (item_value[stack:get_name()] or 0) * stack:get_count()
+				ctf_modebase:get_current_mode().recent_rankings.add(pname, { score = score }, false)
```
LoneWolfHT, line 305: Make sure this is crash-safe, there are times where there isn't an ongoing mode and `ctf_modebase:get_current_mode()` returns false/nil

mods/ctf/ctf_teams/team_chest.lua
```diff
@@ -245,6 +297,15 @@ for _, team in ipairs(ctf_teams.teamlist) do
...
+			end
+			meta:set_string("dropped", "")
```
LoneWolfHT, line 307: Make sure this change is actually getting applied to the stack you put in.
I'm not sure if `stack` is a deep copy of the actual stack or not.

mods/pvp/dropondie/init.lua
```diff
@@ -2,6 +2,10 @@ dropondie = {}
 
 local function drop_list(pos, inv, list)
 	for _, item in ipairs(inv:get_list(list)) do
+		if minetest.registered_items[item:get_name()].stack_max == 1 then
```
LoneWolfHT, line 5: I'd be slightly worried about `stack_max` being nil/nonexistent here

LoneWolfHT (commented):
You could replace the meta string `"dropped"` with an int value. `meta:get_int()` returns 0 for unset values, and you could treat 1 as truthy
`meta:set_int("dropped", 1)`, etc

LoneWolfHT:
To play it safe I would prefer to only give 1 or less score for items we reward returning, and slowly increase that as we watch things play out on the main server

mods/ctf/ctf_teams/team_chest.lua
farooqkz: I have assigned values to items for further use. The item metadata is added to an item only if it's not stackable.

mods/ctf/ctf_teams/team_chest.lua
farooqkz: You are right. The metadata is not set so score is given if the player takes and puts over again and again

mods/pvp/dropondie/init.lua
farooqkz: I'll do a research about that.

farooqkz:
> To play it safe I would prefer to only give 1 or less score for items we reward returning, and slowly increase that as we watch things play out on the main server

I disagree with the reward being the same, one point, for all the items. And less than one is a bad idea IMO

mods/pvp/dropondie/init.lua
farooqkz: On second thought, why should it matter if `stack_max` is `nil`?

farooqkz: image

LoneWolfHT:
> I disagree with the reward being the same, one point, for all the items. And less than one is a bad idea IMO

Did you read the `and slowly increase that as we watch things play out on the main server` part?

farooqkz:
> > I disagree with the reward being the same, one point, for all the items. And less than one is a bad idea IMO
> 
> Did you read the `and slowly increase that as we watch things play out on the main server` part? The item values can be differentiated again as we increase them

Yeah I know but it doesn't make sense to me even temporary. But if you mean some stuff can be temporary non-sensible, that's your decision. I am committing to give exactly one for all items.

mods/pvp/dropondie/init.lua
LoneWolfHT: :+1:

LoneWolfHT: > On second thought, why should it matter if `stack_max` is `nil`?

Comparing numbers with nil causes a crash

mods/ctf/ctf_teams/team_chest.lua
LoneWolfHT: You can fix this by setting the stack in the destination index to your stack variable

LoneWolfHT:
Thanks.
https://github.com/MT-CTF/capturetheflag/pull/1173#discussion_r1254898943 needs to be fixed and this will be ready for testing

mods/pvp/dropondie/init.lua
farooqkz: you sure?

```
13:33:49 ~ $ lua
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> local a = nil
> if a == 1 then print("H") end
> if b == 1 then print("H") end
> 
```

LoneWolfHT: Ah, `==` doesn't count as a comparison (unlike `>=`, etc) it seems. Good to know.

LoneWolfHT:
Your rebase didn't work at all there

LoneWolfHT:
With rebasing I tend to use `git rebase -i <PR branch>~<number of previous commits I want to have the option to squash>` (e.g `git rebase -i master~6`), squash all my commits together, then `git pull -r mainrepo master` to rebase off of the CTF repo's master branch

I'm laying that out to show you some things you can do, I highly recommend googling/figuring out what those commands do on your own though

LoneWolfHT:
Items returned to team chest shouldn't give score if they're your own

farooqkz:
@LoneWolfHT Done.

LoneWolfHT (approved):
Rebase conflict

farooqkz:
> Rebase conflict

WDYM?

LoneWolfHT:
image

farooqkz:
@LoneWolfHT Done

LoneWolfHT:
How are you rebasing your PR? This shouldn't be showing up: 
image

farooqkz:
@LoneWolfHT What was state of this? Do I need to do anything?

PR #1174, opened 2023-07-03
Add the "Thief" class

### LINKED ISSUE #1172: Thief class
Adding thief class to Classes mode. A thief cannot use any weapon. Upon right clicking on some enemy with the special tool, a random slot from enemy's inventory is chosen and if not empty and not class item, it is added to thief's inventory. A thief has 1.5x speed, as much as medic's HP.

Addition of this class is dependent on #1010

FlamingRCCars:
Fixes #1172, I nerfed it a little to make it more balanced, and hopefully farooq can use it to add his score thingy, but I tested it, and it works.


FlamingRCCars:
Landar suggests waiting to add more classes until we get flagwars added, as such I'm closing this.

PR #1176, opened 2023-07-04
Update per map init.lua execution

165135:
Fix for bug where every map init.lua would execute at the start of the match.


LoneWolfHT:
This PR *makes it* run at the start of every match?!

FlamingRCCars:
@165135 it would appear you are a little lost, per map init.lua files are only supposed to be executed when the server is starting, hence `init.lua`, what you have here will make it run at the start of every match, which would create a sad case of crashing.

LoneWolfHT:
This feature hasn't been approved for official maps yet either, wait until instructions on its use are added to the map documentation

PR #1180, opened 2023-07-12
add bonus for killing a flag carrier

lazylier:
This is supposed to multiply the kill score by 1 + .25 for each flag held by the victim.

### DIFF
--- a/mods/ctf/ctf_modebase/features.lua
+++ b/mods/ctf/ctf_modebase/features.lua
@@ -9,8 +9,13 @@ local teams_left
 local function calculate_killscore(player)
 	local match_rank = recent_rankings.players()[player] or {}
 	local kd = (match_rank.kills or 1) / (match_rank.deaths or 1)
-
-	return math.max(1, math.round(kd * 7))
+	local flag_multiplier = 1
+	for tname, p in pairs(ctf_modebase.flag_taken) do
+		if p.p == player then
+			flag_multiplier = flag_multiplier + .25
+		end
+	end
+	return math.max(1, math.round(kd * 7 * flag_multiplier))
 end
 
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_modebase/features.lua
```diff
@@ -9,8 +9,13 @@ local teams_left
 local function calculate_killscore(player)
 	local match_rank = recent_rankings.players()[player] or {}
 	local kd = (match_rank.kills or 1) / (match_rank.deaths or 1)
-
-	return math.max(1, math.round(kd * 7))
+	local flag_multiplier = 1
+	for tname, p in pairs(ctf_modebase.flag_taken) do
+		if p.p == player then
```
farooqkz, line 14: Use a better variable name than `p`

mods/ctf/ctf_modebase/features.lua
```diff
@@ -9,8 +9,13 @@ local teams_left
...
+			flag_multiplier = flag_multiplier + .25
```
farooqkz, line 15:
```suggestion
			flag_multiplier = flag_multiplier + 0.25
```

Because it's easier to read

farooqkz (changes requested):

farooqkz:
@lazylier For the next time, you can simply commit my suggestion from Github UI.

@LoneWolfHT looks good to me. Except that the multiplier could be further discussed.

LoneWolfHT (approved):

LoneWolfHT:
This can be merged when I reset ranks on August 16

LoneWolfHT:
This doesn't benefit from a testing session, as it's pretty simple

PR #1183, opened 2023-07-13
Fix poison nades disappearing

165135:
Drop all items before respawn to clear bugged poison smoke nades in inventory if they are thrown near enemy base and then the player dies before they land.

### DIFF
--- a/mods/ctf/ctf_modebase/mod.conf
+++ b/mods/ctf/ctf_modebase/mod.conf
@@ -1,2 +1,2 @@
 name = ctf_modebase
-depends = ctf_api, ctf_core, ctf_teams, ctf_gui, ctf_map, ctf_healing, crafting, mhud, default, binoculars, dropondie
+depends = ctf_api, ctf_core, ctf_teams, ctf_gui, ctf_map, ctf_healing, crafting, mhud, default, binoculars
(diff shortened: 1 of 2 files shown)


farooqkz (changes requested):
Not sure it's the best solution but I guess it works anyway. You should add the mod `dropondie` to dependencies of `ctf_modebase` in `mod.conf`

165135:
> Not sure it's the best solution but I guess it works anyway. You should add the mod `dropondie` to dependencies of `ctf_modebase` in `mod.conf`

image
I'm not familiar with dependencies, but it causes an error, maybe because dropondie has ctf_modebase as a dependency also

165135:
Well, at least now it won't _crash_

farooqkz:
> Well, at least now it won't _crash_

If you are making it optional, the code should handle if there is no dropondie.

LoneWolfHT:
This is a bad approach, `ctf_modebase` shouldn't try to do anything with `dropondie`, it should provide an API that dropondie can use, but that's it.

Try checking player hp when the poison grenade is added back to the players inventory. The hp should be 0 if the player is dead/respawning, so you can just drop the grenade where it landed

LoneWolfHT:
Probably going to merge https://github.com/MT-CTF/capturetheflag/pull/1201 unless you want a different approach

PR #1186, opened 2023-07-15
Adding 'all' option to `/r`

src4026:
For reference: https://discord.com/channels/447819017391046687/1008040841102766153/1116006701636780184
Is this how I should do it, or is another way better?


src4026:
Why is it showing `pname` as undefined variable, although the code for it seems the same for me in the actual file and the one in this PR?

165135:
> Why is it showing `pname` as undefined variable, although the code for it seems the same for me in the actual file and the one in this PR?

I think it is because `pname` isn't passed into the function `get_rank` directly, maybe try putting it as an argument

src4026:
> > Why is it showing `pname` as undefined variable, although the code for it seems the same for me in the actual file and the one in this PR?
> 
> I think it is because `pname` isn't passed into the function `get_rank` directly, maybe try putting it as an argument 

Thanks! ~~I'll mention it here too I guess~~ :)

farooqkz:
@src4026 Try this approach instead:

 - Make the rank command function a named local function and assign it as function of the command
 - After getting information from the `params`, check if mode name is `all`
 - If it is all, create an empty `return_str`, iterate over all modes and append their returning str, if any, to the return_str, you have
 - Finally return this `return_str`
 
That is, you call the rank command function from itself.  Please try and if you got lost, let me know :)

src4026:
Ready for review.

PR #1187, opened 2023-07-19
Make /t command easier

Jo5629:
I wanted to make it so where the /t command could become easier. The command would be activated by putting a semicolon before the message.

This is so mobile players would have easier functionality when talking to their own team specifically.

All that I did was copy & paste the code from the /t and changed it to fit what I had wanted.

### DIFF
--- a/mods/ctf/ctf_chat/init.lua
+++ b/mods/ctf/ctf_chat/init.lua
@@ -116,11 +116,11 @@ end)
 minetest.register_on_chat_message(function(name, message)
 	local check_message = string.sub(message, 1, 1)
 
-	if message:len() < 2 and check_message == "!" then
+	if message:len() < 2 and check_message == ":" then
 		return false
 	end
 
-	if check_message == "!" then
+	if check_message == ":" then
 		local message = string.sub(message, 2, string.len(message))
 		local tname = ctf_teams.get(name)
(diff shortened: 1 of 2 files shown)


LuismiInTheZone:
Two things:
1. I think /t is pretty easy
2. I haven't tested it but how do you intend to use the a on the keyboard? just for example: a hello team? So what if I want to put a teammate is killing everyone to the public?

Jo5629:
My Answers:
>1. I think /t is pretty easy

Fair enough. I see what you mean, but I wanted to make it more simple. Also, I can not disagree with your opinion for the /t command being easy to execute.

>2. I haven't tested it but how do you intend to use the a on the keyboard? just for example: a hello team? So what if I want to put a teammate is killing everyone to the public?

Sorry for the mixup. I meant for it to be the **semicolon** to activate the team channel function, not the a key. My fault for not making it clear. Thank you for pointing out an issue in the comment.

Jo5629:
>1. I think /t is pretty easy

> 2. I haven't tested it but how do you intend to use the a on the keyboard? just for example: a hello team? So what if I want to put a teammate is killing everyone to the public?

Sorry I do not think I made the point clear. I wanted to use the colon key, not the a key. My fault for not making it clear.

PR #1188, opened 2023-07-25
Adding a command that prints the current mode: `/mode`

src4026:


LoneWolfHT (approved):
Note the changes I made. I only thought of listing the matches played when I reviewed this though

LoneWolfHT:
> I just noticed them. Also what does the `HumanReadable()` do?

It changes things like `nade_fight` into `Nade Fight`

LoneWolfHT:
> Well, I had one small reason for keeping it like that, however the Human Readable one is better anyways 🤷. 👍

I expect people who need the technical name to know how to convert it, but the technical name could be added in parenthesis after it I guess

PR #1194, opened 2023-08-02
Victory score

lazylier:
This has been suggested a few times I think, most recently [here.](https://discord.com/channels/447819017391046687/448136308108296213/1135090580272205874) Each player on a winning team gets score based on how many points he's contributed to that team relative to everyone else.

### DIFF
--- a/mods/ctf/ctf_modebase/features.lua
+++ b/mods/ctf/ctf_modebase/features.lua
@@ -423,6 +423,14 @@ return {
 			capture_reward = capture_reward + score
 		end
 
+		for teammate in pairs(ctf_teams.online_players[pteam].players) do
+			if teammate ~= pname then
+				local teammate_value = (recent_rankings.get(teammate)[pteam.."_score"] or 0) / (team_scores[pteam].score or 1)
+				local victory_bonus = capture_reward * teammate_value
+				recent_rankings.add(teammate, {score = victory_bonus}, true)
+			end
+		end
+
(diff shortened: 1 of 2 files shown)


farooqkz:
I am very doubtful about this. Regarding the incentive you are proposing, it is still to advantage of a player to let their teammate who has the flag die. Because when the player themselves cap, they will get this "victory score" *and* capture score.

lazylier:
> Because when the player themselves cap, they will get this "victory score" _and_ capture score.

That is not how it's supposed to work. You only get the victory score if you didn't capture.

farooqkz:
> > Because when the player themselves cap, they will get this "victory score" _and_ capture score.
> 
> That is not how it's supposed to work. You only get the victory score if you didn't capture.

oh so I misunderstood. But still, it's not a good incentive IMO for the reasons I will explain later.

165135:
> > > Because when the player themselves cap, they will get this "victory score" _and_ capture score.
> > 
> > 
> > That is not how it's supposed to work. You only get the victory score if you didn't capture.
> 
> oh so I misunderstood. But still, it's not a good incentive IMO for the reasons I will explain later.

@farooqkz Could you elaborate on your reasoning? Just curious.

farooqkz:
I don't exactly remember my reasons, now. And after reading your code, I think it's fine  and a good addition to the game :)

165135:
@LoneWolfHT Was this supposed to be added after the reset?

PR #1195, opened 2023-08-03
Disable usage of classes abilities during build time

165135:
Disable all abilities for classes instead of just paxel usage
From suggestion: https://discord.com/channels/447819017391046687/448136308108296213/1135607133661044788

### DIFF
--- a/mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
+++ b/mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
@@ -151,7 +151,7 @@ ctf_melee.simple_register_sword("ctf_mode_classes:knight_sword", {
 
 		local pname = user:get_player_name()
 
-		if itemstack:get_wear() == 0 then
+		if itemstack:get_wear() == 0 and ctf_modebase.match_started then
 			local step = math.floor(65534 / KNIGHT_USAGE_TIME)
 			ctf_modebase.update_wear.start_update(pname, "ctf_melee:sword_diamond", step, false, function()
 				local player = minetest.get_player_by_name(pname)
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
```diff
@@ -151,7 +151,7 @@ ctf_melee.simple_register_sword("ctf_mode_classes:knight_sword", {
 
 		local pname = user:get_player_name()
 
-		if itemstack:get_wear() == 0 then
+		if itemstack:get_wear() == 0 and ctf_modebase.match_started then
```
farooqkz, line 154: This way the player will get the "Can't use during build time" message even when match has started but item's wear is 0.

mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
```diff
@@ -222,15 +229,22 @@ ctf_ranged.simple_register_gun("ctf_mode_classes:ranged_rifle", {
 			return
 		end
 
-		if itemstack:get_wear() == 0 then
+		if itemstack:get_wear() == 0 and ctf_modebase.match_started then
```
farooqkz, line 232: Same as L154

farooqkz (changes requested):

mods/ctf/ctf_modes/ctf_mode_classes/classes.lua
LoneWolfHT: Not fixed

LoneWolfHT:
I think this has already been done, closing

PR #1196, opened 2023-08-03
add flag attempt streaks

farooqkz:
This only adds flag attempt streaks. When a player does an amount of attempts in 10 minutes or less, they get a streak which is announced with chat to all players. The period, 10 minutes, and the colour of the streak message can be further discussed.


LoneWolfHT:
Conflicts need resolving. I see some debug logging in your code still too

mods/ctf/ctf_modebase/flags/taking.lua
```diff
@@ -78,6 +78,48 @@ function ctf_modebase.flag_on_punch(puncher, nodepos, node)
 		table.insert(ctf_modebase.taken_flags[pname], target_team)
 		ctf_modebase.flag_taken[target_team] = {p=pname, t=pteam}
 
+
+		if ctf_modebase.flag_attempt_history[pname] == nil then
+			ctf_modebase.flag_attempt_history[pname] = {}
+		end
+		table.insert(ctf_modebase.flag_attempt_history[pname], minetest.get_gametime())
```
LoneWolfHT, line 85: This should use actual time, via a lua feature or `minetest.get_us_time()`

farooqkz: done

PR #1201, opened 2023-08-10
Drop the nade with illegal explode points if the thrower is dead

### LINKED ISSUE #1160: The smoke and poison grenades vanish if thrown right before death near bases
Steps to reproduce:

 - Throw a poison or smoke nade near one of the bases where it can't be thrown
 - Quickly before the nade lands, die
 - The nade will return to your now dead inventory
 - You respawn and the nade will vanish

Emojigit:
Fixes #1160

This PR add checks to player HP before adding nades back to thrower's inventory. If the player is head, the nade is dropped at its explode point.

This PR is ready for review.

### DIFF
--- a/mods/pvp/grenades/grenades.lua
+++ b/mods/pvp/grenades/grenades.lua
@@ -166,7 +166,7 @@ local register_smoke_grenade = function(name, description, image, damage)
 						if distance_from_flag <= 15 and (damage or pteam == flagteam) then
 							minetest.chat_send_player(pname, "You can't explode smoke grenades so close to a flag!")
 							if player:get_hp() <= 0 then
-								-- Drop the nade at its explode point if the thrower is head
+								-- Drop the nade at its explode point if the thrower is dead
 								-- Fixes https://github.com/MT-CTF/capturetheflag/issues/1160
 								minetest.add_item(pos, ItemStack("grenades:"..name))
 							else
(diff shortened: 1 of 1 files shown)


farooqkz:
Code LGTM and solves the problem. However, during testing, I discovered another bug :D https://github.com/MT-CTF/capturetheflag/issues/1203

farooqkz:
> Fixes #1160 Fixes #1203
> 
> This PR add checks to player HP before adding nades back to thrower's inventory. If the player is head, the nade is dropped at its explode point.
> 
> This PR is ready for review.

And if the player respawns before the nade lands, their HP is not zero and therefore the newly found bug is not solved.

LoneWolfHT:
That bug can be solved in a separate PR if you want

Emojigit:
> That bug can be solved in a separate PR

The logic of solving #1203 may can also solve this problem, so they will be the same code. However, the fix isn't trivial, and I don't have a plan to work on it soon. This PR is able to merge, but the codes inside this PR may be replaced in the future.

LoneWolfHT:
I'll leave this for now then

farooqkz:
> I'll leave this for now then

What do you mean?



> That bug can be solved in a separate PR if Emojigit wants

Yeah no reason to make this PR wait.

PR #1208, opened 2023-08-12
Donate function params

165135:
Allow for user to input the donate function params backwards


LoneWolfHT:
Please use tabs for indentation next time, also note the changes I made

PR #1210, opened 2023-08-16
Make it possible to send PMs through /pm

a-blob:


farooqkz:
I don't see a reason why we need this.

a-blob:
Since it says "PM from" when you receive a /msg, many people receiving it the first time will try /PM to reply.

farooqkz:
> Since it says "PM from" when you receive a /msg, many people receiving it the first time will try /PM to reply.

Yeah but still I don't see any reason. We could simply add a hint when they receive a PM `reply with /msg` or even a new random message.

a-blob:
How about "msg from...."?

farooqkz:
> How about "msg from...."? Adding `reply with /msg` every time seems a bit spammy.

A random message then?

a-blob:
I was thinking that maybe there could be a list opened by `/tips` that lists things such as how to change classes, how to send team messages, how to send PMs, etc... Random messages are easy to ignore and they only appear sometimes.

farooqkz:
> I was thinking that maybe there could be a list opened by `/tips` that lists things such as how to change classes, how to send team messages, how to send PMs, etc... Random messages are easy to ignore and they only appear sometimes.

We could show all random messages in an Inventory tab.

src4026:
> things such as how to change classes, how to send team messages, how to send PMs, etc... 
Perhaps a beginner's guide/ that can be accessed upon first joining. I am thinking of something like this, since we're on the topic of tips:
- An inventory tab that includes basic things in section like:
    - How to Play:
    - Modes
    - Learning
    - Using commands
- As for the "Using commands" section, I was thinking of an alternative method of making it an extra inventory tab listing out commands that are available to the normal CTF player, since using /help is time taking and can be made easier.
- This could be showed upon joining and be connected to the rules also, I haven't fully thought the process out in my head though, so it is subject to discussion as well.
Is this important enough to make an Issue for?

a-blob:
Showing all random messages might be a bit clogged and confusing. I like s20's idea better.

a-blob:
I drafted this simple mod for the game that adds a "Tips" inventory tab https://github.com/a-blob/capturetheflag/tree/inventory_tips/mods/other/inventory_tips
Of course both the title and the content in the tab is highly subject to change. The current content is a few selected random messages simplified to be shorter.
I'll close this PR for now and maybe make another one when the tips is done. In the mean-time feel free to create PRs there.

PR #1213, opened 2023-08-18
Add two random messages about `/league` and `/leagues` and a few edits

src4026:
In this PR, I have added two random messages about the two new commands plus three edits regarding commans as verified by my grammar editor.


farooqkz (approved):
LGTM

PR #1214, opened 2023-08-19
Add a random message about sending PMs to players.

a-blob:


LoneWolfHT:
This has conflicts that need to be resolved

a-blob:
Hmm, I just noticed the conflict. Should be solved now.

PR #1215, opened 2023-08-20
Adding a command that prints the current map: /map

Lucyucy:
Prints the current map name and map author


src4026:
How about also adding the map duration? I remember LandarVargan saying it would be nice to do so with a command like `/map` when I once had a go with it, but left it for later.

olliy1or:
Also isn't there already a command named /map?

PR #1218, opened 2023-08-24
Alterations to water and lava borders

Emojigit:
Two new commands are added:
* `/ctf_map fixborder`: Selecting two coordinates, and then convert the old barriers within that area into their new variants.
* `/ctf_map setborder`: Turn the selected area into the border. It works by looking up a reverse map of `ctf_map.barrier_nodes`.

### DIFF
--- a/mods/ctf/ctf_map/editor_functions.lua
+++ b/mods/ctf/ctf_map/editor_functions.lua
@@ -206,6 +206,65 @@ ctf_map.register_map_command("setbarrier", function(name, params)
 		"Select two corners of where the barrier should cover.")
 end)
 
+local function clearbarrier(player)
+	minetest.chat_send_player(player, minetest.colorize(ctf_map.CHAT_COLOR,
+		"Clearing barriers. This will take a few seconds..."))
+
+	-- From map_functions.lua
+	local pos1 = getpos_players[player].positions[1]
+	local pos2 = getpos_players[player].positions[2]
+	pos1, pos2 = vector.sort(pos1, pos2)
+
+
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
You can remove the air barrier code, but leave the new command

Emojigit:
@LoneWolfHT Changes done. Ready for review.

mods/ctf/ctf_map/nodes.lua
```diff
@@ -72,7 +72,9 @@ minetest.register_node("ctf_map:ind_lava", {
 	description = "Indestructible Lava Barrier Stone",
 	groups = {immortal = 1},
 	tiles = {"ctf_map_ind_lava.png"},
-	is_ground_content = false
+	is_ground_content = false,
+	paramtype = "light",
+	light_source = default.LIGHT_MAX - 1,
```
LoneWolfHT, line 77: I think there is a `minetest.LIGHT_MAX` which should be used instead

LoneWolfHT:
Might be worth adding a `remove_barrier` command to convert barrier nodes back. Doesn't have to be in this PR though

Emojigit:
> Might be worth adding a `clearbarrier` command to convert barrier nodes back. Doesn't have to be in this PR though
> 
> I'll rename `border` to `barrier` in your command names pre-merge unless you do that before me

All these done.

Emojigit:
This PR is also worth closing for a clearer commit history and target. I will review and probably rewrite the codes introduced in the PR before re-opening another PR.

PR #1220, opened 2023-08-25
Changes to HP markers and miscellaneous additions to `random_messages`

src4026:
# Changes
- HP markers now append itemstrings.
- When HP is `0`, i.e., when a player places a marker upon death, is shows a different message: `"<player> died here"`
- Adds random messages about `/mhp`/HP markers, `/map` command, `mode:all` addition to the `/r` command.

### DIFF
--- a/mods/other/random_messages/init.lua
+++ b/mods/other/random_messages/init.lua
@@ -77,7 +77,7 @@ function random_messages.read_messages()
 		"Use /leagues to list leagues and the placement needed to get to them.",
 		"Use /mhp or left/right click while holding the zoom key (default: Z)"
 		.. "pointed towards your feet to place an HP marker.",
-		"Use /map to print the current map name and map author.",
+		"Use /map to print the current map name and its author.",
 		"To check someone's ranking in all modes, use /rank mode:all <player_name>",
 	}
 end
--- a/mods/other/random_messages/init.lua
+++ b/mods/other/random_messages/init.lua
@@ -75,7 +75,7 @@ function random_messages.read_messages()
 		"Use /mp <player> to send a marker to a specific teammate",
 		"To check someone's league in all modes, use /league <player>",
 		"Use /leagues to list leagues and the placement needed to get to them.",
-		"To place an HP marker run /mhp or left/right click while holding the zoom key (default: Z) "..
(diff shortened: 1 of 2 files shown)


farooqkz:
@src4026 I am not against merging this one. But I believe it's a good idea to restrict each PR to a group of related changes.

mods/other/random_messages/init.lua
```diff
@@ -75,6 +75,10 @@ function random_messages.read_messages()
 		"Use /mp <player> to send a marker to a specific teammate",
 		"To check someone's league in all modes, use /league <player>",
 		"Use /leagues to list leagues and the placement needed to get to them.",
+		"Use /mhp or left/right click while holding the zoom key (default: Z)"
+		.. "pointed towards your feet to place an HP marker.",
+		"Use /map to print the current map name and map author.",
```
farooqkz, line 80:
```suggestion
		"Use /map to print the current map name and its author.",
```

src4026:
As a side note, I have not tested the death message change feature because apparently in singleplayer CTF/or the one I have locally, there is no respawn screen as on the official server.

LoneWolfHT:
> As a side note, I have not tested the death message change feature because apparently in singleplayer CTF/or the one I have locally, there is no respawn screen as on the official server.

Did you `/ctf_start` first?

mods/other/random_messages/init.lua
```diff
@@ -75,6 +75,10 @@ function random_messages.read_messages()
 		"Use /mp <player> to send a marker to a specific teammate",
 		"To check someone's league in all modes, use /league <player>",
 		"Use /leagues to list leagues and the placement needed to get to them.",
+		"To place an HP marker run /mhp or left/right click while holding the zoom key (default: Z) "..
```
a-blob, line 78:
```suggestion
		"To place an HP marker, run /mhp or left/right click while holding the zoom key (default: Z) "..
```

a-blob:
> As a side note, I have not tested the death message change feature because apparently in singleplayer CTF/or the one I have locally, there is no respawn screen as on the official server.

Respawn delay only applies if the match has started.

src4026:
Oh, that's why.

PR #1230, opened 2023-09-12
Prevent crash if missing barriers file

a-blob:
Fixes https://discord.com/channels/447819017391046687/448136831385468930/1148625553784578170
Prevents the game from crashing if some maps are missing the barrier files. Now it logs an error to specify which maps are missing the files.

image


LoneWolfHT:
Thanks, tweaked this implementation a little in e37adc999b2f93ee1eb67961b38deaa975d22654

PR #1238, opened 2023-10-16
Improve settings GUI

a-blob:
Improve the settings GUI so texts are no longer cut-off like this:
image

I also made the settings take the full inventory since I don't think anyone needs to view their inventory while using settings. Doing this also prevents users from getting through the trouble of scrolling through the settings inventory and accidentally modifying options with the scrollwheel.
image

### DIFF
--- a/mods/apis/ctf_settings/init.lua
+++ b/mods/apis/ctf_settings/init.lua
@@ -82,15 +82,15 @@ minetest.register_on_mods_loaded(function()
 			end
 
 			local form = {
-			{"box[-0.1,-0.1;%f,%f;#00000055]", FORMSIZE.x - SCROLLBAR_W, FORMSIZE.y},
+			{"box[-0.1,-0.1;%f,9.4;#00000055]", FORMSIZE.x - SCROLLBAR_W, FORMSIZE.y},
 				{"scroll_container[-0.1,0.3;%f,%f;settings_scrollbar;vertical;0.1]",
-					FORMSIZE.x - SCROLLBAR_W,
-					FORMSIZE.y + 0.7
+					FORMSIZE.x - SCROLLBAR_W + 2,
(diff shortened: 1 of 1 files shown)


src4026:
In case, there one happens to puts a long enough text as a setting, would it still cut off?

a-blob:
>In case, if one happens to put a long enough text as a setting, would it still cut off?
Yes. There are solutions though, like adding a new line or adding a horizontal scrollbar.

src4026:
> Yes. There are solutions though, like adding a new line or adding a horizontal scrollbar.

New lines would be temporary and not applicable to all, the scrollbar sounds better. Wrapping text could be another way if possible.

a-blob:
Wrapping text is not possible here AFAIK, unless you wrap it manually with \n.

mods/apis/ctf_settings/init.lua
```diff
@@ -82,23 +82,23 @@ minetest.register_on_mods_loaded(function()
 			end
 
 			local form = {
-			{"box[-0.1,-0.1;%f,%f;#00000055]", FORMSIZE.x - SCROLLBAR_W, FORMSIZE.y},
+			{"box[-0.1,-0.1;%f,9.4;#00000055]", FORMSIZE.x - SCROLLBAR_W, FORMSIZE.y},
```
LoneWolfHT, line 85: You removed the `%f` but not the value to be inserted there, `FORMSIZE.y`

LoneWolfHT: You should probably just edit the FORMSIZE variable at the top of the file, so you're not dealing with magic numbers

a-blob: Oh, my bad, completely overlooked that when committing
Yeah, I'll just edit the FORMSIZE variable.

PR #1244, opened 2023-11-23
Update ctf_map README.md to include details on the new Indestructible Water and Lava Barrier Glass

a-blob:


a-blob:
Might need some further improvement/more explanation.

PR #1249, opened 2023-11-26
Change the position of the inbox HUD and more language supports

Methro:
## Changes Made

1. **HUD Position Changed to Bottom Left Corner:**
   - The position of the mail HUD has been changed to the bottom left corner instead of the center. This was done due as the previous center position could be distracting during gameplay. The HUD will now be in a less intrusive location.

screenshot_20231126_124416

2. **Mail Message Made More Perceptible:**
   - The perception of the mail message uses `minetest.colorize` now, the "Mail from **playername**" message is highlighted in a specific color, making it easier to read in the chat and providing a clear indication to the player that they have received mail.
   
   
screenshot_20231126_124343

3. **Additional Support for Other Languages:**
   - Added support for additional languages, specifically, Spanish, Brazilian Portuguese and Russian for email mod

### DIFF
--- a/mods/other/email/hud.lua
+++ b/mods/other/email/hud.lua
@@ -11,7 +11,7 @@ function email.update_hud(player)
 			email.hud:add(player, "email:icon", {
 				hud_elem_type = "image",
 				name = "MailIcon",
-				position = {x=0.52, y=0.52},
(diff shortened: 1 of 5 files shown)


LoneWolfHT:
Thanks!

PR #1250, opened 2023-11-27
Make Reinforced Cobblestone harder to break for enemy team

a-blob:
Make it so that Reinforced Cobblestone is easier to break for the team that placed it but harder to break for other teams.

### DIFF
--- a/mods/ctf/ctf_map/ctf_traps.lua
+++ b/mods/ctf/ctf_map/ctf_traps.lua
@@ -173,6 +173,48 @@ minetest.register_node("ctf_map:reinforced_cobble", {
 	description = "Reinforced Cobblestone",
 	tiles = {"ctf_map_reinforced_cobble.png"},
 	is_ground_content = false,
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
This works? Feels a little hacky.
It might be better to make all reinforced cobble the hardened version, but have enemies dig it twice by replacing it with a second version of itself one dug by an enemy that has a cracked texture applied. Node dig predictions should be considered with that approach

a-blob:
A bit hacky, but works very well. I'd rather stick to this approach unless there's some unforeseen problem that might cause issues. Currently it is not complete though.

LoneWolfHT:
How does this work with internet lag? You might be able to simulate it with minetest.after()

a-blob:
The typical internet lag doesn't seem to affect it much. Bad internet lag (1+ seconds) makes it seem like it is dug _"twice by replacing it with a second version of itself once dug by an enemy"._

PR #1255, opened 2023-12-10
Add commands to replace destructible nodes with their indestructible variants and vice versa (if a variant exists)

src4026:
This PR aims to add the commands `/destructify` and `/indestructify` which replace indestructible nodes with their destructible variants and vice versa respectively.


LoneWolfHT:
Instead of looping through nodes manually try using `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])`. It might push some things over to the C++, and is easier to comprehend

src4026:
> Instead of looping through nodes manually try using `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])`. It might push some things over to the C++, and is easier to comprehend

If you're talking about the code from line number 256-258, I am not able to understand what I will need to put in the `nodenames` parameter in this case.

LoneWolfHT:
> If you're talking about the code from line number 256-258, I am not able to understand what I will need to put in the `nodenames` parameter in this case.

I copied that line from the lua_api.txt, did you check there?

src4026:
> > If you're talking about the code from line number 256-258, I am not able to understand what I will need to put in the `nodenames` parameter in this case.
> 
> I copied that line from the lua_api.txt, did you check there?

Yes, but since I have to enter either a node name (the example I saw in the API was `default:dirt`) or a group of nodes (the example in the API was `group:tree`). For the command, I'm looping through all the nodes and then checking if they start with `ctf_map:`; so, by your suggestion, I am not able to realise how I would go about doing it for only finding blocks that are of that item string -> the `group:` "method" for `nodename`?
Hmm, is there a `group:` for indestructible blocks? :thinking: Checking right now

src4026:
In the meanwhile, it looks like I also have to handle the "walls_" case, will commit soon

src4026:
> Hmm, is there a group: for indestructible blocks?

I've tried searching for this and am unsure if I've found it. Sorry if it is obvious, and I'm just not able to see it

LoneWolfHT:
> > Hmm, is there a group: for indestructible blocks?
> 
> I've tried searching for this and am unsure if I've found it. Sorry if it is obvious, and I'm just not able to see it

`immortal` is the group you're looking for https://github.com/MT-CTF/capturetheflag/blob/01e5bf87b1405986329e299a2576c60d075c1cef/mods/ctf/ctf_map/nodes.lua#L120

src4026:
Ah, I wasn't sure if it was that exactly but I did notice it in the node definitions. Thanks a lot!

src4026:
Hey there, just letting you know that I can work on this after 26th Jan only; a bit delayed

src4026:
So I've got `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])` working for the `/destructify` command in which I find all the `immortal` group blocks. I'll still have to resort to manual looping for the "non-`immortal`" nodes, isn't it? (or VoxelManip, but I don't think that might be needed?).

In the meanwhile, I'll try optimising my mod prefix fixes using the local `mod_prefixes` table provided in `nodes.lua`

LoneWolfHT:
> I fixed the `/indestructify` command's "indestructification" process by using the mod_prefixes table. But I'm finding the process to achieve the opposite mapping of mods to prefixes for the `/destructify` command unnecessarily long. Do you think I should leave it hardcoded or is a simpler way to use the `mod_prefixes` table to do this?

I would maybe change how things work so that the `mod_prefixes` table works both ways

LoneWolfHT:
> So I've got `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])` working for the `/destructify` command in which I find all the `immortal` group blocks. I'll still have to resort to manual looping for the "non-`immortal`" nodes, isn't it? (or VoxelManip, but I don't think that might be needed?).
> 
> In the meanwhile, I'll try optimising my mod prefix fixes using the local `mod_prefixes` table provided in `nodes.lua`

The manual method via voxelmanip might be best in the future, can wait and see how things go though

src4026:
Alright. I forgot to commit my previous changes before my vacations. Now I'll make a comitt with all the changes later

src4026:
Closed it since I have no intention of working on it. I would be glad if anybody else would like to continue and add the feature to map making.

PR #1258, opened 2023-12-29
Allow grenade explosions to break ice on water

Emojigit:
Currently, grenade explosions break snow layers. Thus, it should make sense to break the more fragile ice layers.

This PR is ready for review.


LoneWolfHT:
Sorry, just fixed this

PR #1265, opened 2024-01-29
Use voxel manip for barrier placing commands

a-blob:


appgurueu:
Have you benchmarked that this is actually substantially faster? Barriers are a bit special because they are relatively sparse. VoxelManip can't offer sub-mapblock granularity. So with this PR, you'd be loading & setting at least 15 non-barrier blocks for each barrier block. This may very well offset any performance gains from using VoxelManip.

a-blob:
The main problem with the current one is not speed, but that the barriers will have holes if a chunk within the selected area is not rendered yet.
image

LoneWolfHT (approved):
Is this ready and safe for merge?

a-blob:
@LoneWolfHT I think so

PR #1266, opened 2024-01-29
Prevent class switch cooldown if switching to the same class as before

a-blob:


Panquesito7 (approved):

farooqkz:
I approve as well. It's so trivial.

farooqkz:
@LoneWolfHT Why not merge it already?

farooqkz:
@a-blob can you please sync your fork with the master branch? I didn't notice this earlier.

PR #1277, opened 2024-04-02
Update bounty_algo.lua: new bounty calculation algorithm

farooqkz:

There has been two popular suggestions about bounties:

 - There shouldn't be a maximum value
 - There shouldn't be a bounty under 10

Other than these, it also makes sense to have bounty on strong useful medics.

This PR aims to address these suggestions. Since getting higher KDs is a harder than getting lower KDs, e.g. the hardship growth rate is itself polynomial, I've used an exponential function here to address that.

Edit:

Putting bounty on medics who heal a lot is now a popular suggestion.


farooqkz:
This PR needs to be tested on some server to see how it'll behave and produce what sort of results.

farooqkz:
Alright it is creating non sensible results on the experimental server.

PR #1284, opened 2024-05-05
Add TNT for CTF

farooqkz:
This PR aims to add TNTs to CTF, one of popular suggestions made by players.

 - [x] TNT should be placed only near enemy flag
 - [x] TNT shouldn't damage self or teammates
 - [x] TNT shouldn't remove unremovable blocks
 - [x] TNT sticks should be available in treasure chests
 - [x] TNT recipe
 - [x] TNT gives points for each block removed


farooqkz:
This PR is ready to review. Though I cannot tell how to fix the remaining luacheck lints. There is one issue though:

When TNT explodes and removes blocks, it gives score to the owner according to the number of removed blocks. Like 70 removed blocks give 70 score. And since a TNT can only be placed near enemy base, this is the score for enemy base sabotage and/or destruction. However, 70 score seems too much to give away. Perhaps I divide the removed blocks number by something like... 5?

AnonomousAccount:
Does it damage players? IMO it should be used to remove blocks, not kill people. That's what nades are for

AnonomousAccount:
Also, IDK about awarding score...

LoneWolfHT:
I'd rather not award score, and I'd like to keep the rules for placement as simple as possible.

I'd like to restrict this change to nade fight at first

farooqkz:
> I'd rather not award score, and I'd like to keep the rules for placement as simple and non-restrictive as possible.
> 
> It would probably be best to restrict this change to nade fight at first

I'm in favor of awarding score as TNT can be placed only near an enemy base. And it is most likely enemy base destruction which is score-worthy and contributes to the goal of game which is CTF.

farooqkz:
BTW, what do you think about adding a new stat: "Enemy base destruction". Could be number of blocks removed from the enemy base.

AnonomousAccount:
What count's as an "enemy base" tho, unfourtunatly, i can see how this could be exploited

farooqkz:
> What count's as an "enemy base" tho, unfourtunatly, i can see how this could be exploited

Hello. It cannot be exploited since TNTs can be placed only close to enemy base.

farooqkz:
It doesn't fit within the current gameplay and more thoughts should be put into if we are to add explosives which remove nodes to the game.

PR #1286, opened 2024-05-10
Add concept of HUD events channels and add the flag events to channel 1

farooqkz:

This PR aims to fix a bug: Since there is just one channel for HUD events per player, the flag events and score events are both shown in the same channel making it a bit annoying. This PR does these two:

 - Add concept of channels to `hud_events`. channel is `0-2` or `nil` which defaults to `0`.
 - Move flag events announcements from channel `0` to channel `1`

This enables players to see both events at the same time.
screenshot_20240510_130715


mods/apis/hud_events/init.lua
```diff
@@ -41,7 +47,7 @@ local function show_quick_hud_event(player, huddef)
 		hud:add(player, "hud_event_quick", {
 			hud_elem_type = "text",
 			position = {x = 0.5, y = 0.5},
-			offset = {x = 0, y = 45},
+			offset = {x = 0, y = 45 + (huddef.channel or 0) * 15},
```
LoneWolfHT, line 50: I would avoid assuming order of operations here and put in another set of parentheses

mods/apis/hud_events/init.lua
```diff
@@ -17,15 +17,21 @@ local HUD_COLORS = {
 	dark = 0x212529,
 }
 
-local hud_queues = {}
+local hud_queues = {
+	{}, -- channel 0
```
LoneWolfHT, line 21: Could you rework the code so that channels aren't hard-coded like this? That way you can use any channel you want when showing a HUD event

You could define some 'constant's in `hud_events` that map to a channel number if you want though (like an enum I guess)

LoneWolfHT (commented):
Nice, this is a simple but impactful change

mods/apis/hud_events/init.lua
farooqkz: okay which one should I do? the former or the latter?

mods/apis/hud_events/init.lua
farooqkz: quick events have got no channel

olliy1or: BODMAS rule says there shouldn't be an issue, is it followed on these kind of operations? X)

mods/apis/hud_events/init.lua
LoneWolfHT: The former at least, you can do both though

mods/apis/hud_events/init.lua
LoneWolfHT: I guess I'll assume this was tested locally

farooqkz:
Ready for another review.

PR #1290, opened 2024-05-13
Prevent death when inside of an indestructible block 

pl608:
- still damages if player is within 6 blocks of an enemy flag
- does not damage the player if it is within  6 blocks of *their own flag*


olliy1or:
Why? How can a player even get inside indestructible blocks

pl608:
You can clip into indestructible blocks by going up a shift/jump ladder
that ends in one. Another (and much more common way) is like on the Wall
map in one of the side doors if there is a 2 block pillar and you shift
jump into it you will clip into the roof and die

PR #1295, opened 2024-05-24
fix capture announcement message

olliy1or:
- This PR has been tested locally... NOT


LoneWolfHT (changes requested):
I'm pretty sure this will crash due to an invalid amount of arguments

---

I think the best way to do this is making the `text` var hold the `and got %d points`, but the `ctf_modebase.announce()` further down shouldn't

LoneWolfHT:
Closing until revived

PR #1300, opened 2024-06-06
Landmines V2

pl608:

- uses damage per second and on_hp_change to detect a player walking on the landmine (which also gives the landmine some randomness as to if it trips or not) 
- will damage players around it if they are not on the same team as the placer
- adds particles and sound effects identical to a frag nade
- the landmine texture was admittedly stolen from the other landmine pr
- probably this should have been a pr to the landmine pr but I did it how I did it
- and I nuked line 101 and 109 because I didn't like the look of them
itis not a chest item currently and can only be obtained via /give or creative inventory


farooqkz:
Landmines has been already implemented and ready to merge. Please don't open another PR when an already existing PR exists without coordinating with LV/LW.

PR #1301, opened 2024-06-07
Update nodes.lua

pl608:
I don't know if the way chests are currently, is intended but this only changes a treasure chest to visited if an item is taken from it.


mods/ctf/ctf_map/nodes.lua
```diff
@@ -239,11 +239,13 @@ local chest_def = {
...
 		end
+		minetest.swap_node(pos, {name = "ctf_map:chest_opened"})
+		minetest.get_meta(pos):set_string("infotext", chestv)
 	end,
-	on_rightclick = function(pos, node, clicker, itemstack, pointed_thing)
+	--[[on_rightclick = function(pos, node, clicker, itemstack, pointed_thing)
 		minetest.swap_node(pos, {name = "ctf_map:chest_opened"})
 		minetest.get_meta(pos):set_string("infotext", chestv)
-	end
+	end]]
```
LoneWolfHT, line 248: Delete it, don't comment it out. git/Github will save the old code for you

mods/ctf/ctf_map/nodes.lua
```diff
@@ -239,11 +239,13 @@ local chest_def = {
...
 }
 
 local ochest_def = table.copy(chest_def)
```
LoneWolfHT, line 251: This is for the open chest def, below here you should save the old inventory_take function and replace it in the def with one that does the swap checking, and then calls the old function. That way you won't be setting the chest to the open state when it's already in that state

LoneWolfHT (changes requested):

LoneWolfHT (approved):

PR #1302, opened 2024-06-10
give way more score for killing a flag thief

farooqkz:

The suggestion originates from here: https://discord.com/channels/447819017391046687/448136308108296213/1246872747838279710

The flag multiplier which multiplies the score is 1 for no flag, and multiplied by 2 for each flag added. Like 2 for a flag and 4 for two flags. Because it's way harder to not get killed and steal the second flag.


farooqkz:
Okay... I thought the change is trivial

farooqkz:
Self review done

mods/ctf/ctf_modebase/features.lua
```diff
@@ -191,7 +191,7 @@ local function calculate_killscore(player)
 	local flag_multiplier = 1
 	for tname, carrier in pairs(ctf_modebase.flag_taken) do
 		if carrier.p == player then
-			flag_multiplier = flag_multiplier + 0.25
+			flag_multiplier = flag_multiplier * 2
```
LoneWolfHT, line 194: I'm pretty sure this is a literal exponential increase. For now I'd suggest this instead:
```suggestion
			flag_multiplier = flag_multiplier + 1
```
That should result in the score being the normal(?) kill score * the number of flags they have, starting at x2 for one flag

LoneWolfHT (changes requested):

mods/ctf/ctf_modebase/features.lua
farooqkz: Yes it's an exponential increase. Getting flag of one team is not that hard. But getting flag of one team and at the same time getting flag of another one, is extremely hard. You have teammembers of the first team on your tail, everyone know your location every second, and then you have to approach base of another team to get their flag.

Therefore, I think exponential makes perfect sense.

LoneWolfHT: The suggestion is trying to encourage people to kill flag carriers, what you're saying is maybe justification for rewarding the flag carrier.

If this score increase isn't enough it could be changed to +2 or more

farooqkz: Maybe put it to a vote among players?

farooqkz: Or otherwise ask for a third party opinion. Since I don't agree with you, a third party opinion helps.

LoneWolfHT (approved):

PR #1303, opened 2024-06-10
Change donate command to accept multiple recipients

lazylier:
I have tested this about as well as I can think. I do not have the ability to test with more than 2 clients. It might not be written as cleanly as possible, although I've tried.

I've tried to keep flexibility with the order of parameters. Player names are checked for validity until there is at least one valid player name and a score amount, after which all remaining text is assumed to be the donation message.

[suggestion here](https://discord.com/channels/447819017391046687/448136308108296213/1249032532847099944)

### DIFF
--- a/mods/ctf/ctf_modebase/ranking_commands.lua
+++ b/mods/ctf/ctf_modebase/ranking_commands.lua
@@ -182,12 +182,13 @@ minetest.register_chatcommand("donate", {
 
 			if receivers[pname] then
 				return false, "You cannot donate more than once to the same person."
-			end
-
-			if not receivers[pname] then
+			else
 				receivers[pname] = true
 			end
+		end
 
+		local pname, names, count = next(receivers), "", 0
+		while pname do
 			current_mode.recent_rankings.add(pname, {score=score}, true)
 			current_mode.recent_rankings.add(name, {score=-score}, true)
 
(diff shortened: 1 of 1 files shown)


mods/ctf/ctf_modebase/ranking_commands.lua
```diff
@@ -161,17 +179,35 @@ minetest.register_chatcommand("donate", {
...
+				"Player '%s' donated %s score to player '%s'", name, score, pname
+			))
+
+			names = names .. pname
+
+			pname = next(pnames, pname)
+			if pname then
+				names = names .. ", "
+			end
+		end
```
LoneWolfHT, line 199: - This should be a `for pname, team in pairs(pnames) do`

- It might be faster to take `score_total` from the player donating before the loop, instead of running it multiple times.

- As for strings like `names`, I'd prefer to just do `names = names .. pname .. ", "` and `string:sub(1, -3)` after the loop to remove the trailing `", "`

mods/ctf/ctf_modebase/ranking_commands.lua
```diff
@@ -97,30 +97,57 @@ end)
...
 
-		if ctf_core.to_number(pname) then
-			pname, score = score, pname
-		end
+		local pcount = 0
 
-		dmessage = (dmessage and dmessage ~= "") and (":" .. dmessage) or ""
+		for p in string.gmatch(param, "%S+") do
+			if #dmessage > 0 then
+				dmessage = dmessage .. " " .. p
```
LoneWolfHT, line 113: Why is the list (table) size of a string being counted?

LoneWolfHT (changes requested):

mods/ctf/ctf_modebase/ranking_commands.lua
lazylier: > Why is the list (table) size of a string being counted?

That is because as soon as the message is started all subsequent words are assumed to be part of the message, so not to be checked for being score or player names.

LoneWolfHT (approved):

PR #1304, opened 2024-06-14
Add mushroom farming

farooqkz:

This PR is a draft because I think it needs more polish. Or it might not need. Reviews are very welcome!


LoneWolfHT:
What happens if you time the per-step lua time with `(endtime - starttime) / 1e6` using `minetest.get_us_time()` for the times? Placing a lot of mushrooms to test with?

farooqkz:
> What do you get if you time the per-step lua time with `(endtime - starttime) / 1e6` using `minetest.get_us_time()` for the times; Placing a lot of mushrooms to test with?

Not sure what you mean. Are you suggesting using this method instead of the current method with `globalstep`? BTW, This PR spawns mushrooms anywhere dark enough. So you don't need mushrooms to farm mushrooms. I think this is better since selfish players could prevent others from farming mushroom. If unsure, we can do a poll.

LoneWolfHT:
> Not sure what you mean. Are you suggesting using this method instead of the current method with `globalstep`?

What I posted is how you measure how much lag it's causing, by timing how long it takes to get from the start of the globalstep to the end of it

farooqkz:
Closing as it won't be merged.

PR #1305, opened 2024-06-17
Add new command /top

mrtechtroid:
- [x] This PR has been tested locally. 
- Based on suggestion [Discord](https://discord.com/channels/447819017391046687/1008040841102766153/1226029267209687161) have implemented the command `/top` whose majority code has been taken from `\top50`. 
- To prevent abuse/server lag, the value taken is limited to between [1,200] which can be updated to higher or lower as required.

### DIFF
--- a/mods/ctf/ctf_modebase/ranking_commands.lua
+++ b/mods/ctf/ctf_modebase/ranking_commands.lua
@@ -259,6 +259,42 @@ minetest.register_chatcommand("top50", {
 		})
 	end,
 })
+minetest.register_chatcommand("top", {
+	description = "Show the top players(value less than 200)",
+	params = "[mode:technical modename] <value>",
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
I feel like it might be a good idea to fix the ranking backends before merging this

mrtechtroid:
@LoneWolfHT now since backend is fixed? This could be merged?

LoneWolfHT:
> @LoneWolfHT now since backend is fixed? This could be merged? 

It will need to be reworked to fit the new API, and I would eventually like to change it to have a a GUI button like in the match summary, which may or may not be easy to do

LoneWolfHT:
Going to close this in favor of a website-based equivalent, since the website doesn't take up game thread time when accessing the rankings db

PR #1312, opened 2024-06-25
Fix bug - Poison grenades deal no damage to enemies.

mrtechtroid:
- Due to damage amount being very low, `1` instead of `10`, the poison grenade did not do damage at all to players. 
- Caused due to PR f847ac8

### DIFF
--- a/mods/pvp/grenades/grenades.lua
+++ b/mods/pvp/grenades/grenades.lua
@@ -205,7 +205,7 @@ local register_smoke_grenade = function(name, description, image, damage)
 								local dname = target:get_player_name()
 								local dteam = ctf_teams.get(dname)
 								if dname ~= pname and dteam ~= pteam then
-									target:punch(thrower, 1, {
+									target:punch(thrower, 10, {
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
That's punch_interval, the damage is down where `fleshy=1` is. Maybe that needs to be increased to 2? A punch interval of 1 should be fine

mrtechtroid:
It does work(does damage as expected), when the punch_interval was increased to 10. However could you let me know of any alternative approaches to fix it.

farooqkz:
hmm it works

farooqkz (approved):
works

LoneWolfHT:
> It does work(does damage as expected), when the punch_interval was increased to 10. However could you let me know of any alternative approaches to fix it.

There might be an on_punchplayer or something that's ignoring 1hp damage with punch intervals of 1

What does 2hp of fleshy damage do?

mrtechtroid:
Yes, punch interval 1, with fleshy = 2 also works, in fixing the issue.

PR #1315, opened 2024-06-26
Add the /flag command to teleport you to the flag during build time

Themostrandom:
[ x ] This PR has been tested locally
Suggestion : https://discord.com/channels/447819017391046687/448136308108296213/1255286879448989801


LoneWolfHT:
I don't like the idea of running a command to trigger the teleport 🤔

farooqkz:
> > I don't like the idea of running a command to trigger the teleport 🤔
> 
> I kinda dislike it too. But I am in favor of having a way to teleport back to flag during the build time. What is your suggestion? How about a button in inventory?

@Themostrandom To not waste your time, it is a good idea to consult LW about the implementation before you write the code.

LoneWolfHT:
> I kinda dislike it too. But I am in favor of having a way to teleport back to flag during the build time. What is your suggestion? How about a button in inventory?

I'm thinking more like punching the barrier to teleport back

farooqkz:
> > I kinda dislike it too. But I am in favor of having a way to teleport back to flag during the build time. What is your suggestion? How about a button in inventory?
> 
> I'm thinking more like punching the barrier to teleport back

And what if people are on the other side? Maybe punching the barrier *and* border glass?

mrtechtroid:
What about a new "bound_item" such as a compass, which when used during build time allows one to come back to flag?

farooqkz:
> What about a new "bound_item" such as a compass, which when used during build time allows one to come back to flag?

I would slightly prefer a button in inventory over this. Not really well thought about it however.

PR #1316, opened 2024-06-27
Modification of Paxel item

Themostrandom:
- Suggestion : my
- This mode just modifies the way the paxel is coded: its texture is no longer a combination of textures but a single texture: ctf_mode_classes_paxel. This allows you to change the texture of the item without changing the texture of the bronze shovel and pickaxe. (The texture I've put back is the old one).


LoneWolfHT:
This PR has some extra stuff in it that needs to be removed

Themostrandom:
> This PR has some extra stuff in it that needs to be removed

How ?

LoneWolfHT:
Are you aware of this? https://github.com/minetest/minetest/blob/master/doc/texture_packs.md#texture-overrides

PR #1321, opened 2024-06-29
add feature to steal from enemy teamchest while it is open in classic

farooqkz:


farooqkz:
Alright, this long awaiting feature finally would get added to CTF. It's working however when the close formspec event is fired, my function gets executed but player name won't get removed from the table for reasons I can't find now(due to lack of enough sleep?). @LoneWolfHT You are welcome to fix and merge it. The line is https://github.com/MT-CTF/capturetheflag/pull/1321/files#diff-34cefccf3df0840765bb2561068d664b0eed93136d8983bf0afe27a96587c234R133


BTW, doors and chests have their own mod now. Because otherwise `ctf_teams` would depend on `ctf_modebase` and vice versa, thus making a circular dependency tree.

farooqkz:
The problem has been solved. Ready for review and merge.

LoneWolfHT:
Adding this to classic makes me a little wary, since it wasn't a feature in CTFv1. Where was the discussion on this again?

farooqkz:
@LoneWolfHT  Status?

PR #1322, opened 2024-07-02
Update ranking_commands.lua

qasdwer01:
Just added a space in /donate's dmessage so that description becomes "donated 5 score to player1 and player2: good job" instead of "donated 5 score to player1 and player2:good job"


Themostrandom:
Hello, i think you can remove **:**

PR #1323, opened 2024-07-02
Modification of paxel item

Themostrandom:
-     [ x ] This PR has been tested locally
-     Suggestion : my
-     This mode just modifies the way the paxel is coded: its texture is no longer a combination of textures but a single texture: ctf_mode_classes_paxel. This allows you to change the texture of the item without changing the texture of the bronze shovel and pickaxe. (The texture I've put back is the old one).


LoneWolfHT:
Are you aware of this? https://github.com/minetest/minetest/blob/master/doc/texture_packs.md#texture-overrides

Themostrandom:
> Savez-vous cela? [https://github.com/minetest/minetest/blob/master/doc/texture-packs.md-texture-overrides](https://github.com/minetest/minetest/blob/master/doc/texture_packs.md#texture-overrides)

I think it's within the reach of more people to change the name of the texture, the change is small but avoids having to make a .txt...

LoneWolfHT:
> I think it's within the reach of more people to change the name of the texture, the change is small but avoids having to make a .txt... 
> 

I'd prefer to prominently provide a detailed tutorial.
There are a lot of other items/nodes in the game it would be painful and redundant to do this for.

PR #1328, opened 2024-07-15
Sound volume of flags

ClementMR:
- [ ] This PR has been tested locally.

Add a setting to adjust the sound volume of a flag.

It allows the player to configure their flag sound volume using a horizontal scroll bar.


LoneWolfHT:
I think the scrollbar settings should have a min/max and step option, so that you aren't stuck with the default options.

ClementMR:
> I think the scrollbar settings should have a min/max and step option, so that you aren't stuck with the default options.

Done! I also added a label on top of the setting

Capture d'écran 2024-07-18 154603

PR #1333, opened 2024-07-24
Fix two problems with returning dropped items

farooqkz:

 - Give score to a player for returning dropped items only when the match has started
 - Give score to a player for returning dropped items based on the value the item has


farooqkz:
Stale

PR #1336, opened 2024-08-06
Fix bug in nametag bug fix

Silvager:
A recent commit, "Attempt to fix nametag bug", commit id bff7fd0, introduced a bug which is fixed here.
The commit added some code to mods/other/playertag/init.lua, and in the code was a call to self.object:get_parent().
This function does not exist, and therefore the server would crash after the end of each match when it tried to call it.
This PR replaces self.object:get_parent() with an equivalent self.object:get_attach().
Since self.object:get_attach() works slightly differently than how self.object:get_parent() was supposed to work, a bit more code is added in to compensate.

- [x ] This PR has been tested locally

### DIFF
--- a/mods/other/playertag/init.lua
+++ b/mods/other/playertag/init.lua
@@ -174,7 +174,11 @@ minetest.register_entity("playertag:tag", {
 	on_punch = function() return true end,
 	on_deactivate = function(self, removal)
 		if not removal then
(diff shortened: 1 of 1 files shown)


LoneWolfHT:
Interesting, I put this on the main server immediately after pushing. No crashes yet

Silvager:
You're welcome!

PR #1349, opened 2024-10-16
add beds

astro-ctf:
Add beds (not interactable) to ctf.
No ctf_map beds, only the destroyable ones.


LoneWolfHT:
How hard would it be to make your changes to the beds from the `ctf_changes` mod, adding the beds mod to the update script? https://github.com/MT-CTF/capturetheflag/blob/master/scripts/update_mtg.sh

astro-ctf:
Done, added the beds and updated the mod update script.

LoneWolfHT:
You might be able to use this to extract the bed nodes https://content.luanti.org/packages/Warr1024/defripper/

astro-ctf:
@LoneWolfHT  I tried with defripper, but it didnt work, because it doesnt export functions and beds have some for e.g. on_rotate(). So I guess we go back to what I done first ?

LoneWolfHT:
> @LoneWolfHT I tried with defripper, but it didnt work, because it doesnt export functions and beds have some for e.g. on_rotate(). So I guess we go back to what I done first ?

You might be able to manually allow functions like that through using this: https://gitlab.com/sztest/defripper#custom-property-filtering

LoneWolfHT:
You may need to make a new PR since this one seems to delete the update_mtg.sh script

farooqkz:
What is status of this?

astro-ctf:
@farooqkz  Still working on it, but its gonna take some time because i have other things to do

PR #1350, opened 2024-11-02
allow donating maximum score by donating 0 or less

lazylier:

[Suggestion](https://discord.com/channels/447819017391046687/1008040841102766153/1288113654742519838)

This should allow players to donate half their score by using 0 or any negative number as the donation amount. I've tested it as well as I can.

### DIFF
--- a/mods/ctf/ctf_modebase/ranking_commands.lua
+++ b/mods/ctf/ctf_modebase/ranking_commands.lua
@@ -96,9 +96,8 @@ ctf_api.register_on_match_end(function()
 end)
 
 ctf_core.register_chatcommand_alias("donate", "d", {
-	description = [[Donate your match score to your teammate
-Can be used only once in 2.5 minutes
-Replace <score> with :max or any negative number to donate the maximum amount]],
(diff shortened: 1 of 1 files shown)


farooqkz:
Hello! Nice doing. However I don't like using `0` or negative numbers for this. I would prefer `max` for this. like `/donate LandarVargan max` to donate the maximum amount you can donate. I will review code after you've changed this. Cheers!

lazylier:
> Hello! Nice doing. However I don't like using `0` or negative numbers for this. I would prefer `max` for this. like `/donate LandarVargan max` to donate the maximum amount you can donate. I will review code after you've changed this. Cheers!

That would be more complicated since you'd have to also check if 'max' is a player name and figure out how to handle that

farooqkz:
Maybe `:max`?

LoneWolfHT:
I'm fine with either method, although 0 shouldn't turn into the max amount

lazylier:
now it should take :max or negative numbers, and I added a line to the description explaining that

farooqkz:
Code looks good to me.

LoneWolfHT (approved):

PR #1352, opened 2024-11-12
Pistolsound

Themostrandom:
- Suggestion : https://discord.com/channels/447819017391046687/448136308108296213/1305868186951880704


farooqkz:
IMO, this needs a poll, and the sound you submitted does not feed.

LoneWolfHT:
You should create a discord suggestion allowing people to play the sound and see if they like it. I'll escalate to a poll if I think it might be a bad idea despite positive reception

PR #1355, opened 2024-11-16
Buff pistol

Themostrandom:
I've improved the pistol which is an underused weapon because it's not strong, I've reduced the fire interval by 0.2 seconds and increased the damage by 0.5, this can be modified if you wish.

- Suggestion: https://discord.com/channels/447819017391046687/448136308108296213/1305893099884511243


farooqkz:
The suggestion on Discord is heavily downvoted.

LoneWolfHT:
You forgot to update the description, either way though I think this is something that most players don't want judging by the vote

PR #1363, opened 2024-12-04
Make landmine recipe cheaper

AnonomousAccount:
Makes Landmines craftable with damage cobble instead of frags. This makes then cheaper and also available to make in nade fight

### DIFF
--- a/mods/ctf/ctf_modebase/crafting.lua
+++ b/mods/ctf/ctf_modebase/crafting.lua
@@ -176,6 +176,6 @@ crafting.register_recipe({
 
 crafting.register_recipe({
 	output = "ctf_landmine:landmine",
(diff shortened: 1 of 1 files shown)


farooqkz:
@LoneWolfHT Let's forget about cosmetic and merge it with the recipe we both agree on. You said "4 steel ingots and 6 coal lumps". I think I agree with the general idea but I feel it's a bit cheap. Maybe 5 steel ingots? Anyway let's don't keep this waiting for too long.

LoneWolfHT:
I'll see how 4 steel goes

LoneWolfHT:
[1db60c1](https://github.com/MT-CTF/capturetheflag/commit/1db60c114719996dae1d28ccc97c19c4ded4129c)

PR #1366, opened 2024-12-09
New liquid: Poisonous Water

mrtechtroid:
- Two damage per second. (changable to 1 hp/sec)
- Greenish tint.
- Opinion Required: Whether to use Minetest ABM instead of damage_per_second parameter, where any player in a one-block-radius of the poisonous water, causes the player to get damaged. Like lets say a person is standing on a cobblestone, around which posinous water is surrounded, then he will also get a damage.


mods/other/poison_water/mod.conf
```diff
@@ -0,0 +1,6 @@
+name=poison_water
+description = Adds poisonous water and a poisonous water bucket.
+depends = default, bucket
+optional_depends =
```
farooqkz, line 4:
```suggestion
depends = default
optional_depends = bucket
```

mrtechtroid: Hi bucket is a necessary depend, else the code crashes.

farooqkz: hmm up to you. but you could make it optional using an `if`

mrtechtroid: Tried with an if and it caused a crash, thats why had to make it a depends.

mrtechtroid: The same reason is causing the Automatic Lua check to fail, since its not using the depends properly.

PR #1371, opened 2024-12-30
Adds volumetric lighting support.

astra0081X:
- [ ✅] This PR has been tested locally

### DIFF
--- a/mods/other/volumetric-lighting/LICENSE
+++ b/mods/other/volumetric-lighting/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 ROllerozxa
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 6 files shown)


farooqkz:
Thanks so much for this. I haven't tested. But at a first glance, the setting shouldn't be in the form of commands. CTF already has a dedicated settings api you can use to add per player settings.

astra0081X:
Could you clarify where I can find the settings API you're referring to?

farooqkz:
> Could you clarify where I can find the settings API you're referring to?

Search for `ctf_settings` in the code.

astra0081X:
Okay, so should I add it as a list or scroll bar?

astra0081X:
👍🏻

astra0081X:
Added using the settings api. The values are shown as percentages not numbers, it is up to the players to decide if we finalize in numbers or percentages.

LoneWolfHT:
Probably best to start with 0 to 2x.
I think for simplicity the command should be removed in favor of the settings menu too

astra0081X:
IIRC it already starts at 0. The reason the limit is 1 because any value other than that is very uncomfortable to look at. And the command has already been removed (https://github.com/MT-CTF/capturetheflag/pull/1371#issuecomment-2573288145)

LoneWolfHT:
Tried it out and ended up doing heavy tweaking, merged manually

PR #1384, opened 2025-03-14
Custom AFK-Kick message

AnonomousAccount:
Sends the message "Player left the game (inactivity)" when the AFK timer kicks a player

### DIFF
--- a/mods/other/afkkick/init.lua
+++ b/mods/other/afkkick/init.lua
@@ -45,6 +45,7 @@ minetest.register_on_joinplayer(function(player)
 end)
 
 minetest.register_on_leaveplayer(function(player)
+	minetest.chat_send_all("*** " .. player:get_player_name() .. " left the game (inactivity)")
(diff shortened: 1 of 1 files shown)


mods/other/afkkick/init.lua
```diff
@@ -45,6 +45,7 @@ minetest.register_on_joinplayer(function(player)
 end)
 
 minetest.register_on_leaveplayer(function(player)
+	minetest.chat_send_all("*** " .. player:get_player_name() .. " left the game (inactivity)")
```
LoneWolfHT, line 48: This will cause 2 leave messages to be sent every time someone leaves the game, regardless of if they were kicked for inactivity or not.

You'll probably want to selectively override `core.send_leave_message()` and append ` (inactivity)` to the end of the leave message if they were kicked for inactivity

LoneWolfHT (changes requested):

PR #1385, opened 2025-03-25
add coal lump crafting to classes and nade_fight

lazylier:

[suggestion](https://discord.com/channels/447819017391046687/448136308108296213/1349197066928394271)

This adds a recipe to craft 4 coal lumps from 1 coal block to classes and nades.

### DIFF
--- a/mods/ctf/ctf_modebase/crafting.lua
+++ b/mods/ctf/ctf_modebase/crafting.lua
@@ -141,6 +141,13 @@ crafting.register_recipe({
 	always_known = false,
 })
 
+-- Coal Lumps x4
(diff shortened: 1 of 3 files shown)


AnonomousAccount:
Coal blocks can already be used in furnaces, however this change would be helpful for making torches

lazylier:
> Coal blocks can already be used in furnaces, however this change would be helpful for making torches

I was thinking of landmines

AnonomousAccount:
> I was thinking of landmines

Hmmm better not add steel blocks -> steel ingots, otherwise people could craft 100s of mines on Appleton (and maybe some other maps)

PR #1394, opened 2025-05-09
Fix bug which can be exploited.

nrv-21:
The server can crash if entity is nil.


farooqkz:
Good catch! Real good contribution!

s1lverf1sh4:
Uploading PoC for future reference:

https://github.com/user-attachments/assets/03e29d0e-1bd0-4219-9ea6-7979875d196a

PR #1415, opened 2025-08-08
Make gun shooting look cooler

a-blob:

Adds some particles when the bullet hits stuff + minor tweaks to make the flying bullet more visible.
Demonstration:

https://github.com/user-attachments/assets/780a2c24-c8ac-47be-bb5e-88fa6b9d0639


a-blob:
Oh right forgot shotguns exist, disabled for them. I'll try a suggestion.

PR #1429, opened 2026-05-31
Fix medic healing.

Signal-mt:

Player HP was recently multiplied by 10 to permit finer-grained damage application; various damage sources thus had to also be multiplied by 10 to compensate. It seems that healing was missed, so this fixes it too. I didn't actually test this locally, but it's a pretty trivial change, and I followed the call chain and the logic checks out.

(Note that I wrote out `* 10` to make it more obvious that this scaling takes place.)


LoneWolfHT:
Thanks!

PR #1430, opened 2026-05-31
Fix display issues related to HP scaling.

Signal-mt:

Whilst playing, I noticed that a) the '\<player\> is already at %f HP' message had the HP scaled by 10, and b) that my HP healed stat was increasing more rapidly than it should have (10x more rapidly, actually). This should fix both those issues. Note that I rounded the number that goes into rankings to keep it an integer like everything else, but it could also be desirable to have it as a decimal value. Untested like before, but similarly trivial.


LoneWolfHT:
heh, nice catches

